PR #13, opened 2016-10-06
Dockerfile converted to Alpine Linux based image and PHP7

eniac111:
I made an image based on Alpine Linux with PHP7. Memcached is missing for now because it's unstable for php7.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,8 +17,9 @@ RUN { \
 # https://docs.nextcloud.com/server/9/admin_manual/installation/source_installation.html
 RUN set -ex \
   && apk update \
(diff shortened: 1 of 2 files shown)


pierreozoux:
Thanks a lot for your PR.

I'm currently waiting docker to review the current code to make it an [official image](https://github.com/docker-library/official-images/pull/2185).

Once this is done, we can start to add different tags (alpine and apache).
For that to work, we'd need some rework on the repo to have a folder architecture that would look like [WordPress](https://github.com/docker-library/wordpress) for instance.

I think the target would be to have 3 tags:
 - apache
 - fpm
 - alpine-fpm

If you can rework your PR to tend to this goal, this would facilitate a lot our work!

Thanks!

eniac111:
Of course, I'll give it some time in the weekend. 
Btw, I noticed a bug in this image today. Don't merge it for now.

PR #15, opened 2016-10-09
Install LDAP PHP extension

radhus:
The LDAP extension is required for LDAP integration.

With this extension installed, I was able to upgrade and migrate fine from Owncloud 9.1 to Nextcloud 10 with this Docker image!

Note; I see that a lot of development libraries are installed and kept, including the libldap2-dev I add here. Multiple RUN commands are also run which prevents cleanup from reducing the size of the final Docker image. Would it be wanted with a PR with merging the RUN's and with cleanup?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN apt-get update && apt-get install -y \
   libfreetype6-dev \
   libicu-dev \
   libjpeg-dev \
+  libldap2-dev \
   libmcrypt-dev \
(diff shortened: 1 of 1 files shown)


pierreozoux:
Thanks for your PR!

Where do you see that libs are kept?

radhus:
I think I misunderstood how development libraries are used by the PHP extensions, so ignore my comment for now :+1:

Issue #20, opened 2016-11-28
No version tags


cyphar:
Currently this repo (and the [associated repo on the Docker Hub]()) doesn't have any version tags. This means that users can't pin their install to a particular version. The best way of handling this would be to have different branches of this repo with different `nextcloud` versions pinned in each Dockerfile. AFAIK it should be possible to configure the Docker hub to handle this.

Also, are there any plans to get this integrated into the _actual_ `docker-library/`, like `owncloud` is? You could always emails @tianon and similar maintainers of that project to see if they'd be open to having `nextcloud` hosted in [a similar way to `owncloud`](https://github.com/docker-library/owncloud).

cyphar:
Yeah, there's an option for this when creating an automated build image (and can be configured after you create it).

cyphar:
Also, we really should provide `-apache` tags for Apache-based images (`php:5.6-apache`).

tianon:
See https://github.com/docker-library/official-images/pull/2185 for where this is proposed for inclusion as an official image. :smile: :+1: (with version tags, but not an Apache variant)

cyphar:
Alright, I'll make PRs against this project to add Apache variants as well as subdirectories for each major/minor version.

cyphar:
#21 has my changes.

PR #21, opened 2016-11-29
version: move to separate versions and variants

cyphar:
In order to match ownCloud's Docker images, we need to provide -apache
and -fpm variant images as well as maintaining version tags for older
images.

These scripts are improved versions of the ownCloud scripts[1] because
they also automatically generate the directories with each ./update.sh,
and also generate all aliases without needing human intervention.

[1]: https://github.com/docker-library/owncloud/blob/master/generate-stackbrew-library.sh

Fixes: #20
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/10.0/apache/Dockerfile
+++ b/10.0/apache/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:5.6-fpm
+FROM php:5.6-apache
 
 RUN apt-get update && apt-get install -y \
   bzip2 \
@@ -29,6 +29,7 @@ RUN { \
     echo 'opcache.fast_shutdown=1'; \
     echo 'opcache.enable_cli=1'; \
   } > /usr/local/etc/php/conf.d/opcache-recommended.ini
+RUN a2enmod rewrite
 
 # PECL extensions
 RUN set -ex \
@@ -36,8 +37,9 @@ RUN set -ex \
  && pecl install memcached-2.2.0 \
  && pecl install redis-2.2.8 \
  && docker-php-ext-enable apcu redis memcached
+RUN a2enmod rewrite
 
-ENV NEXTCLOUD_VERSION 10.0.1
+ENV NEXTCLOUD_VERSION 10.0.2
 VOLUME /var/www/html
 
 RUN curl -fsSL -o nextcloud.tar.bz2 \
@@ -55,4 +57,4 @@ RUN curl -fsSL -o nextcloud.tar.bz2 \
 COPY docker-entrypoint.sh /entrypoint.sh
 
 ENTRYPOINT ["/entrypoint.sh"]
-CMD ["php-fpm"]
+CMD ["apache2-foreground"]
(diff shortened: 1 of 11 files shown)


cyphar:
This will need to be matched with a corresponding change to docker-library/official-images#2185 so that we use the output of `generate-stackbrew-library.sh`.

cyphar:
This is what the new stackbrew config looks like:

```
# This file is generated via https://github.com/nextcloud/docker/blob/8f5641eda47fa1bbaab34c991565db6a06989987/generate-stackbrew-library.sh

Maintainers: Nextcloud <docker@nextcloud.com> (@nextcloud),
             Pierre Ozoux <pierre@ozoux.net> (@pierreozoux)
GitRepo: https://github.com/nextcloud/docker.git

Tags: 10.0.1-apache, 10.0-apache, 10-apache, apache, 10.0.1, 10.0, 10, latest
GitCommit: 8f5641eda47fa1bbaab34c991565db6a06989987
Directory: 10.0/apache

Tags: 10.0.1-fpm, 10.0-fpm, 10-fpm, fpm
GitCommit: 8f5641eda47fa1bbaab34c991565db6a06989987
Directory: 10.0/fpm

Tags: 9.0.54-apache, 9.0-apache, 9-apache, 9.0.54, 9.0, 9
GitCommit: 8f5641eda47fa1bbaab34c991565db6a06989987
Directory: 9.0/apache

Tags: 9.0.54-fpm, 9.0-fpm, 9-fpm
GitCommit: 8f5641eda47fa1bbaab34c991565db6a06989987
Directory: 9.0/fpm
```

pierreozoux:
Thanks a lot for this!

Again, I'm waiting https://github.com/docker-library/official-images/pull/2185 to be merged to make it as quick as possible.

Then we'll add the different features/tags.

Quick question for nextcloud maintainers @Gomez @LukasReschke (ping the relevant person if you know):
 - which versions are planned to be maintained?
Will we maintain almost all like in ownCloud? Or just the 2 lasts? It is an importatn question as it will help us shape right this repo.

pierreozoux:
@Gomez @LukasReschke can you answer my questions above please? Thanks!
#25 depends on it also.

Gomez:
@MorrisJobke Should know this too

MorrisJobke:
In the server we maintain basically last two stable versions and limited support for all other (mostly security stuff) to not be overwhelmed by all the backport. This also highly depends on how much customers are on which version, but we recommend to stay with one of the two latest major version (currently 10.x and 11.x)

I think those versions should be enough for this docker image

pierreozoux:
@cyphar would you be kind enough and update the PR with 10.0.2 and 11.0.0?

Thanks a lot again for your amazing work :)

cyphar:
Sure, I'll fix this up.

pierreozoux:
Thanks a lot!

PR #25, opened 2017-01-04
Upgrade to Nextcloud 11

gorlug:
I'm doing a fresh Nextcloud install and wanted to use the new version 11 and just replacing the version number in the Dockerfile seems to be working fine.

Or ist there some kind of upgrade function inside the image planned and that's why it's not updated yet? 

Anyway, thanks for the Dockerfile!


pierreozoux:
Will close in favor of #21

PR #39, opened 2017-03-03
Update from 11.0.1 to 11.0.2

ambis:
Tested by building apache image locally and running through the installer with sqlite. Was also built against the merged 7.1 PR. Verified 11.0.2 was in version.php.

This is my first PR in years so hope I got things right.


cyphar:
Just run `update.sh` which will update all supported versions. #41 did that afaics.

ambis:
Ah, right. Thanks!

PR #42, opened 2017-03-03
Changed the php version back to 5.6 for Nextcloud 9 and 10

### LINKED ISSUE #40: "This version of Nextcloud is not compatible with PHP 7.1"
I might be missing something basic, but I can no longer run the nextcloud 9 or 10 docker images (v11 works fine)

I get the following output from my build script (pasted in below). Instead of "php occ upgrade" actually running (in this example it should tell me that Nextcloud is not installed) it says:
`This version of Nextcloud is not compatible with PHP 7.1.<br/>You are currently running 7.1.2.`

I don't think v9 or v10 support PHP 7.1 ... ? I expect this was introduced as of #34. Sadly I need v9 and v10 in order to perform a migration from OC v8.

build.sh :
```
#!/bin/sh

if [ "$1" == "" ]; then
        echo Usage: $0 version
        echo e.g. $0 10.0
        exit
fi
NEXTCLOUDVER=$1

if [ -f docker-entrypoint.sh ]; then rm docker-entrypoint.sh; fi
if [ -f Dockerfile ]; then rm Dockerfile; fi

wget https://github.com/nextcloud/docker/raw/master/${NEXTCLOUDVER}/apache/docker-entrypoint.sh https://github.com/nextcloud/docker/raw/master/${NEXTCLOUDVER}/apache/Dockerfile
chmod +x docker-entrypoint.sh

# Stop any running nextcloud instances (any)
docker ps | grep nextcloud-test- | tr -s ' ' | cut -f 2 -d ' ' | xargs -n 1 docker stop

# Remove the old instance before we rebuild it
docker rm nextcloud-test-${NEXTCLOUDVER}

docker build -t nextcloud-test-${NEXTCLOUDVER} . && \
docker run --name nextcloud-test-${NEXTCLOUDVER} \
        -d \
        -p 127.0.0.1:8081:80 \
        nextcloud-test-${NEXTCLOUDVER}

docker exec -it nextcloud-test-${NEXTCLOUDVER} bash -c "apt-get update && apt-get install -y sudo vim"

# This will fail:
docker exec -it nextcloud-test-${NEXTCLOUDVER} bash -c "sudo -u www-data php occ upgrade"

```

tilosp:
Closes #40

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -16,7 +16,11 @@ for latest in "${latests[@]}"; do
 	for variant in apache fpm; do
 		# Create the version+variant directory with a Dockerfile.
 		mkdir -p "$version/$variant"
-		cp Dockerfile.template "$version/$variant/Dockerfile"
+		if [ "$version" == "11.0" ]; then
+			cp Dockerfile-php7.template "$version/$variant/Dockerfile"
+		else
+			cp Dockerfile.template "$version/$variant/Dockerfile"
+		fi
 
 		echo "updating $latest [$version] $variant"
 
--- a/update.sh
+++ b/update.sh
@@ -6,6 +6,11 @@ declare -A cmd=(
 	[fpm]='php-fpm'
 )
 
+# version_greater_or_equal A B returns whether A >= B
+function version_greater_or_equal() {
+	[[ "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1" || "$1" == "$2" ]];
+}
+
 latests=( $(curl -sSL 'https://nextcloud.com/changelog/' |tac|tac| \
 	grep -o "\(Version\|Release\)\s\+[[:digit:]]\+\(\.[[:digit:]]\+\)\+" | \
 	awk '{ print $2 }' | sort -V ) )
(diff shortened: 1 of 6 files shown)


update.sh
```diff
@@ -16,7 +16,11 @@ for latest in "${latests[@]}"; do
 	for variant in apache fpm; do
 		# Create the version+variant directory with a Dockerfile.
 		mkdir -p "$version/$variant"
-		cp Dockerfile.template "$version/$variant/Dockerfile"
+		if [ "$version" == "11.0" ]; then
```
cyphar: This is not really a future-compatible change. I would recommend defining a function like this:

```
# version_gte A B returns whether A >= B
function version_gte() {
	[[ "$(printf '%s\n' "$@" | sort -V | head -n1)" != "$1" || "$1" == "$2" ]];
}
```

Which you could then use like this:

```
template="Dockerfile.template"
if version_gte "$version" "11.0"; then
    template="Dockerfile-php7.template"
fi
cp "$template" "$version/$variant/Dockerfile"
```

pierreozoux:
Thanks a lot @tilosp for this, this is amazing, I'll merge it right away.

PR #47, opened 2017-03-13
Add some examples on how to use this image

pierreozoux:


jkaberg:
While I agreed on the "one process per container" philosophy I think you should mention how to do this for novice users. Might save you a few issue tickets 😄 

I did it by creating an 15 min cron on the host running nextcloud:latest image, the cron command:
```docker exec nextcloud su - www-data -s /bin/bash -c 'php -f /var/www/html/cron.php'```

PR #22, opened 2016-12-07
ADD Support of SMBClient for External storage

jodumont:
It might be nice to support SMBClient for External storage.
It could be possible just by adding smbclient package.

PR #60, opened 2017-04-01
add smb client

jmccoy555:


tilosp:
We have already decided to not include the smb client. #22 
If you need smb support you can use this [Dockerfile](https://github.com/nextcloud/docker/blob/master/.examples/Dockerfile.smb)

Issue #26, opened 2017-01-05
For big installations, chown of data folder is taking a looong time


pierreozoux:
This line has to be optimised:
https://github.com/nextcloud/docker/blob/master/docker-entrypoint.sh#L6

arnaudmm:
Hello, same probleme here, why the entrypoint.sh does a "tar -C |tar xf" by the way ?

As used here, we don't mount entire /var/www/html, but apps/ config/ and data/ separatly, so this system is not useful I think, and source could be moved to the right place at build time (in Dockerfile)

correct me if i'm wrong

SnowMB:
Yeah we have been discussing this matter a little bit. See here https://github.com/nextcloud/docker/pull/23

tilosp:
Fixed in #65

PR #65, opened 2017-04-14
Rework Updating

tilosp:
TODO:

- [x] version comparison

- [x] occ upgrade

- [x] show disabled apps

- [x] split the app folder

- [x] disable the updater

- [x] replace `chown -R www-data /var/www/html` #26

### DIFF
--- a/10.0/apache/apps.config.php
+++ b/10.0/apache/apps.config.php
@@ -1,0 +1,15 @@
+<?php
+$CONFIG = array (
+  "apps_paths" => array (
+      0 => array (
+              "path"     => OC::$SERVERROOT."/apps",
+              "url"      => "/apps",
+              "writable" => false,
+      ),
+      1 => array (
+              "path"     => OC::$SERVERROOT."/custom_apps",
+              "url"      => "/custom_apps",
+              "writable" => true,
+      ),
+  ),
+);
--- a/10.0/apache/docker-entrypoint.sh
+++ b/10.0/apache/docker-entrypoint.sh
@@ -13,7 +13,7 @@ fi
 image_version=$(php -r 'require "/usr/src/nextcloud/version.php"; echo "$OC_VersionString";')
 
 if version_greater "$installed_version" "$image_version"; then
-    echo "Downgrade not supported"
+    echo "Can't start Nextcloud because the version of the data ($installed_version) is higher than the docker image version ($image_version) and downgrading is not supported. Are you sure you have pulled the newest image version?"
     exit 1
 fi
 
--- a/10.0/apache/docker-entrypoint.sh
+++ b/10.0/apache/docker-entrypoint.sh
@@ -18,6 +18,10 @@ if version_greater "$installed_version" "$image_version"; then
 fi
 
 if version_greater "$image_version" "$installed_version"; then
+    if [ "$installed_version" != "0.0.0~unknown" ]; then
+        su - www-data -s /bin/bash -c 'php /var/www/html/occ app:list' > /tmp/list_before
+    fi
+
     rsync -a --delete --exclude /config/ --exclude /data/ --exclude /apps/ /usr/src/nextcloud/ /var/www/html/
 
     if [ ! -d /var/www/html/config ]; then
@@ -34,6 +38,11 @@ if version_greater "$image_version" "$installed_version"; then
 
     if [ "$installed_version" != "0.0.0~unknown" ]; then
         su - www-data -s /bin/bash -c 'php /var/www/html/occ upgrade --no-app-disable'
+
+        su - www-data -s /bin/bash -c 'php /var/www/html/occ app:list' > /tmp/list_after
+        echo "The following apps have beed disabled:"
+        diff <(sed -n "/Enabled:/,/Disabled:/p" /tmp/list_before) <(sed -n "/Enabled:/,/Disabled:/p" /tmp/list_after) | grep '<' | cut -d- -f2 | cut -d: -f1
+        rm -f /tmp/list_before /tmp/list_after
     fi
 fi
 
--- a/10.0/apache/docker-entrypoint.sh
+++ b/10.0/apache/docker-entrypoint.sh
@@ -28,9 +28,14 @@ if version_greater "$image_version" "$installed_version"; then
         cp -arT /usr/src/nextcloud/config /var/www/html/config
     fi
 
-    mkdir -p /var/www/html/custom_apps
+    if [ ! -d /var/www/html/data ]; then
+        cp -arT /usr/src/nextcloud/data /var/www/html/data
+    fi
 
-    chown -R www-data /var/www/html
+    if [ ! -d /var/www/html/custom_apps ]; then
+        cp -arT /usr/src/nextcloud/custom_apps /var/www/html/custom_apps
+        cp -a /usr/src/nextcloud/config/apps.config.php /var/www/html/config/apps.config.php
+    fi
 
     if [ "$installed_version" != "0.0.0~unknown" ]; then
         su - www-data -s /bin/bash -c 'php /var/www/html/occ upgrade --no-app-disable'
--- a/10.0/apache/docker-entrypoint.sh
+++ b/10.0/apache/docker-entrypoint.sh
@@ -31,6 +31,10 @@ if version_greater "$image_version" "$installed_version"; then
     done
 
     chown -R www-data /var/www/html
+
+    if [ "$installed_version" != "0.0.0~unknown" ]; then
+        su - www-data -s /bin/bash -c 'php /var/www/html/occ upgrade --no-app-disable'
+    fi
 fi
 
 exec "$@"
--- a/10.0/apache/docker-entrypoint.sh
+++ b/10.0/apache/docker-entrypoint.sh
@@ -18,7 +18,18 @@ if version_greater "$installed_version" "$image_version"; then
 fi
 
 if version_greater "$image_version" "$installed_version"; then
-    tar cf - --one-file-system -C /usr/src/nextcloud . | tar xf -
+    rsync -a --delete --exclude /config/ --exclude /data/ --exclude /apps/ /usr/src/nextcloud/ /var/www/html/
+
+    if [ ! -d /var/www/html/config ]; then
+        cp -arT /usr/src/nextcloud/config /var/www/html/config
+    fi
+
+    mkdir -p /var/www/html/apps
+    for app in `find /usr/src/nextcloud/apps -maxdepth 1 -mindepth 1 -type d | cut -d / -f 6`; do
+        rm -rf /var/www/html/apps/$app
+        cp -arT /usr/src/nextcloud/apps/$app /var/www/html/apps/$app
+    done
+
     chown -R www-data /var/www/html
 fi
 
--- a/10.0/apache/docker-entrypoint.sh
+++ b/10.0/apache/docker-entrypoint.sh
@@ -1,7 +1,23 @@
 #!/bin/bash
 set -e
 
-if [ ! -e '/var/www/html/version.php' ]; then
+# version_greater A B returns whether A > B
+function version_greater() {
+	[[ "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1" ]];
+}
+
+installed_version="0.0.0~unknown"
+if [ -f /var/www/html/version.php ]; then
+    installed_version=$(php -r 'require "/var/www/html/version.php"; echo "$OC_VersionString";')
+fi
+image_version=$(php -r 'require "/usr/src/nextcloud/version.php"; echo "$OC_VersionString";')
+
+if version_greater "$installed_version" "$image_version"; then
+    echo "Downgrade not supported"
+    exit 1
+fi
+
+if version_greater "$image_version" "$installed_version"; then
     tar cf - --one-file-system -C /usr/src/nextcloud . | tar xf -
     chown -R www-data /var/www/html
 fi
--- a/10.0/apache/docker-entrypoint.sh
+++ b/10.0/apache/docker-entrypoint.sh
@@ -22,17 +22,13 @@ if version_greater "$image_version" "$installed_version"; then
         su - www-data -s /bin/bash -c 'php /var/www/html/occ app:list' > /tmp/list_before
     fi
 
-    rsync -a --delete --exclude /config/ --exclude /data/ --exclude /apps/ /usr/src/nextcloud/ /var/www/html/
+    rsync -a --delete --exclude /config/ --exclude /data/ --exclude /custom_apps/ /usr/src/nextcloud/ /var/www/html/
 
     if [ ! -d /var/www/html/config ]; then
         cp -arT /usr/src/nextcloud/config /var/www/html/config
     fi
 
-    mkdir -p /var/www/html/apps
-    for app in `find /usr/src/nextcloud/apps -maxdepth 1 -mindepth 1 -type d | cut -d / -f 6`; do
-        rm -rf /var/www/html/apps/$app
-        cp -arT /usr/src/nextcloud/apps/$app /var/www/html/apps/$app
-    done
+    mkdir -p /var/www/html/custom_apps
 
     chown -R www-data /var/www/html
 
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -13,7 +13,7 @@ fi
 image_version=$(php -r 'require "/usr/src/nextcloud/version.php"; echo "$OC_VersionString";')
 
 if version_greater "$installed_version" "$image_version"; then
-    echo "Downgrade not supported"
+    echo "Can't start Nextcloud because the version of the data ($installed_version) is higher than the docker image version ($image_version) and downgrading is not supported. Are you sure you have pulled the newest image version?"
     exit 1
 fi
 
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -18,6 +18,10 @@ if version_greater "$installed_version" "$image_version"; then
 fi
 
 if version_greater "$image_version" "$installed_version"; then
+    if [ "$installed_version" != "0.0.0~unknown" ]; then
+        su - www-data -s /bin/bash -c 'php /var/www/html/occ app:list' > /tmp/list_before
+    fi
+
     rsync -a --delete --exclude /config/ --exclude /data/ --exclude /apps/ /usr/src/nextcloud/ /var/www/html/
 
     if [ ! -d /var/www/html/config ]; then
@@ -34,6 +38,11 @@ if version_greater "$image_version" "$installed_version"; then
 
     if [ "$installed_version" != "0.0.0~unknown" ]; then
         su - www-data -s /bin/bash -c 'php /var/www/html/occ upgrade --no-app-disable'
+
+        su - www-data -s /bin/bash -c 'php /var/www/html/occ app:list' > /tmp/list_after
+        echo "The following apps have beed disabled:"
+        diff <(sed -n "/Enabled:/,/Disabled:/p" /tmp/list_before) <(sed -n "/Enabled:/,/Disabled:/p" /tmp/list_after) | grep '<' | cut -d- -f2 | cut -d: -f1
+        rm -f /tmp/list_before /tmp/list_after
     fi
 fi
 
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -28,9 +28,14 @@ if version_greater "$image_version" "$installed_version"; then
         cp -arT /usr/src/nextcloud/config /var/www/html/config
     fi
 
-    mkdir -p /var/www/html/custom_apps
+    if [ ! -d /var/www/html/data ]; then
+        cp -arT /usr/src/nextcloud/data /var/www/html/data
+    fi
 
-    chown -R www-data /var/www/html
+    if [ ! -d /var/www/html/custom_apps ]; then
+        cp -arT /usr/src/nextcloud/custom_apps /var/www/html/custom_apps
+        cp -a /usr/src/nextcloud/config/apps.config.php /var/www/html/config/apps.config.php
+    fi
 
     if [ "$installed_version" != "0.0.0~unknown" ]; then
         su - www-data -s /bin/bash -c 'php /var/www/html/occ upgrade --no-app-disable'
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -31,6 +31,10 @@ if version_greater "$image_version" "$installed_version"; then
     done
 
     chown -R www-data /var/www/html
+
+    if [ "$installed_version" != "0.0.0~unknown" ]; then
+        su - www-data -s /bin/bash -c 'php /var/www/html/occ upgrade --no-app-disable'
+    fi
 fi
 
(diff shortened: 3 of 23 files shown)


mkuron:
Wouldn't it make sense to automatically execute `occ upgrade` after installing an update? Automatically performing any needed database migrations is standard on most other dockerized applications.

tilosp:
You are right. This PR is WIP and i want to implement this pseudo code of @pierreozoux:
```
if version is different in /usr/src and /var/www/html then:
  tar cf - --one-file-system -C /usr/src/nextcloud . | tar xf -
  php occ app:list > /tmp/list_before
  for app in `ls /usr/src/nextcloud/apps/`; do rm -rf ./apps/$app;cp -TR /usr/src/nextcloud/apps/$app/ ./apps/$app/; chown -R www-data:www-data ./apps/$app/; done
  php occ upgrade --no-app-disable
  php occ app:list > /tmp/list_after
  echo "the following app have beed disabled:
  diff <(sed -n "/Enabled:/,/Disabled:/p" /tmp/list_before) <(sed -n "/Enabled:/,/Disabled:/p" /tmp/list_after) | grep '<' | cut -d- -f2 | cut -d: -f1
fi
```
  echo "the following app have beed disabled:
  diff <(sed -n "/Enabled:/,/Disabled:/p" /tmp/list_before) <(sed -n "/Enabled:/,/Disabled:/p" /tmp/list_after) | grep '<' | cut -d- -f2 | cut -d: -f1
fi`

10.0/apache/docker-entrypoint.sh
```diff
@@ -18,7 +18,18 @@ if version_greater "$installed_version" "$image_version"; then
 fi
 
 if version_greater "$image_version" "$installed_version"; then
-    tar cf - --one-file-system -C /usr/src/nextcloud . | tar xf -
+    rsync -a --delete --exclude /config/ --exclude /data/ --exclude /apps/ /usr/src/nextcloud/ /var/www/html/
```
mkuron: Is there any reason why the tarball isn't directly extracted to /var/www/html in the Dockerfile? Then you wouldn't need to copy here.

tilosp: Yes because we need to use a volume to mount it into nginx. And a container on its own will not overwrite the content of a volume.

10.0/apache/docker-entrypoint.sh
```diff
@@ -1,9 +1,49 @@
...
+	[[ "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1" ]];
+}
+
+installed_version="0.0.0"
+if [ -f /var/www/html/version.php ]; then
+    installed_version=$(php -r 'require "/var/www/html/version.php"; echo "$OC_VersionString";')
+fi
+image_version=$(php -r 'require "/usr/src/nextcloud/version.php"; echo "$OC_VersionString";')
+
+if version_greater "$installed_version" "$image_version"; then
```
tilosp: This check will break the update app. This mean if we keep it we should disable the update app.

10.0/apache/docker-entrypoint.sh
```diff
@@ -1,9 +1,49 @@
...
+    echo "Downgrade not supported"
```
tilosp: This check will break the update app. This mean if we keep it we should disable the update app.

docker-entrypoint.sh
```diff
@@ -1,9 +1,49 @@
...
+installed_version="0.0.0"
+if [ -f /var/www/html/version.php ]; then
+    installed_version=$(php -r 'require "/var/www/html/version.php"; echo "$OC_VersionString";')
+fi
+image_version=$(php -r 'require "/usr/src/nextcloud/version.php"; echo "$OC_VersionString";')
+
+if version_greater "$installed_version" "$image_version"; then
+    echo "Downgrade not supported"
+    exit 1
+fi
```
tilosp: This check will break the update app. This mean if we keep it we should disable the update app.

10.0/apache/apps.config.php
```diff
@@ -0,0 +1,15 @@
+<?php
+$CONFIG = array (
+  "apps_paths" => array (
+      0 => array (
+              "path"     => OC::$SERVERROOT."/apps",
+              "url"      => "/apps",
+              "writable" => false,
+      ),
+      1 => array (
+              "path"     => OC::$SERVERROOT."/apps2",
```
pierreozoux: What about having a better naming than `2`?
What about `custom_apps` or `user_apps` ?

tilosp: done

docker-entrypoint.sh
```diff
@@ -1,9 +1,45 @@
...
 set -e
 
-if [ ! -e '/var/www/html/version.php' ]; then
-    tar cf - --one-file-system -C /usr/src/nextcloud . | tar xf -
+# version_greater A B returns whether A > B
+function version_greater() {
+	[[ "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1" ]];
+}
+
+installed_version="0.0.0"
```
cyphar: Minor nit, it would be nice if we made this `0.0.0~unknown` or something like that. But it's not necessary.

docker-entrypoint.sh
```diff
@@ -1,9 +1,45 @@
...
+fi
+image_version=$(php -r 'require "/usr/src/nextcloud/version.php"; echo "$OC_VersionString";')
+
+if version_greater "$installed_version" "$image_version"; then
+    echo "Downgrade not supported"
+    exit 1
+fi
+
+if version_greater "$image_version" "$installed_version"; then
+    if version_greater "$installed_version" "0.0.0"; then
```
cyphar: Maybe we should do something like `[[ "$installed_version" != "0.0.0" ]]` in all of these cases to make it more clear which is the upgrade path and which is the fresh install path?

docker-entrypoint.sh
tilosp: this could cause problems inside the `version_greater` function

docker-entrypoint.sh
tilosp: done

docker-entrypoint.sh
cyphar: No it won't -- I work with rpm-style versions quite a lot. `sort -V` does version checks against the leading digits first. In general you would use `~` for things like `1.0.0~rc1` which is a **lower** version number than `1.0.0` (or `1.0.0~rc2`). However something like `0.0.0+rc1` is a **higher** version number than `1.0.0`.

tilosp:
> Though, can you explain what you mean in the issue description by "replace chown -R www-data"?

I want to fix #26.
`rsync -a` should copy the owner and this mean we could just set the right owner in the source dir at build time

docker-entrypoint.sh
tilosp: thanks for explaining, i changed it.

10.0/apache/docker-entrypoint.sh
```diff
@@ -1,9 +1,45 @@
...
+}
+
+installed_version="0.0.0~unknown"
+if [ -f /var/www/html/version.php ]; then
+    installed_version=$(php -r 'require "/var/www/html/version.php"; echo "$OC_VersionString";')
+fi
+image_version=$(php -r 'require "/usr/src/nextcloud/version.php"; echo "$OC_VersionString";')
+
+if version_greater "$installed_version" "$image_version"; then
+    echo "Downgrade not supported"
```
SnowMB: Should probably be more specific. Something like 
```
echo "Can't start Nextcloud because the version of the data ($installed_version) is higher than the docker image version ($image_version) and downgrading is not supported. Are you sure you have pulled the newest image version?"
```

ChessSpider:
Hi Guys,
I'm wondering, why we're holding on to the extracting source-code into the directory?

If I pull version 11.0.2-fpm, or fpm, or latest, then I would reasonably expect to find that source-code in the source directory.

I suggest this:
1. Place the source-code directly into the /var/www/html directory with the right owner
   1. I'd argue the owner of most files should be e.g. root, as to prevent any in-place upgrades/changes of source-code from a security perspective
   1. This also means the automated updates of nextcloud should be disabled, and instead, the Docker way of upgrading should be used (docker --pull). This is good, as it means that any system security problems for the underlying OS and dependancies are also patched. (FROM php, FROM ubuntu,...)
   1. You also don't need to install rsync..  
1. Define VOLUMEs for all directories containing user-content or dynamic data. I believe this to be the /data/ directory, the /config directory and the /app directory?
   1. The end-user can always override these VOLUMEs by redefining them in their docker-compose or docker run cmd
1. Define VOLUMEs for directories containing static data, like .js and .css files. 
   1. The end user could either use volumes_from, or preferably, override this VOLUME with a named-volume and mount that volume as read-only in their nginx 
1. Make an entrypoint script which checks the installed version of nextcloud (should be possible via either config.php or DB), and if required, do an automated occ:upgrade.
1. Possibly also make the entrypoint-script enable automatic installation of nextcloud by reading ENV variables (though  I'd personally not use it due to security concerns with exposing credentials like that, some might like it)

tilosp:
@ChessSpider
but all of those had to be recreated if you want to update, or am i wrong?

ChessSpider:
@tilosp When you do a docker --pull to update a container it should reuse the VOLUMEs of the previous version of the container, right? I'm assuming here honestly, as I always declare my VOLUMEs explicitly in the docker-compose..

tilosp:
@ChessSpider Yes but a newly created container will not overwrite files in a volume it will just add new ones.

ChessSpider:
What I propose is something in the direction of:

```
FROM php:5.6-apache
RUN apt-get install........ etc
RUN tar -xjf nextcloud.tar.bz2 -C /var/www/html

# Volumes for dynamic data
VOLUME /var/www/html/config
VOLUME /var/www/html/data
VOLUME /var/www/html/custom_apps
# volumes for template purposes or static files which are to be served by nginx in a fpm environment
VOLUME /var/www/html/vendor
VOLUME /var/www/html/css
VOLUME /var/www/html/static all files...
```

wait, what, why are the static files all over the place? Are they not collected somewhere in one directory ?

```
root@1a83c9988257:/var/www/html# find . -name '*.css' 
./core/vendor/select2/select2.css
./core/vendor/strengthify/strengthify.css
./core/vendor/jcrop/css/jquery.Jcrop.css
./core/vendor/jquery-ui/themes/base/jquery-ui.css
./core/css/global.css
./core/css/mobile.css
./core/css/fixes.css
./core/css/tooltip.css
./core/css/inputs.css
./core/css/share.css
./core/css/jquery.ocdialog.css
./core/css/jquery-ui-fixes.css
./core/css/header.css
./core/css/fonts.css
./core/css/multiselect.css
./core/css/icons.css
./core/css/styles.css
./core/css/apps.css
./core/css/lostpassword/resetpassword.css
./core/css/systemtags.css
./core/css/update.css
./core/doc/user/_static/bootstrap-3.1.0/css/bootstrap-theme.css
./core/doc/user/_static/bootstrap-3.1.0/css/bootstrap.css
./core/doc/user/_static/bootstrap-3.1.0/css/bootstrap.min.css
./core/doc/user/_static/bootstrap-3.1.0/css/bootstrap-theme.min.css
./core/doc/user/_static/basic.css
./core/doc/user/_static/pygments.css
./core/doc/user/_static/bootstrap.css
./core/doc/user/_static/main.min.css
./core/doc/user/_static/bootstrap-sphinx.css
./core/doc/user/_static/styles.css
./core/doc/user/_static/bootstrap-responsive.css
./core/doc/user/_static/style.css
./core/doc/admin/_static/bootstrap-3.1.0/css/bootstrap-theme.css
./core/doc/admin/_static/bootstrap-3.1.0/css/bootstrap.css
./core/doc/admin/_static/bootstrap-3.1.0/css/bootstrap.min.css
./core/doc/admin/_static/bootstrap-3.1.0/css/bootstrap-theme.min.css
./core/doc/admin/_static/basic.css
./core/doc/admin/_static/pygments.css
./core/doc/admin/_static/bootstrap.css
./core/doc/admin/_static/main.min.css
./core/doc/admin/_static/bootstrap-sphinx.css
./core/doc/admin/_static/styles.css
./core/doc/admin/_static/bootstrap-responsive.css
./core/doc/admin/_static/style.css
./core/search/css/results.css
./settings/css/settings.css
./3rdparty/sabre/dav/lib/DAV/Browser/assets/openiconic/open-iconic.css
./3rdparty/sabre/dav/lib/DAV/Browser/assets/sabredav.css
./themes/example/core/css/styles.css
./apps/password_policy/css/settings-admin.css
./apps/encryption/css/settings-personal.css
./apps/encryption/css/settings-admin.css
./apps/files_pdfviewer/vendor/pdfjs/web/viewer.css
./apps/files_pdfviewer/css/viewer.css
./apps/files_pdfviewer/css/minmode.css
./apps/files_pdfviewer/css/style.css
./apps/gallery/css/mobile.css
./apps/gallery/css/github-markdown.css
./apps/gallery/css/slideshow.css
./apps/gallery/css/public.css
./apps/gallery/css/upload.css
./apps/gallery/css/gallerybutton.css
./apps/gallery/css/share.css
./apps/gallery/css/styles.css
./apps/gallery/css/authenticate.css
./apps/gallery/css/error.css
./apps/files_external/css/external.css
./apps/files_external/css/settings.css
./apps/files_videoplayer/css/style.css
./apps/files_videoplayer/videojs/src/video-js.css
./apps/workflowengine/css/admin.css
./apps/external/css/style.css
./apps/user_ldap/vendor/ui-multiselect/jquery.multiselect.css
./apps/user_ldap/css/settings.css
./apps/files_texteditor/css/mobile.css
./apps/files_texteditor/css/DroidSansMono/stylesheet.css
./apps/files_texteditor/css/style.css
./apps/files_sharing/css/mobile.css
./apps/files_sharing/css/public.css
./apps/files_sharing/css/sharetabview.css
./apps/files_sharing/css/authenticate.css
./apps/files_sharing/css/sharebreadcrumb.css
./apps/files_sharing/css/sharedfilelist.css
./apps/files_sharing/css/404.css
./apps/files/css/mobile.css
./apps/files/css/upload.css
./apps/files/css/files.css
./apps/files/css/detailsView.css
./apps/serverinfo/css/style.css
./apps/user_saml/css/personal.css
./apps/user_saml/css/admin.css
./apps/files_versions/css/versions.css
./apps/templateeditor/css/settings-admin.css
./apps/theming/css/settings-admin.css
./apps/files_retention/css/retention.css
./apps/logreader/css/app.css
./apps/logreader/build/main.css
./apps/federation/css/settings-admin.css
./apps/activity/css/settings.css
./apps/activity/css/style.css
./apps/survey_client/css/admin.css
./apps/comments/css/comments.css
./apps/twofactor_backupcodes/css/style.css
./apps/files_trashbin/css/trash.css
./apps/federatedfilesharing/css/settings-personal.css
./apps/systemtags/css/systemtagsfilelist.css
./apps/firstrunwizard/css/colorbox.css
./apps/firstrunwizard/css/firstrunwizard.css
./apps/notifications/css/styles.css 
```

ChessSpider:
just to clarify, the simplified docker-compose for my nginx and app of a project of mine is this:
```
  version: '2'

  networks:
    net:
      driver: bridge
  services:
    pulsejam_web:
      ports:
        - "8000:8000"
      image: nginx
      user: www-data
      volumes:
        - ./nginx.conf:/etc/nginx/nginx.conf:ro
        - ./app/static:/var/www/html/static/:ro
        - ./app/dynamic:/var/www/html/dynamic/:ro
      tmpfs:
        - /var/cache/nginx/:uid=33,rw,noexec,nosuid
        - /run:uid=33,rw,noexec,nosuid
      security_opt:
        - no-new-privileges
      cap_drop:
        - net_raw
        - mknod
      links:
        - app
      networks:
        - net
    app:
      image: app
      user: lowprivuser
      links:
        - db
      volumes:
        - ./app/static/:/usr/src/app/static/
        - ./app/dynamic/:/usr/src/app/dynamic/
      security_opt:
        - no-new-privileges
      cap_drop:
        - net_raw
        - mknod
      tmpfs:
        - /tmp:rw,noexec,nosuid
      networks:
        - net
```

tilosp:
@ChessSpider 
if you run these commands you can see that the content of the volume is not getting updated.
```bash
$ sudo docker run --rm -v test:/usr/src/nextcloud nextcloud:9 bash -c exit
$ sudo docker run --rm -v test:/usr/src/nextcloud nextcloud:10 bash -c exit
$ sudo docker run --rm -v test:/test debian cat /test/version.php
<?php 
$OC_Version = array(9,0,57,2);
$OC_VersionString = '9.0.57';
```
Your static files volumes would have the same problem.

ChessSpider:
Yeah but that's with the old VOLUME declaration on /var/www/html right? So that makes sense.

I agree that static files volume would have that behavior, but it would only be a problem while downgrading. Which we don't support anyway. When upgrading, old static files will be overwritten. Deleted static files will not be referenced in the HTML and thus should not be a problem. 

Also static files should only have to be a VOLUME for the FPM-usecase.

I do worry about the static files being scattered across various folders. That does make it a lot less clear which folders should be a VOLUME and which ones should not. There's really no need at all to mount folders with php files in the nginx docker container when using fpm.

tilosp:
@ChessSpider no the Dockerfile places the source-code directly into `/usr/src/nextcloud'. A container will not overwrite the files in a volume when started, this is the reason why we copy them into the volume in the entry-point script.

ChessSpider:
@tilosp yeah, but that's why https://github.com/nextcloud/docker/pull/23 and others are suggesting to remove the volume on /var/www/html right? 
Im also argueing to remove the too broad volume on /var/www/html, and replace it with a volume with a smaller scope: the directory/directories with static files.  And only for the fpm image. 
Are you saying that is not feasible and we need to have a broad volume on the entire /var/www/html directory?

tilosp:
@ChessSpider i'am saying if we use volumes with a smaller scope, we would still have to update the content of the small scope volumes. And this would then again require us to copy the files into the volume. I'am perfectly fine with removing the volume declaration completely and let the user decide which scope they want to use.

ChessSpider:
@tilosp Right, we're finally aligned :).  You're correct about the static files. I think the VOLUME on /var/www/html needs to be removed either way. You want users to upgrade using PULL instead of using the nextcloud updater, and you want users to get the correct version when using the docker updater. 

A different solution has to be figured out for the fpm use-case though. Maybe some kind of asset management is possible? Previously I'd say to re-run the asset pipelining feature, and put the VOLUME on the assets/ folder. But that's gone. 

Do you have a suggestion? Is there a feasible way to concentrate static files in one directory, or is there a limited number of directories which contain static files? 

Alternatively, we can provide clear instructions which previous volumes has to be removed and recreated in case you want to use it in a php-fpm setup. The neighbors at Wonderfall unfortunately don't have this problem as they run php-fpm alongside nginx inside the container https://github.com/Wonderfall/dockerfiles/blob/master/nextcloud/11.0/Dockerfile

10.0/apache/docker-entrypoint.sh
```diff
@@ -1,9 +1,45 @@
 #!/bin/bash
 set -e
 
-if [ ! -e '/var/www/html/version.php' ]; then
-    tar cf - --one-file-system -C /usr/src/nextcloud . | tar xf -
+# version_greater A B returns whether A > B
+function version_greater() {
+	[[ "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1" ]];
```
pierreozoux: I just tried this on an alpine box, and the busybox sort is complaining about the `V`...

I ended up using this:
https://github.com/pierreozoux/docker-alpine/blob/03573c088fc71bc9bd6a6e0b4af262935289621c/builder/scripts/apk-install#L4-L6

cyphar: Yeah, that's because `sort -V` is a GNU coreutils thing. I really am not a fan of that script you linked, GNU's version checking is much more sophisticated. Are we basing our images on `alpine`? I feel like we shouldn't because of the other issues that Alpine container images have...

cyphar:
@ChessSpider @tilosp Removing `VOLUME` isn't going to help users who want to use Docker's volume drivers, it'll mean that they can't use any such drivers. While I understand the worry about the copying code, I really don't agree with the idea that we should make the default usage of our images be **unable to extract the data in a sane way so that it is persistent**. If you look at how other images work, (for example [mysql](https://github.com/docker-library/mysql/blob/master/8.0/Dockerfile)) you'll notice that they use `VOLUME` for its intended purpose -- defining which parts of a container's filesystem should be persistent across restarts.

From my experience, splitting up the `VOLUME` causes **nothing but headaches**. Right now I have a NextCloud install that I need to purge with fire because of the issues I've had with upgrading it -- some but not all of the files get deleted properly (for example). It's a mess.

ChessSpider:
@cyphar A user can always create a volume on startup with ```docker run -v``` , regardless of any VOLUME declaration in the Dockerfile? Removing all VOLUME declarations should not limit the user in any way. 

The MySQL Dockerfile you reference only put a VOLUME on the ```/var/lib/mysql``` folder. That is  equivalent of the ```data```, ```custom_app``` and ```config``` directory of Nextcloud. Nextcloud should probably do that too.

10.0/apache/docker-entrypoint.sh
pierreozoux: Actually, we can install coreutils in alpine too.
https://wiki.alpinelinux.org/wiki/How_to_get_regular_stuff_working

What are the other issues?

I really like alpine, and as free software advocates, I believe it is our mission to make this IT world a better place. Alpine is a lot more ecological, that's it.

10.0/apache/docker-entrypoint.sh
tilosp: done

tilosp:
should be ready for merge

ChessSpider:
I'm not so sure.. with the removal of the VOLUME you might get a lot of angry users if you don't tell them they have to manually set a VOLUME on certain directories. IMHO I'd actually add a VOLUME for the config, data, and custom_apps folder. Everybody would want this, and one can always override it in the ```docker run```

and without a solution for fpm users you might get a lot of angry people too, as it now breaks existing functionality. This needs to be really well communicated. 

I really do like the changes though, it is definitely an improvement on what it was.

tilosp:
@pierreozoux @cyphar can we get this merged before [11.0.3](https://github.com/nextcloud/server/releases/tag/v11.0.3) is released?

tianon:
Looks generally OK to me, although I'd feel better if @yosifkit took a look as well.

My only question after the look I did was whether there are any technical reasons downgrading isn't supported ATM? (not thinking anything here should change, more curious whether there's some indicator that could possibly be used in the future for determining data compatibility when doing upgrades/downgrades so that this could potentially be expanded :+1:)

yosifkit:
Seems fine to me!  Seems like we should be able to do something similar in the WordPress image. :heart_eyes:

SnowMB:
> My only question after the look I did was whether there are any technical reasons downgrading isn't supported ATM? (not thinking anything here should change, more curious whether there's some indicator that could possibly be used in the future for determining data compatibility when doing upgrades/downgrades so that this could potentially be expanded 👍)

I think this would be something for the guys working on the server as this is not docker specific 😉

cyphar:
/me will test this out soon once I figure out how much of my current data might be FUBAR.

PR #19, opened 2016-11-26
Add php-imap extension for imap based login

yadutaf:
This PR adds php-imap extension to the Docker image so that it can be used to integrate with existing mail servers. I use it to have the same login / password on my existing private mail server and my nextcloud instance.

PR #66, opened 2017-04-16
Enable PHP IMAP extension

mkuron:
This needed for Nextcloud to authenticate against an IMAP server using the _user_external_ app. The app is already included in the image, but the necessary PHP extension is added by this pull request.


mkuron:
Ah, sorry. Looks like there already was #19 about the same thing which wasn't merged because there was a discussion at @owncloud still going on about this. I think that the ~1MB this adds to the image are reasonable (unlike SMB support for _user_external_, which takes up ~60MB) and comparable to what other PHP modules add that are not needed in every installation (e.g. ldap). So perhaps this could be considered again.

PR #73, opened 2017-05-12
Added libpcre3-dev to make pcre.h available

gpkvt:
Without pcre.h the build process will fail with the latest php:7.1-apache base image


tilosp:
It's no longer failing. https://travis-ci.org/nextcloud/docker/builds
I'am not sure why it failed that day.

PR #82, opened 2017-06-01
Update README.md

gauravarora:
volumes_from instead of volumes-from in docker-compose.yml


tilosp:
Thanks a lot for your work!

PR #84, opened 2017-06-02
Added 12.0 for fpm

xsteadfastx:
Changed opcache options to versions recommendations.

### DIFF
--- a/12.0/fpm/Dockerfile
+++ b/12.0/fpm/Dockerfile
@@ -1,0 +1,73 @@
+FROM php:7.1-fpm
+
+RUN apt-get update && apt-get install -y \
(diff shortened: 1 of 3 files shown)


tilosp:
The Dockerfiles in the sub directories are auto generated by [`update.sh`](https://github.com/nextcloud/docker/blob/master/update.sh). So if you want to change them you need to change [`update.sh`](https://github.com/nextcloud/docker/blob/master/update.sh). I have already done this in #85.

xsteadfastx:
@tilosp what's with the opcache options?

tilosp:
Are these opcache options only for nextcloud 12?

PR #100, opened 2017-06-12
[chore] dev env dockerfile draft

m4dz:
Hey Nexclouders!

for quick start development, I've made a `Dockerfile` to build a development docker environment. It's a first draft, based on the `apache` version that I rewrote a little bit. I now need your inputs to find how we can improve it :smiley:. So I have few questions:

1. yet, it excludes some PHP modules like
  - `smbclient`, `ftp` and `imap`
  - `gmp`
  - `memcached` and `redis`
  - `imagick` and the preview deps tools (`ffmpeg`, `avconv`, `LibreOffice`)
should we consider, as the image if for development purposes, that we need to provide them, or is it too much?

2. It actually uses a `sqlite` db for storage, which is fine for development I guess, but once again, if this image is used for development, we maybe need to also embed a MariaDB and a PgSQL, or at least give a simple way to compose a docker image with those DB, isn't it?

3. I see that the dockerfiles are generated from templates. I so could contribute to them for the `nextcloud:dev` dockerfile so I can be generated as well. Do it sounds good?

4. We can actually run it with `cd nextcloud/server && docker --rm -it -p 8080:80 -v $(pwd):/var/www/html nextcloud:dev`, but even if my proposal is a little bit different than the `apache` version, maybe the latter is sufficient for dev purposes, so this PR isn't really relevant. Any thoughts about it?

Thanks for your opinions!

### DIFF
--- a/12.0/dev/Dockerfile
+++ b/12.0/dev/Dockerfile
@@ -1,0 +1,66 @@
+FROM php:7.1-apache
+
+RUN set -ex \
+  && buildDeps=" \
+    rsync \
+    libbz2-dev \
+    libcurl4-openssl-dev \
+    libfreetype6-dev \
+    libicu-dev \
+    libjpeg-dev \
+    libldap2-dev \
+    libmcrypt-dev \
+    libmemcached-dev \
+    libpng12-dev \
+    libpq-dev \
+    libxml2-dev \
+  " \
+  && apt-get update \
+  && apt-get install -y $buildDeps --no-install-recommends \
+  && rm -rf /var/lib/apt/lists/*
+
+# https://docs.nextcloud.com/server/12/admin_manual/installation/source_installation.html#prerequisites-for-manual-installation
+RUN set -ex \
+  && phpModules=" \
+    bz2 \
+    curl \
+    gd \
+    exif \
+    intl \
+    mbstring \
+    mcrypt \
+    ldap \
+    mysqli \
+    opcache \
+    pdo_mysql \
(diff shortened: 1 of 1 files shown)


SnowMB:
In my opinion a dev docker file belongs directly to the server repo. There was a pull request a while ago, but no reaction to change requests so it's still open. See https://github.com/nextcloud/server/pull/4355.

> It actually uses a sqlite db for storage, which is fine for development I guess, but once again, if this image is used for development, we maybe need to also embed a MariaDB and a PgSQL, or at least give a simple way to compose a docker image with those DB, isn't it?

Just run a plain database container. What's the problem with it? It's documented in the examples.

m4dz:
Thanks @SnowMB! Fine for running a plain database container, seems legit, just wanna ask for any opinions, but it's also my favorite choice. About hosting the Dockerfile in the `server` repo, I'm not sure: IMHO nextcloud can provide an already built image at `nextcloud:dev` as it provides a `nexcloud:fpm` image. I don't think any developer have to build its own dev image, nope?

tilosp:
I don't think that the dev image should be part of the official docker repo. Images on there should be "generally useful" and i don't think that the dev image is "generally useful". 

If you want to provide a prebuild image we could publish it to a docker hub organization.

cosenal:
@m4dz I would encourage you to contribute to my PR instead, https://github.com/nextcloud/server/pull/4355

If you see the comments there, there is an agreement that a Docker setup belongs to the main repo.

PR #113, opened 2017-06-25
Add multi arch support

tilosp:


tilosp:
@tianon is this all we need to do to support multiple architectures?

tianon:
Looks OK at a quick glance, but I'll cc @yosifkit too (since I'm on PTO this week).

Has the build been verified against at least i386? (Adding `i386/` to the `FROM` line is probably the easiest way to test that.  Any other build testing you can do is also appreciated, but we can always trim this list back down if the builds do end up failing. 👍)

tilosp:
@tianon is there an easy way to fix this `docker-php-ext-configure ldap --with-libdir=lib/x86_64-linux-gnu`?
`uname -m` returns `x86_64` when i run `docker run -it --rm i386/php uname -m`

tilosp:
@tianon I added travis tests for i386

10.0/apache/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN apt-get update && apt-get install -y \
 
 # https://docs.nextcloud.com/server/9/admin_manual/installation/source_installation.html
 RUN docker-php-ext-configure gd --with-png-dir=/usr --with-jpeg-dir=/usr \
-  && docker-php-ext-configure ldap --with-libdir=lib/x86_64-linux-gnu \
+  && docker-php-ext-configure ldap --with-libdir=lib/*-linux-gnu \
```
yosifkit: Rather than the `*-linux-gnu`, you may want to use something similar to [httpd](https://github.com/docker-library/httpd/blob/d5f06d182e52822b281e9c5f992309a281a756a7/2.4/Dockerfile#L91) or [php](https://github.com/docker-library/php/blob/3ee1cd8287f6fc7269a731ceb6199ec8dee8b727/Dockerfile-debian.template#L98-L99). (bonus is that `dpkg-dev` is already installed in the Debian-based php images that you are based on).

It looks like you would need to use the `debMultiarch="$(dpkg-architecture --query DEB_BUILD_MULTIARCH)"` line from [php](https://github.com/docker-library/php/blob/3ee1cd8287f6fc7269a731ceb6199ec8dee8b727/Dockerfile-debian.template#L122).

tilosp: done

tilosp:
@yosifkit can i merge this now?

yosifkit:
LGTM!

PR #115, opened 2017-06-28
Copy config, data, custom_apps, and themes to volume when empty

vfreex:
When Nextcloud performs an upgrade or clean installation,
it will check whether /var/www/html/{config,data,custom_apps,themes} exist.
If not, it will copy
/usr/src/nextcloud/{config,data,custom_apps,themes} to /var/www/html.

This leads to a problem: If those subdirectories are existent but
empty, it will not do the copy. This situation is common when you mount
volumes to those subdirectories, like:

```
version: "2.1"
services:
    app:
        image: nextcloud:12-apache
        volumes:
            - nextcloud:/var/www/html:Z
            - nextcloud-custom_apps:/var/www/html/custom_apps:Z
            - nextcloud-config:/var/www/html/config:Z
            - nextcloud-data:/var/www/html/data:Z
            - nextcloud-themes:/var/www/html/themes:Z
        ports:
            - 8080:80/tcp
    db:
        image: mariadb
        volumes:
            - db:/var/lib/mysql:Z
        environment:
            MYSQL_USER: nextcloud
            MYSQL_DATABASE: nextcloud
            MYSQL_PASSWORD: nextcloud
            MYSQL_ROOT_PASSWORD: nextcloud

volumes:
    nextcloud:
    nextcloud-custom_apps:
    nextcloud-config:
    nextcloud-data:
    nextcloud-themes:
    db:
```

This patch will fix this issue by checking whether those subdirectories
are empty.

### DIFF
--- a/10.0/apache/docker-entrypoint.sh
+++ b/10.0/apache/docker-entrypoint.sh
@@ -6,6 +6,12 @@ function version_greater() {
 	[[ "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1" ]];
 }
 
+# return true if specified directory is empty
+function directory_empty() {
+    [ -n "$(find "$1"/ -prune -empty)" ]
+}
+
+
 installed_version="0.0.0~unknown"
 if [ -f /var/www/html/version.php ]; then
     installed_version=$(php -r 'require "/var/www/html/version.php"; echo "$OC_VersionString";')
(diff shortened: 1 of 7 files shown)


tilosp:
Thanks a lot for your work!

10.0/apache/docker-entrypoint.sh
```diff
@@ -21,25 +27,13 @@ if version_greater "$image_version" "$installed_version"; then
...
-        cp -arT /usr/src/nextcloud/config /var/www/html/config
-    fi
-
-    if [ ! -d /var/www/html/data ]; then
-        cp -arT /usr/src/nextcloud/data /var/www/html/data
-    fi
-
-    if [ ! -d /var/www/html/custom_apps ]; then
-        cp -arT /usr/src/nextcloud/custom_apps /var/www/html/custom_apps
-        cp -a /usr/src/nextcloud/config/apps.config.php /var/www/html/config/apps.config.php
```
menghan: This pull request breaks this line: if /var/www/html/config exists but /var/www/html/custom_apps doesn't, config/apps.config.php don't get copied

vfreex: Thanks for the review. I'll fix that.

PR #118, opened 2017-06-29
Ensure config/apps.config.php get copied

vfreex:
PR #115 breaks the logic that config/apps.config.php get copied
after custom_apps: https://github.com/nextcloud/docker/pull/115#discussion_r124702790.

This patch is going to copy that file if it doesn't exist.

### DIFF
--- a/10.0/apache/docker-entrypoint.sh
+++ b/10.0/apache/docker-entrypoint.sh
@@ -35,6 +35,10 @@ if version_greater "$image_version" "$installed_version"; then
         fi
     done
 
+    if [ ! -f /var/www/html/config/apps.config.php ]; then
+        cp -a /usr/src/nextcloud/config/apps.config.php /var/www/html/config/apps.config.php
+    fi
+
     if [ "$installed_version" != "0.0.0~unknown" ]; then
         su - www-data -s /bin/bash -c 'php /var/www/html/occ upgrade --no-app-disable'
 
--- a/10.0/fpm/docker-entrypoint.sh
+++ b/10.0/fpm/docker-entrypoint.sh
@@ -35,6 +35,10 @@ if version_greater "$image_version" "$installed_version"; then
         fi
     done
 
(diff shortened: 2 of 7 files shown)


SnowMB:
Thanks for the quick fix!

tilosp:
@vfreex @SnowMB this cp is there to upgrade existing setups which were created before #65. 
I thought it would be safe to remove it because most of the setups are upgraded by now. 
But maybe i am wrong. 
If we want to keep it we should use the old check. Because right now the file will we recreated in case the user decides to remove it.

vfreex:
@tilosp 
By default `custom_apps` is empty if the user haven't installed any apps. In that case, the script don't know if the config file has been deleted by the user or it hasn't been copied. So the old check doesn't work.
I would suggest creating an empty file to prevent from copying, but the user should learn about that.

tilosp:
I do not want to check if the `custom_apps` folder is empty. In old setups there simply is no custom apps folder.
```bash
for dir in config data themes; do
    if [ ! -d /var/www/html/"$dir" ] || directory_empty /var/www/html/"$dir"; then
        cp -arT /usr/src/nextcloud/"$dir" /var/www/html/"$dir"
    fi
done
if [ ! -d /var/www/html/custom_apps ]; then
    cp -a /usr/src/nextcloud/config/apps.config.php /var/www/html/config/apps.config.php
fi
if [ ! -d /var/www/html/custom_apps ] || directory_empty /var/www/html/custom_apps; then
    cp -arT /usr/src/nextcloud/custom_apps /var/www/html/custom_apps
fi
```

vfreex:
@tilosp 
But if I mount `custom_apps` from a volume, it will not copy `apps.config.php`, resulting in a permission error because `apps` is not writable by default in this image.

vfreex:
Sorry, my mistake. This situation only happens when upgrading from an old version because in newer version the whole `config` will be copied.
Your solution looks better than mine.

tilosp:
I implemented this in #119

PR #119, opened 2017-06-29
Ensure apps.config.php get only copied when upgrading from old version

tilosp:


10.0/apache/docker-entrypoint.sh
```diff
@@ -29,16 +29,20 @@ if version_greater "$image_version" "$installed_version"; then
...
     
-    for dir in config data custom_apps themes; do
+    for dir in config data themes; do
         if [ ! -d /var/www/html/"$dir" ] || directory_empty /var/www/html/"$dir"; then
             cp -arT /usr/src/nextcloud/"$dir" /var/www/html/"$dir"
         fi
     done
 
-    if [ ! -f /var/www/html/config/apps.config.php ]; then
+    if [ ! -d /var/www/html/custom_apps ]; then
```
vfreex: I think `if [ ! -d /var/www/html/custom_apps ] && [ ! -f /var/www/html/config/apps.config.php ]; then` would be better because we don't want to overwrite the file in any case.

tilosp: done

vfreex:
LGTM

PR #121, opened 2017-07-01
Enable APCu

### LINKED ISSUE #109: Enable APCu
Nextcloud wants to have some sort of memory caching enabled: [docs](https://docs.nextcloud.com/server/12/admin_manual/configuration_server/caching_configuration.html). APCu is already installed in the image and ready to be used, it just needs to be activated in the Nextcloud configuration:

```
'memcache.local' => '\OC\Memcache\APCu',
```

I propose to have this pre-configured in the image, just like the app paths in `apps.config.php`. I'd add a new file `apcu.config.php` and add this setting there. What do others think about this proposal?

```
<?php
$CONFIG = array (
  memcache.local' => '\OC\Memcache\APCu',
);
```

tilosp:
closes #109


10.0/apache/Dockerfile
```diff
@@ -44,6 +44,8 @@ RUN a2enmod rewrite
 ENV NEXTCLOUD_VERSION 10.0.5
 VOLUME /var/www/html
 
+COPY config/* /usr/src/nextcloud/config/
```
tianon: `COPY` has annoying and often inconsistent cache behavior -- any particular reason this needs to be moved up here and can't stay down below with the other `COPY`?

tilosp: Yes because the permissions for this file are set in the RUN block below.

tilosp:
@tianon should i move it down again and add an additional RUN afterwards to set the permissions

PR #124, opened 2017-07-05
Example: remove X-Frame-Options from nginx.conf

Chewie:
According to https://github.com/nextcloud/server/issues/4605, The X-Frame-Options should be dropped from nginx to let the php processing handle the case. On nextcloud 12, having that header in Nginx causes a warning in the admin panel.


SnowMB:
Thanks for this PR! LGTM @tilosp, @pierreozoux ?

tilosp:
Thank you very much for your contribution!

PR #131, opened 2017-07-17
update directory permissions to be compatible with non-root

### LINKED ISSUE #69: fpm container file ownership and permission changes
The nextcloud:11.0.3-fpm container has changed the ownership and permissions of the files in /var/www/html.  This prevents my nginx container that has mounted the /var/www/html volume from accessing the files:
> 017/05/07 02:45:14 [crit] 7#7: *1718 stat() "/var/www/html/status.php" failed (13: Permission denied),

The files in nextcloud:11.0.3-fpm are owned by root:www-data and are not world readable:
> $ docker run -it --rm nextcloud:11.0.3-fpm ls -lha /var/www/html/version.php
> -rw-r----- 1 root www-data 380 Apr 22 06:29 /var/www/html/version.php

Previously (<=nextcloud:11.0.2-fpm) the files were owned by www-data:nogroup and were world readable.
> $ docker run -it --rm nextcloud:11.0.2-fpm ls -lha /var/www/html/version.php
> -rw-r--r-- 1 www-data nogroup 277 Feb 26 19:44 /var/www/html/version.php

Was this changed intended?

### LINKED ISSUE #107: Directory permissions for running Nextcloud without root
To make the official Nextcloud Docker image compatible with OpenShift out-of-the-box, some directory permissions need to be amended. For more information about this topic, see [OpenShift Docs -> Creating Images -> Guidelines -> OpenShift Container Platform-Specific Guidelines -> Support Arbitrary User IDs](https://docs.openshift.com/container-platform/3.5/creating_images/guidelines.html#openshift-container-platform-specific-guidelines). The important part:

> For an image to support running as an arbitrary user, directories and files that may be written to by processes in the image should be owned by the root group and be read/writable by that group

```
RUN chgrp -R 0 /some/directory \
  && chmod -R g+rwX /some/directory
```

This means all directories which get writen into should have these permissions. At least

* `/usr/src/nextcloud/custom_apps`
* `/usr/src/nextcloud/config`
* `/usr/src/nextcloud/data`
* `/usr/src/nextcloud/themes`

A working Dockerfile can be found here: https://github.com/tobru/nextcloud-openshift/blob/dev/Dockerfile. With these changes the Nexcloud image runs perfect on OpenShift.

Are the maintainers of the official Dockerfiles willing to incorporate the proposed permission change? I'm happy to create a PR, but wanted to first discuss this topic. My wish is to bring Nextcloud to OpenShift without having to customize the official Docker image as done in https://github.com/tobru/nextcloud-openshift/blob/dev/Dockerfile

### LINKED ISSUE #143: nextcloud:12-apache: Maximal upload size not changeable & persisted
In order to change the maximal upload size via the UI (Admin -> Additional Settings -> File handling), Nextcloud must have write access to the `.htaccess` (and the `.user.ini`) in the webroot directory. In the container, the file belongs to `root:www-data`, but is 0x640 so it can't be modified by the apache process.

Also, as being part of the image, the modified `.htaccess` doesn't survive container re-creation.

I understand that this settings needs to be part of the `.htaccess`, but the information should be stored somewhere else too and written to `.htaccess` during container launch.

tobru:
This commit updates the directory permissions to be more compatible when
running the image without root f.e. on OpenShift or when specifying it
when running with `docker run --user www-data:root ...`.
It adds detection logic to the entrypoint script as sudo is not always
allowed.

This change in directory permissions was also proposed by the official
documentation, see https://github.com/nextcloud/documentation/commit/22e2530.

The `chown` before the volume definition is needed to prepare the volume
as it inherits the permissions.

Closes #107 Closes #143 Closes #69

### DIFF
--- a/10.0/apache/Dockerfile
+++ b/10.0/apache/Dockerfile
@@ -46,8 +46,6 @@ ENV NEXTCLOUD_VERSION 10.0.6
 RUN chown -R www-data:root /var/www/html
 VOLUME /var/www/html
 
-COPY config/* /usr/src/nextcloud/config/
-
 RUN curl -fsSL -o nextcloud.tar.bz2 \
     "https://download.nextcloud.com/server/releases/nextcloud-${NEXTCLOUD_VERSION}.tar.bz2" \
  && curl -fsSL -o nextcloud.tar.bz2.asc \
@@ -65,6 +63,7 @@ RUN curl -fsSL -o nextcloud.tar.bz2 \
  && chmod +x /usr/src/nextcloud/occ
 
 COPY docker-entrypoint.sh /entrypoint.sh
+COPY config/* /usr/src/nextcloud/config/
 
 ENTRYPOINT ["/entrypoint.sh"]
 CMD ["apache2-foreground"]
--- a/10.0/apache/Dockerfile
+++ b/10.0/apache/Dockerfile
@@ -43,7 +43,8 @@ RUN a2enmod rewrite
 
 ENV NEXTCLOUD_VERSION 10.0.6
 
-RUN chown -R www-data:root /var/www/html
+RUN chown -R www-data:root /var/www/html && \
+    chmod -R g=u /var/www/html
 VOLUME /var/www/html
 
 RUN curl -fsSL -o nextcloud.tar.bz2 \
--- a/10.0/apache/Dockerfile
+++ b/10.0/apache/Dockerfile
@@ -42,6 +42,8 @@ RUN set -ex \
 RUN a2enmod rewrite
 
 ENV NEXTCLOUD_VERSION 10.0.6
+
+RUN chown -R www-data:root /var/www/html
 VOLUME /var/www/html
 
 COPY config/* /usr/src/nextcloud/config/
@@ -58,16 +60,8 @@ RUN curl -fsSL -o nextcloud.tar.bz2 \
  && tar -xjf nextcloud.tar.bz2 -C /usr/src/ \
  && rm nextcloud.tar.bz2 \
  && rm -rf /usr/src/nextcloud/updater \
- # https://docs.nextcloud.com/server/11/admin_manual/installation/installation_wizard.html#setting-strong-directory-permissions
  && mkdir -p /usr/src/nextcloud/data \
  && mkdir -p /usr/src/nextcloud/custom_apps \
- && find /usr/src/nextcloud/ -type f -print0 | xargs -0 chmod 0640 \
- && find /usr/src/nextcloud/ -type d -print0 | xargs -0 chmod 0750 \
- && chown -R root:www-data /usr/src/nextcloud/ \
- && chown -R www-data:www-data /usr/src/nextcloud/custom_apps/ \
- && chown -R www-data:www-data /usr/src/nextcloud/config/ \
- && chown -R www-data:www-data /usr/src/nextcloud/data/ \
- && chown -R www-data:www-data /usr/src/nextcloud/themes/ \
  && chmod +x /usr/src/nextcloud/occ
 
 COPY docker-entrypoint.sh /entrypoint.sh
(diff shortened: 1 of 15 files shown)


tobru:
@tilosp What do you think about this PR? I'm open to discuss the possibilities and changes, I would love to see the official Nextcloud Image to run on platforms like OpenShift which require containers to run without root permissions.

tilosp:
@tobru it looks like this change breaks the image when running as root
```shell
$ sudo docker run -it --rm nexttest:10-apache ls -l /var/www/html
total 136
drwxr-xr-x. 1 nobody nogroup   562 Apr 22 06:23 3rdparty
-rw-r--r--. 1 nobody nogroup  8868 Apr 22 06:20 AUTHORS
-rw-r--r--. 1 nobody nogroup 34520 Apr 22 06:20 COPYING-AGPL
drwxr-xr-x. 1 nobody nogroup   848 Apr 22 06:21 apps
drwxr-xr-x. 1 nobody nogroup   112 Apr 22 06:20 config
-rw-r--r--. 1 nobody nogroup  3949 Apr 22 06:20 console.php
drwxr-xr-x. 1 nobody nogroup   332 Apr 22 06:23 core
-rw-r--r--. 1 nobody nogroup  5689 Apr 22 06:21 cron.php
drwxr-xr-x. 1 root   root        0 Jul 31 09:27 custom_apps
drwxr-xr-x. 1 root   root        0 Jul 31 09:27 data
-rw-r--r--. 1 nobody nogroup 39782 Apr 22 06:21 db_structure.xml
-rw-r--r--. 1 nobody nogroup   179 Apr 22 06:21 index.html
-rw-r--r--. 1 nobody nogroup  2628 Apr 22 06:21 index.php
drwxr-xr-x. 1 nobody nogroup    58 Apr 22 06:21 l10n
drwxr-xr-x. 1 nobody nogroup    94 Apr 22 06:21 lib
-rwxr-xr-x. 1 nobody nogroup   283 Apr 22 06:21 occ
drwxr-xr-x. 1 nobody nogroup    70 Apr 22 06:21 ocs
drwxr-xr-x. 1 nobody nogroup    18 Apr 22 06:21 ocs-provider
-rw-r--r--. 1 nobody nogroup  3184 Apr 22 06:21 public.php
-rw-r--r--. 1 nobody nogroup  5431 Apr 22 06:21 remote.php
drwxr-xr-x. 1 nobody nogroup    34 Apr 22 06:21 resources
-rw-r--r--. 1 nobody nogroup    26 Apr 22 06:21 robots.txt
drwxr-xr-x. 1 nobody nogroup   270 Apr 22 06:21 settings
-rw-r--r--. 1 nobody nogroup  1919 Apr 22 06:21 status.php
drwxr-xr-x. 1 nobody nogroup    26 Apr 22 06:21 themes
-rw-r--r--. 1 nobody nogroup   276 Apr 22 06:23 version.php
```
You can see directories like config, custom_apps,... are not writable by the web-server.

tilosp:
@tobru Can you test if it is still able to run on OpenShift.

tobru:
@tilosp Thanks for you efforts, will test it as soon as possible and report back!

tobru:
@tilosp I did another update, correcting the directory permissions so that members of the root group have write access. For more background information why this is needed, have a look at the [origin documentation](https://docs.openshift.org/latest/creating_images/guidelines.html#openshift-origin-specific-guidelines) (Support Arbitrary User IDs).

To test it locally, run it like this:

```
docker run --rm -ti -u 1000180000:root <image>
```

I tested both, fpm and apache, with the `-u` parameter and without. Both are imho now having the correct directory permission. The apache image doesn't work with an unprivileged user because Apache wants to bind to port 80 which of course only works with root rights - but I think this should be handled in a different issue as this one is all about directory permissions.

tianon:
Seems generally sane -- love the `Dockerfile` changes (anything that moves `COPY` down closer to the end is a big :+1: from me).  Usually we recommend against use of `su` since it has odd process management / TTY behavior, but in this case, it's only being used for one-off script runs in the entrypoint, not for the whole service, so I think it should be fine. :+1:

tobru:
Thanks a lot for merging this!! 👍

PR #147, opened 2017-08-28
Add autoconfig.php

tilosp:
Allows configuration of the database and the data directory via environment variables.

### DIFF
--- a/.config/autoconfig.php
+++ b/.config/autoconfig.php
@@ -22,3 +22,8 @@ if (getenv('SQLITE_DATABASE')) {
 if (getenv('NEXTCLOUD_DATA_DIR')) {
     $AUTOCONFIG["directory"] = getenv('NEXTCLOUD_DATA_DIR');
 }
+
+if (getenv('NEXTCLOUD_ADMIN_USER') && getenv('NEXTCLOUD_ADMIN_PASSWORD')) {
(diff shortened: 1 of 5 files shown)


tilosp:
@SnowMB this allows us to simplify the docker-compose examples

SnowMB:
Just tested this with mariadb and it works smoothly. 👍 

Can we add admin login and data directory to skip the installation process entirely?

```php
 "adminlogin"    => "root",
 "adminpass"     => "root-password",
 "directory"     => "/www/htdocs/nextcloud/data",
```

https://docs.nextcloud.com/server/12/admin_manual/configuration_server/automatic_configuration.html#all-parameters

tilosp:
I added environment variables for the admin login but they only work if you setup everything using environment variables.

@SnowMB
I would prefer to not set the data directory default in the `autoconfig.php`. Because this leaves the option to setup Nextcloud using the web gui.

SnowMB:
Maybe if we check before if any of the other autoconfig variables is present? If not we show the full installation page, otherwise use environment variables where possible and default values. Or introduce a variable to trigger the autoconfig like `AUTOCONFIG=true`?

I think for most users the ability to change the directory is not important and it could lead to data loss if the data directory is not part of a volume / mount.

PR #152, opened 2017-09-14
Pretty URLs for apache

### LINKED ISSUE #151: Hiding index.php in URL for Apache
Colud you please hide `index.php` in the URL in the Apache image by default.

tilosp:
Closes #151

### DIFF
--- a/.config/apache-pretty-urls.config.php
+++ b/.config/apache-pretty-urls.config.php
@@ -1,0 +1,4 @@
+<?php
+$CONFIG = array (
+  'htaccess.RewriteBase' => '/',
+);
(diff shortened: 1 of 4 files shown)


SnowMB:
Does not  work for me. Tried the same.

```
docker-compose exec -u 33:33 web /var/www/html/occ maintenance:update:htaccess
```
Has to be run after the installation (opening nextcloud via browser) to make it work.


Perhaps we can make it work when the autoconfig is present and use `occ  maintenance:install` ([link](https://docs.nextcloud.com/server/12/admin_manual/installation/command_line_installation.html)).

But I would like to make it work for every installation. So we have to find a way to run a script after the installation is triggered via browser.

tilosp:
@SnowMB for me it works after reloading the page

tilosp:
@SnowMB it only works on new installation. Have you removed the old volumes using `docker-compose down -v`?

SnowMB:
Ok You're right. I didn't try to reload the page.

Issue #95, opened 2017-06-10
nextcloud:fpm Request Entity Too Large


l0calh05t:
I am running a docker-compose setup using nextcloud:fpm, nginx, nginx-proxy and docker-letsencrypt-nginx-proxy-companion (based on the example in .example, but without a non-sqlite db, redis or collabora). This works for the most part, however I cannot upload files larger than 2MB. According to #32 docker-letsencrypt-nginx-proxy-companion can cause this, but adding a `client_max_body_size` setting to the default or VIRTUAL_HOST-specific vhost.d files has no effect.

Could this be caused by the fast_cgi in the fpm module? I tried adding `fastcgi_param PHP_VALUE "upload_max_filesize = 4000M \n  post_max_size=4000M";` to the server config but this doesn't appear to have any effect either.

SnowMB:
Seems like your proxy setting was fine, but the error is in the image: #97

kunago:
Using docker-compose there should be some way to overwrite the conf files but trying nearly everything I am still unable to make it work. Does anyone have a hint?

bang-uin:
Seem like the error is coming from the nginx proxy.

Please try to add 

`client_max_body_size 10G;`

to the end of _proxy/vhost.d/default_

Restart the proxy and try to upload a > 2MB file.

l00ptr:
I have the same issue. It works if I add the client_max_body_size 1G; to proxy/vhost.d/default. But this file is eresased/templated if I restart the letsencrypt-companion container. Don't know where I can put some extra config for it nginx or letsencrypt container.

bang-uin:
Hmm, this is strange. 
I have updated everything, restarted all instances and even rebooted.
My entry is still there.

l00ptr:
did you restart your letsencrypt-companion container? 

Restarting it in my case always overwritte the default file

```
 % sudo cat default    
## Start of configuration add by letsencrypt contain
location /.well-known/acme-challenge/ {
    allow all;
    root /usr/share/nginx/html;
    try_files $uri =404;
    break;
}
client_max_body_size 1G;
## End of configuration add by letsencrypt container


 % sudo docker restart letsencrypt-companion
letsencrypt-companion



 % sudo cat default 
## Start of configuration add by letsencrypt container
location /.well-known/acme-challenge/ {
    allow all;
    root /usr/share/nginx/html;
    try_files $uri =404;
    break;
}
## End of configuration add by letsencrypt container
```

[...]

PR #125, opened 2017-07-06
[WIP] Rework of Examples Section

tilosp:
Closes #103 Closes #98 Closes #95 Closes #75 Closes #86 Closes #104

TODO:
- [x] split the examples into subfolders
- [x] docker-compose example:
  - [x] use JrCs proxy companion #98 
  - [x] Use volumes instead of mounted folders 
  - [x] add proxy settings for upload size #95
  - [x] don't use an external network as it is not needed when the proxy is defined in the same file. Use multiple networks in the file instead
  - [x] add example for apache version
  - [x] change mounts to only `/var/www/html` since it's simpler and less error-prone (would prevent that "config directory not writeable"-error #75 #86) 
  - [x] remove container names (not needed)
  - [x] remove `${DOMAIN}`, it isn't explained anywhere
  - [x] Add example with "indiehosters"-style 2nd container running a cron job #123 
  - [x] Don't use `volumes_from` in the fpm example since it isn't supported in docker-compose file version 3
- [x] derived Dockerfiles:
  - [x] Add examples for both cases apache and fpm
  - [x] Add build tests

PR #162, opened 2017-09-30
add client_max_body_size to upload files > 2 MB

mr-bolle:
With the current configuration i can only upload files < 2MB. If i upload bigger files i get the error message "Request Entity Too Large" back. I add the Solution from [SnowMB @issues/95](https://github.com/nextcloud/docker/issues/95#issuecomment-322683966)


tilosp:
This change is already part of #125.

PR #166, opened 2017-10-02
Add Play with Docker button

tilosp:


tilosp:
I chose postgres because it supports UTF8 4 byte characters without requiring extra configuration (some apps show warning if there is no UTF8 4 byte support, for example the news app). For MySQL this would be required: https://docs.nextcloud.com/server/12/admin_manual/configuration_database/mysql_4byte_support.html

And if we want to have cron the container variant is the only one that works on PWD, because we can't use a Dockerfile to build a customized image. PWD only downloads the `stack.yml` file and  as far as i can tell there is no way to add additional files.

Once we have a way to configure redis via environment variables i would also like to add redis.

pierreozoux:
@tilosp about PG, is it compatible with most apps? What is your experience with it?
(Just asking for a friend :P )

(Actually, I'd be interested to run Nextcloud with https://www.cockroachlabs.com/ :) )

tilosp:
@pierreozoux I migrated my nextcould instance to PostgreSQL about half a year ago (small instance, about 10 Users and no db cluster). 
And the only problem i encountered was that Nextcloud 11/12 is not compatible with the new PostgreSQL 10 (https://github.com/nextcloud/server/issues/5930). 
All the apps i tried worked without any problem. I'am currently running contacts, calendar, tasks, news, deck, u2f, totp, ...

pierreozoux:
Ok, thanks for the answer :)

PR #173, opened 2017-10-23
Allow compose to find all env variables

salinasv:
With two declared "environment" compose only takes the second one. In
thi case it means that the letsencrypt companion and reverse-proxy can't
read the neede variables for the setup, making this example to return a
http 503 error.


tilosp:
Thanks a lot for your contribution!

PR #199, opened 2017-12-17
Added recursive flag to chown command

yotamN:
The current example in the README.md will not change the owner of the subdirectories and because of that I couldn't upload files to the subdirectories in my account, using the `-R` flag worked for me.


tilosp:
Thank you very much for your contribution!

PR #220, opened 2018-01-12
Add cron

### LINKED ISSUE #134: Why use the webcron and not a real cron task
Our owncloud has little access and the cron tasks are not run frequently.
Adding cron for this taks is an option?

tilosp:
Closes #134


tilosp:
@yosifkit or @tianon please review

yosifkit:
Seems sane to me. :+1: :heart:

PR #221, opened 2018-01-13
create a folder for the last Version 13RC1

mr-bolle:
I add a new folder for the current Release Canidate from Nextcloud.
Only if you create a new Container, there should be some issues with the gpg fingerprint (i use the ID from Version 12)

`gpg: Signature made Thu Jan 11 19:44:58 2018 UTC using RSA key ID A724937A
gpg: Good signature from "Nextcloud Security <security@nextcloud.com>"
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 2880 6A87 8AE4 23A2 8372  792E D758 99B9 A724 937A`


SnowMB:
13.0 is now officially released

PR #244, opened 2018-02-06
Manually add nextcloud 13

### LINKED ISSUE #243: Provide image for Nextcloud 13
Now that 13 is final (see https://github.com/nextcloud/server/pull/8187), there should be an official image as well.

tilosp:
Closes #243

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,3 +54,7 @@ jobs:
     - env: VERSION=12.0 VARIANT=fpm ARCH=i386
     - env: VERSION=12.0 VARIANT=apache ARCH=amd64
     - env: VERSION=12.0 VARIANT=apache ARCH=i386
+    - env: VERSION=13.0 VARIANT=fpm ARCH=amd64
+    - env: VERSION=13.0 VARIANT=fpm ARCH=i386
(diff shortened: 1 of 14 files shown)


tilosp:
@tianon @yosifkit
The Jenkins build failed (https://doi-janky.infosiftr.net/job/update.sh/job/nextcloud/1320/) because `generate-stackbrew-library.sh` does not work for Dockerfiles that are not committed. Should we fix this in `generate-stackbrew-library.sh` or in the Jenkins pipeline?

tianon:
Huh, interesting edge case -- we don't have any other repos which will auto-create directories during `update.sh`, which is why we've never run into this before. :smile:

I think the easiest fix would be to update `generate-stackbrew-library.sh` here to include non-committed things, which is usually the opposite of what we do. :laughing:

For reference, here's the pipeline code that scrapes those directory values: https://github.com/docker-library/oi-janky-groovy/blob/c7248109a5d64e6c4a893877d55721427837c54b/update.sh/target-pipeline.groovy#L105-L133

(Which are then passed through to `git add` so that we can use `generate-stackbrew-library.sh` to create a new `library/nextcloud` file and do a build test before pushing those commits.)

Issue #253, opened 2018-02-08
[RFC] Remove cron from container


J0WI:
The current behavior of Cron inside the NextCloud image does not make much sense to me.
In the current example the whole NextCloud container is started in a second instance, that runs Cron. ~But I can also use a very lightweight php-cli container and run the PHP command with an additional `sleep`.~
Since most containers are running on a host system or Kubernetes it's even simpler to create a cronjob or Systemd timer on the host that runs `docker exec -u www-data my-nextcloud php cron.php`.

Would you accept a PR to remove Cron from the image and update docs and examples?

Greek64:
I would be against that.
The whole reason of using Docker or LXC (or at least, one big reason) is to be host independent.
Meaning that anyone can just execute a "docker-compose up -d" and it just works, without further configurations.
And you are now trying to make it again host dependent. That kind of defeats the purpose (IMO)

J0WI:
By default NextCloud uses AJAX Cron anyway. So `docker-compose up -d` will still work for you as before.

A lot of certified images are actually working like this. Some examples:
https://github.com/docker-library/official-images/pull/1203#issuecomment-189392751
https://github.com/docker-library/wordpress/issues/157#issuecomment-356448137
https://github.com/docker-library/postgres/issues/353#issuecomment-335215407

Greek64:
> By default NextCloud uses AJAX Cron anyway

That I do know. 
I was referring to the 

> create a cronjob or Systemd timer on the host 

part, which I strongly disagree, for the aforementioned reasons.

tilosp:
@J0WI 
why would you use a lightweight php-cli container. 
The lightweight container needs at least all the nextcloud dependencies. If you for example use the php:7.1-cli image and install the dependencies you add about 150MB, which i would not call lightweight.

If you add cron to the image it adds about 2MB. Which should not be a big problem for the people which do not use it.

J0WI:
@tilosp you're right, that fails. But how do you think about the `docker exec` solution?

tilosp:
> But how do you think about the docker exec solution?

I see it as an option.
But there are situations were you do not have access to system cron or do not want to use it.
So the extra 2MB won't hurt anyone. And by the way on alpine it only adds 62 Bytes, because busybox is    installed on the alpine base image.

meonkeys:
[...]

PR #259, opened 2018-02-11
Add Alpine variant

J0WI:
This adds a php-fpm-alpine variant with the same dependencies installed as the Debian variants, but without Cron (#253). The `entrypoint.sh` is not fully POSIX compatible, so there are some extra dependencies for it.


J0WI:
Only 64-bit works so far. @tianon do you know how I can build APCu on 32-bit?

Dockerfile-alpine.template
```diff
@@ -0,0 +1,111 @@
+FROM php:7.1-%%VARIANT%%
+
+# entrypoint.sh dependencies
+RUN apk add --no-cache \
+    bash \
+    coreutils \
+    rsync
+
```
tilosp: I think you should merge the top 3 RUN commands into a single one to keep it a similar as possible to the debian one.

tilosp:
we should also update `generate-stackbrew-library.sh` similar to docker-library/wordpress#223 because debian supports different architectures than alpine.

tianon:
```
/usr/local/include/php/ext/pcre/php_pcre.h:29:18: fatal error: pcre.h: No such file or directory
```

Adding `pcre-dev` to `.build-deps` appears to get past that error on `i386`.

tianon:
(https://ram.tianon.xyz/post/2017/12/26/dockerize-compiled-software.html is a writeup about how I usually solve this type of issue generally when Dockerizing software)

J0WI:
_trigger Travis again_

J0WI:
Travis is not building for some reason.

@tianon thanks for the hint. But I'm still confused why this is working on `amd64` without `pcre-dev`.

> we should also update generate-stackbrew-library.sh similar to docker-library/wordpress#223 because debian supports different architectures than alpine.

I think this is not related to this PR. I'd like to only add Alpine images here.

Dockerfile-alpine.template
J0WI: This way is better for build caching. See also other images in the docker-library.
I can adjust the Debian images if required.

J0WI:
Travis is done now. Only fetching from keyserver failed once, which is acceptable.

tilosp:
> I think this is not related to this PR. I'd like to add only Alpine images here.

It is related to this pr. If we merge this now the Jenkins server will try to build the alpine variant on arm32v5 and arm32v7 which will fail because there is no base image.

J0WI:
Okay, I can do this. But the bashbrew dependency makes it a bit tricky. I didn't found a proper way to add bashbrew to Travis, but I don't want to hardcode the architectures again.

tianon:
FWIW, up-to-date pre-built binaries of `bashbrew` can be found at https://doi-janky.infosiftr.net/job/bashbrew/lastSuccessfulBuild/artifact/bin/

J0WI:
@tilosp can you help me out with the bashbrew dependency on Travis?

J0WI:
@tilosp thanks for fixing it. I would like to request a review again.

tilosp:
@J0WI i added the `cron.sh`. This adds only 62 Bytes.

tilosp:
@tianon please review

tilosp:
Thanks a lot for your work!

PR #280, opened 2018-03-12
Add images for the Release Candidates

tilosp:


generate-stackbrew-library.sh
```diff
@@ -72,12 +78,23 @@ for version in "${versions[@]}"; do
 	done) )
 	for variant in "${variants[@]}"; do
 		commit="$(dockerfileCommit "$version/$variant")"
-		fullversion="$(git show "$commit":"$version/$variant/Dockerfile" | awk '$1 == "ENV" && $2 == "NEXTCLOUD_VERSION" { print $3; exit }')"
+		fullversion_with_postfix="$( awk '$1 == "ENV" && $2 == "NEXTCLOUD_VERSION" { print $3; exit }' "$version/$variant/Dockerfile" )"
```
pierreozoux: What is this variable? (I find the usage of postfix confusing)

tilosp: The postifx is the RC version extension
(fullversion_with_postfix contains `13.0.1RC1` and fullversion contains `13.0.1`)
@pierreozoux should i call it `fullversion_with_extension`?

pierreozoux: yes, sounds like a better name.

PR #283, opened 2018-03-15
Manually run update.sh

tilosp:


tilosp:
@tianon another edge case:
Jenkins does not handle auto-deleted folders

A simple solution would be to run `git add -A` and `git commit -m 'Run update.sh'` at the end of the `Commit` stage.

Issue #294, opened 2018-03-26
All requests originate from 172.19.0.X


dachary:
Nextcloud version _13_:
Operating system and version _Debian 9_:
Apache or nginx version _[Docker / nginx proxy / apache](https://github.com/nextcloud/docker/tree/master/.examples/docker-compose/with-nginx-proxy/postgres/apache)_: 

The issue you are facing:

All requests originate from 172.19.0.2 instead of the incoming IP. The main undesirable side effect is that failed login attempts are all attributed to the same IP regardless of their actual origin, causing [login/uploading one image has an added delay of about ~30 seconds](https://github.com/nextcloud/docker/issues/293).

Is this the first time you've seen this error? _N_:

Steps to replicate it:

1. git clone https://github.com/nextcloud/docker
2. cd .examples/docker-compose/with-nginx-proxy/postgres/apache
3. sudo docker-compose up -d

From  **Admin > Logging**:

<pre>
Warning | core | Login failed: 'Max' (Remote IP: '172.19.0.2') | 2018-03-25T23:47:30+0200
</pre>

From the server logs:

<pre>
$ cd docker/.examples/docker-compose/with-nginx-proxy/postgres/apache
$ sudo docker-compose logs proxy
...
proxy_1                  | nginx.1    | a.b.org 86.288.202.127 - - [25/Mar/2018:12:11:25 +0000] "GET /ocs/v2.php/apps/notifications/api/v2/notifications HTTP/2.0" 200 74 "-""Mozilla/5.0 (X11; Linux x86_64; rv:57.0) Gecko/20100101 Firefox/57.0"
$ sudo docker-compose logs app
...
app_1                    | 172.19.0.2 - - [26/Mar/2018:06:56:56 +0000] "GET /ocs/v2.php/apps/notifications/api/v2/notifications HTTP/1.1" 200 689 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:57.0) Gecko/20100101 Firefox/57.0"
</pre>

Chm:
I've got almost the same issue before.

This is due to a security check, NC can't tell if it's true when a same IP says it's a rev-proxy, and keeps login on behalf of someone unlimited.

The solution must match with the rev proxy setup, and I handled it this way:

1. nginx on host, listen on 80/443
2. start container with docker compose, with a separate network 172.32.0.0/24
3. setup compose to build custom image from NC, add a bunch of config php file with one containing
<?php
$CONFIG = array (
[...]

PR #297, opened 2018-03-28
apache: set the remote ip from X-Real-IP by default

dachary:
Relates to: #294

### DIFF
--- a/11.0/apache/Dockerfile
+++ b/11.0/apache/Dockerfile
@@ -95,7 +95,14 @@ RUN { \
 
 VOLUME /var/www/html
 
-RUN a2enmod rewrite
+RUN a2enmod rewrite remoteip ;\
+    {\
+     echo RemoteIPHeader X-Real-IP ;\
+     echo RemoteIPTrustedProxy 10.0.0.0/8 ;\
+     echo RemoteIPTrustedProxy 172.16.0.0/12 ;\
+     echo RemoteIPTrustedProxy 192.168.0.0/16 ;\
+    } > /etc/apache2/conf-available/remoteip.conf;\
+    a2enconf remoteip
 
 ENV NEXTCLOUD_VERSION 11.0.8
 
--- a/13.0/apache/Dockerfile
+++ b/13.0/apache/Dockerfile
@@ -95,7 +95,14 @@ RUN { \
 
 VOLUME /var/www/html
 
-RUN a2enmod rewrite
+RUN a2enmod rewrite remoteip ;\
+    {\
+     echo RemoteIPHeader X-Real-IP ;\
+     echo RemoteIPTrustedProxy 10.0.0.0/8 ;\
+     echo RemoteIPTrustedProxy 172.16.0.0/12 ;\
+     echo RemoteIPTrustedProxy 192.168.0.0/16 ;\
+    } > /etc/apache2/conf-available/remoteip.conf;\
+    a2enconf remoteip
 
 ENV NEXTCLOUD_VERSION 13.0.1
 
--- a/12.0/apache/Dockerfile
+++ b/12.0/apache/Dockerfile
@@ -95,7 +95,14 @@ RUN { \
 
 VOLUME /var/www/html
 
-RUN a2enmod rewrite
+RUN a2enmod rewrite remoteip ;\
+    {\
+     echo RemoteIPHeader X-Real-IP ;\
+     echo RemoteIPTrustedProxy 10.0.0.0/8 ;\
+     echo RemoteIPTrustedProxy 172.16.0.0/12 ;\
+     echo RemoteIPTrustedProxy 192.168.0.0/16 ;\
+    } > /etc/apache2/conf-available/remoteip.conf;\
+    a2enconf remoteip
 
 ENV NEXTCLOUD_VERSION 12.0.6
 
--- a/update.sh
+++ b/update.sh
@@ -18,7 +18,7 @@ declare -A base=(
 )
 
 declare -A extras=(
-	[apache]='\nRUN a2enmod rewrite'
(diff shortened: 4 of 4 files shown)


dachary:
Could someone please confirm the travis error:

<pre>
The command "bash -c "[[ $hash_before = $(git add -A && git write-tree) ]]"" exited with 1.
Skipping the after_script stage due to the configuration.
</pre>

is related to my change? I'm happy to dig further to figure it out. Just checking that I'm not chasing a gost ;-)

dachary:
For the record, I have manually tested the changes with

* cd docker/13.0/apache
* sudo docker build -t nextcloud:apache .
* cd docker/.examples/docker-compose/with-nginx-proxy/postgres/apache
* sudo docker-compose up app
* login with a wrong password
* login with the right password
* go to /settings/admin/logging and verify the failed login attempt is recorded with the remote-ip instead of the local ip

pierreozoux:
This file is generated by script.
Could you add your change to https://github.com/nextcloud/docker/blob/master/update.sh#L20
(then in another commit, run the script itself to update the images)

Nice to see you contribute here, I've been following your work on ceph.

dachary:
> This file is generated by script.

Oh, thanks for the hint, will do!

> Nice to see you contribute here, I've been following your work on ceph.

Thank you for the kind words, makes me feel very welcome :-)

pierreozoux:
the 2 commits, can be in the same PR, it just ease the reviewing to separate them.

dachary:
@pierreozoux Ok, will do, thanks for the advice :-D

dachary:
All pass except https://travis-ci.org/nextcloud/docker/jobs/360574391 because travis has trouble reaching the keyserver. Should I force-repush to trigger another build? It looks like that kind of transient error is fairly common so maybe you prefer to force-merge instead of wasting additional CI runs? I'm happy either way, just let me know ;-)

<pre>
 gpg --keyserver ha.pool.sks-keyservers.net --recv-keys 28806A878AE423A28372792ED75899B9A724937A
gpg: keybox '/tmp/tmp.AEdOLd/pubring.kbx' created
gpg: keyserver receive failed: Address not available
</pre>

pierreozoux:
One last question, in nginx, you have to tell from whom to trust who can set this header, here in apache, there is no such mechanism? It will just trust everybody including me?

dachary:
> in apache, there is no such mechanism?

There is but I misread the [documentation](https://httpd.apache.org/docs/trunk/mod/mod_remoteip.html#remoteiptrustedproxy) and incorrectly concluded it would only allow internal IPs by default. Fixing that, thanks for catching this mistake!

dachary:
Added the [suggested internal IP ranges](https://github.com/nextcloud/docker/issues/294#issuecomment-376454764) and [manually tested](https://github.com/nextcloud/docker/pull/297#issuecomment-377045060) they work as expected on a live instance.

dachary:
* https://travis-ci.org/nextcloud/docker/jobs/360632955
* https://travis-ci.org/nextcloud/docker/jobs/360632956
* https://travis-ci.org/nextcloud/docker/jobs/360632957
* https://travis-ci.org/nextcloud/docker/jobs/360632958

<pre>
W: Failed to fetch http://deb.debian.org/debian/dists/jessie-updates/main/binary-amd64/Packages  Hash Sum mismatch
</pre>

transient error apparently. I won't repush since it seems more effective for someone with access to re-start the failed jobs instead.

13.0/apache/Dockerfile
```diff
@@ -96,6 +96,8 @@ RUN { \
 VOLUME /var/www/html
 
 RUN a2enmod rewrite
+RUN a2enmod remoteip
+RUN { echo RemoteIPHeader X-Real-IP ; echo RemoteIPTrustedProxy 10.0.0.0/8 ; echo RemoteIPTrustedProxy 172.16.0.0/12 ; echo RemoteIPTrustedProxy 192.168.0.0/16 ; } > /etc/apache2/conf-enabled/remoteip.conf; chown www-data:root /etc/apache2/conf-enabled/remoteip.conf
```
tilosp: please reduce the number of layer by merging these three run statements into one

tilosp:
@dachary i removed the `Fixes: #294` because this only fixes it for apache and not fpm

13.0/apache/Dockerfile
dachary: Good idea, will do!

dachary:
@tilosp modified and repushed, thanks for your review :-)

11.0/apache/Dockerfile
```diff
@@ -95,7 +95,7 @@ RUN { \
 
 VOLUME /var/www/html
 
-RUN a2enmod rewrite
+RUN a2enmod rewrite remoteip ; { echo RemoteIPHeader X-Real-IP ; echo RemoteIPTrustedProxy 10.0.0.0/8 ; echo RemoteIPTrustedProxy 172.16.0.0/12 ; echo RemoteIPTrustedProxy 192.168.0.0/16 ; } > /etc/apache2/conf-enabled/remoteip.conf; chown www-data:root /etc/apache2/conf-enabled/remoteip.conf
```
tilosp: having the entire run statement in one line makes it difficult to read.
It would look a lot nicer if you formated it similar to the run statement above that sets the recommended PHP.ini settings

dachary:
@tilosp thanks for the detailed instructions, it makes it super easy for me :-) Amended and repushed !

pierreozoux:
Thanks both of you for the nice work!

dachary:
@pierreozoux silly question: when and how is this going to show in the docker hub?

pierreozoux:
I think the docker bot runs once a day.

PR #299, opened 2018-03-31
update.sh: do not chmod -w Dockerfile

dachary:
In 18db7e679fd92e2d36d6113b1e174e6f1afda2d6 the Dockerfile write
permissions were removed to help remind the new contributor that they
should not be modified. However git does not keep that information (it
only keeps the executable bit in the file permissions), therefore
someone with a fresh clone will not have a file that is read-only.

Removing the write permission will however be a problem for anyone
running the update.sh script twice: it will fail to remove the files
because it uses rm -r instead of rm -fr.

Instead of fixing the removale to use the force flag, we revert the
modification that remove the write permission since it does not
actually help anyone.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -56,7 +56,6 @@ function create_variant() {
 	mkdir -p "$dir"
 
         template="Dockerfile-${base[$variant]}.template"
-        rm -f "$dir/Dockerfile"
         echo "# DO NOT EDIT: created by update.sh from $template" > "$dir/Dockerfile"
 	cat "$template" >> "$dir/Dockerfile"
 
@@ -74,8 +73,6 @@ function create_variant() {
 		s/%%MEMCACHED_VERSION%%/'"${pecl_versions[memcached]}"'/g;
 		s/%%REDIS_VERSION%%/'"${pecl_versions[redis]}"'/g;
 	' "$dir/Dockerfile"
-        # remove write permission as a reminder this file must be modified with this script
-        chmod -w "$dir/Dockerfile"
 
 	# Copy the shell scripts
 	for name in entrypoint cron; do


dachary:
For the record here is the error we get when running `./update.sh` twice:

<pre>
loic@raise:~/software/nextcloud/docker$ bash -x ./update.sh 
+ set -eo pipefail
+ php_version=([default]='7.1')
+ declare -A php_version
+ cmd=([apache]='apache2-foreground' [fpm]='php-fpm' [fpm-alpine]='php-fpm')
+ declare -A cmd
+ base=([apache]='debian' [fpm]='debian' [fpm-alpine]='alpine')
+ declare -A base
+ extras=([apache]='\nRUN a2enmod rewrite\nRUN a2enmod remoteip\nRUN { echo RemoteIPHeader X-Real-IP ; echo RemoteIPTrustedProxy 10.0.0.0/8 ; echo RemoteIPTrustedProxy 172.16.0.0/12 ; echo RemoteIPTrustedProxy 192.168.0.0/16 ; } > /etc/apache2/conf-enabled/remoteip.conf; chown www-data:root /etc/apache2/conf-enabled/remoteip.conf\n' [fpm]='' [fpm-alpine]='')
+ declare -A extras
+ pecl_versions=([APCu]='5.1.11' [memcached]='3.0.4' [redis]='3.1.6')
+ declare -A pecl_versions
+ variants=(apache fpm fpm-alpine)
+ min_version=11.0
+ travisEnv=
+ find . -maxdepth 1 -type d -regextype sed -regex '\./[[:digit:]]\+\.[[:digit:]]\+\(-rc\)\?' -exec rm -r '{}' ';'
rm: remove write-protected regular file './11.0/apache/Dockerfile'? ^C
</pre>

PR #300, opened 2018-04-01
update.sh: prevent apache out of memory

dachary:
With a default number of workers set to 150, the example apache
configuration will use over 4GB of RAM in the following, rather
likely, demonstration scenario:

* sync a directory with 1000 photos in one directory from a desktop to
  nextCloud
* go to the web interface and quickly scroll down to the bottom of the
  list of photos that were just uploaded
* apache will run as many apache workers as possible (150) to create
  thumbnails and use over 4GB of RAM

The default number of workers is reduced to 30 which has been
experienced to fit in a 2GB virtual machine. It takes longer to
complete but does not trigger out of memory problems.

### DIFF
--- a/11.0/apache/Dockerfile
+++ b/11.0/apache/Dockerfile
@@ -102,7 +102,8 @@ RUN a2enmod rewrite remoteip ;\
      echo RemoteIPTrustedProxy 172.16.0.0/12 ;\
      echo RemoteIPTrustedProxy 192.168.0.0/16 ;\
     } > /etc/apache2/conf-available/remoteip.conf;\
-    a2enconf remoteip
+    a2enconf remoteip;\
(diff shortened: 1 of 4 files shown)


pierreozoux:
With this change, I'm wondering if this shouldn't go to upstream.
What do you think @tianon ?

dachary:
> I'm wondering if this shouldn't go to upstream.

You mean the `php:7.1-apache` image?

dachary:
@pierreozoux it is likely that some users of the `php:7.1-apache` rely on the existing default value `MaxRequestWorkers 150` and changing it to `MaxRequestWorkers 30` may cause a regression. Given the recent bugfix for `X-Real-IP`, we have reason to suspect the image was not used for anything but testing. In a testing (or small scale) context, changing to `MaxRequestWorkers 30` will not cause a problem to nextCloud users.

That is the proposed rationale for proposing this change to this repository rather than `php:7.1-apache`. Does it make sense?

J0WI:
deleted user/repo

PR #307, opened 2018-04-03
Create a data dir outside the webroot.

### LINKED ISSUE #306: Add a /var/www/data folder on the image as a preferred way to store data outside the webroot
I'm trying to deploy a nextcloud docker instance following best practices. In this case storing the Nextcloud data folder outside the files available to the public.

I contributed some code in this direction:
https://github.com/nextcloud/docker/commit/5714672aa6f2b5cbac6d5d268763bf06c607a5eb

but now I realized that if I set a docker named volume for that folder with `-v myvolume:/var/www/data` as a command line example that folder will not writable by the www-data user.

My first thought was to modify the entrypoint so we chown the entire /var/www at runtime like is done on the image Dockerfile but it may lead to some undesired chowning.

I feel a more sane solution is to create this /var/www/data folder in the Dockerfile with the correct permissions so any mounts can be writable by www-data.

In my opinion I will also change the default NEXTCLOUD_DATA_DIR to `/var/www/data` instead of `/var/www/html/data` but maybe that is a bigger change.

rodrigoaguilera:
closes #306

### DIFF
--- a/12.0/apache/Dockerfile
+++ b/12.0/apache/Dockerfile
@@ -90,6 +90,7 @@ RUN { \
         echo 'opcache.revalidate_freq=1'; \
     } > /usr/local/etc/php/conf.d/opcache-recommended.ini; \
     \
+    mkdir /var/www/data; \
     chown -R www-data:root /var/www; \
     chmod -R g=u /var/www
 
--- a/12.0/fpm-alpine/Dockerfile
+++ b/12.0/fpm-alpine/Dockerfile
@@ -79,6 +79,7 @@ RUN { \
         echo 'opcache.revalidate_freq=1'; \
     } > /usr/local/etc/php/conf.d/opcache-recommended.ini; \
     \
+    mkdir /var/www/data; \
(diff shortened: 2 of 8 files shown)


rodrigoaguilera:
I do not understand the travis failure. Can someone have a look at it?

tilosp:
You have to run `update.sh` before committing. Travis checks if you have run `update.sh` by running it and checking for changes.

rodrigoaguilera:
Ok that makes sense.

I searched for some info about this for contributors and found nothing. Maybe a contributing.md file at the root describing this simple step can be helpful.

tilosp:
this shouldn't break anything but @pierreozoux please take a look too

tilosp:
> I searched for some info about this for contributors and found nothing. Maybe a contributing.md file at the root describing this simple step can be helpful.

Yes adding a `CONTRIBUTING.md` in addition to the `DO NOT EDIT` notes (#298) makes sense.

PR #308, opened 2018-04-04
Fix app-directory order so that shipped apps can be updated

julijane:
The configuration as used, albeit suggested in the Nextcloud admin manual, does not work correctly. There is a current example. Nextcloud 13.0.1 ships with files_pdfviewer 1.2.0, but an update to 1.2.1 is available. However when the app is updated, while 1.2.1 is installed into the custom_apps folder, it is not actually used because Nextcloud will still load 1.2.0 from the first app folder. By changing the order of the two app folders, everything works as it should be. Additional apps and updates are put into the custom_apps folder. 

This is arguably a bug in the server but I am not sure about that.

### DIFF
--- a/.config/apps.config.php
+++ b/.config/apps.config.php
@@ -2,14 +2,14 @@
 $CONFIG = array (
   "apps_paths" => array (
       0 => array (
-              "path"     => OC::$SERVERROOT."/apps",
-              "url"      => "/apps",
-              "writable" => false,
-      ),
-      1 => array (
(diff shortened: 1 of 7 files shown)


julijane:
See also https://github.com/nextcloud/documentation/pull/739

julijane:
Closing this pull request for now as there is some discussion about the issue going on at https://github.com/nextcloud/server/issues/8929 and it is unclear what will come out of that.

Issue #326, opened 2018-05-03
separate application code from data


rdxmb:
Hello,

with https://github.com/nextcloud/docker/issues/251 there was a problem with missing files in `/var/www/html`

I don't know if I understand it the right way - but it would be the best to separate appliation code from personal data. This could be done , e.g. , via symlinks.

What about https://github.com/nextcloud/docker/commit/5ddd03aa501a53d609ec4f8de3add575fc10f5ce ? Is this a step for that? The folder seems to be empty after an update from 12.7 to 13.0.2

skjnldsv:
Please read the readme about persistent data: https://github.com/nextcloud/docker#persistent-data

rdxmb:
I did. As I use this in Docker SwarmMode, `/var/www/html` is not a docker volume, but a bind mount like

`/mnt/shared/nextcloud:/var/www/html`

Is this not supported with this image? If files in the image within /var/www/html were added with a new version, with a bind mount they will not appear - for they will be "covered" by the mounted files. With docker volumes they get "merged", as long as `nocopy: true` is not set in the compose- or stack-file.

skjnldsv:
you want persistent data or persistent nextcloud instance?
Because I use the bind mount with /data:/var/www/html/data and it works fine :)

rdxmb:
@skjnldsv What do you mean with "persistent nextcloud instance"?

I just want to be able to upgrade nextcloud with changing the image - and have the data, configs, apps persistent.

When changing the image from 12.x to 12.7 there were application files missing in /var/www/html , as reported in https://github.com/nextcloud/docker/issues/251 . I have had a look at the Dockerfiles and have seen that there were files `rsync`-ed with `--delete` to `/var/www/html` when running supervisord . I think there is the problem.

rdxmb:
Maybe I did not get the real meaning of the docs. Is it enough to mount

> /var/www/html/custom_apps installed / modified apps
> /var/www/html/config local configuration
> /var/www/html/data the actual data of your Nextcloud
> /var/www/html/themes/<YOU_CUSTOM_THEME> theming/branding

?

What does

> /var/www/html Main folder, needed for updating

really mean?

 Main folder, needed for updating 

mean?


/var/www/html/custom_apps installed / modified apps
/var/www/html/config local configuration
/var/www/html/data the actual data of your Nextcloud
/var/www/html/themes/<YOU_CUSTOM_THEME> theming/branding

rdxmb:
[...]

PR #331, opened 2018-05-07
update README.md for production environments

rdxmb:
Fixes https://github.com/nextcloud/docker/issues/326

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,47 +41,35 @@ As the fastCGI-Process is not capable of serving static files (style sheets, ima
 By default this container uses SQLite for data storage, but the Nextcloud setup wizard (appears on first run) allows connecting to an existing MySQL/MariaDB or PostgreSQL database. You can also link a database container, e. g. `--link my-mysql:mysql`, and then use `mysql` as the database host on setup. More info is in the docker-compose section.
 
 ## Persistent data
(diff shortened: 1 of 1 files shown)


tilosp:
@rdxmb the updater does not touch the `config`, `data`, `custom_apps` and `themes` directories.
We use rsync for updating the files in the `/var/www/html/` dir and it is configured to ignore those directories. See: https://github.com/nextcloud/docker/blob/adff4fe4397d080909a9853272b05535b09ad172/docker-entrypoint.sh#L44
Rsync does not care if `/var/www/html/` in one volume or split across many. Because by default rsync will ignore file system boundaries and simply sync across them. 
This means that the additional volumes will not improve data persistency across updates. This should already be taken care of by rsync. And if there is some persistency issue with the update script please say so, so that we can get it fixed.

In terms of backups, it should be posible to make backups of a subdir of a volume.
And other that i can not see many benefits to having multiple volumes stored on the same file system.

@pierreozoux what do you think? If you think it makes sense I am fine with recommending splitting of the volume. But we should definitely remove the sentence that says it is reqired for the updater to work.

And if we decided to split the volume up in the example docker-compose files. It will break the setups of users how decided to pull the new version, because they will then have a empty data and config dir.

rdxmb:
@tilosp Thanks for your explanation. The problem did not appear within the excluded directories.
Did you read the linked issues? I guess the problem was with bin-mounting `/var/www/html` and a new file in /var/www/html/lib - which was not found because of the bind-mount.

If the problem is caused by something different, the solution must be different, of course.

Please think about the problem for the missing file and how to get sure that there will be no problem with the next image upgrade.

rdxmb:
update:

Just tried to reproduce this with an upgrade-process from 12.0.1 to 12.0.6 up to 12.0.7 . I must say I cannot reproduce this anymore.

very  :confused:

Issue #327, opened 2018-05-03
FeatureRequest: write docker logs when migrating data


rdxmb:
When upgrading from 12.07 to 13.0.2 the web-container is not reachable for minutes. 
`docker logs` seems empty.

When checking `ps xa` within the container, there are running some `rsync`-processes.

Could you please write some output to the docker logs (stdout)? Echoing `'migrating data. This could take some time...'` would be enough for the first step.

pierreozoux:
Would you like to make a PR? The script is [here](https://github.com/nextcloud/docker/blob/master/docker-entrypoint.sh)

PR #333, opened 2018-05-08
echo to stdout when migration-script is running

rdxmb:
fixes https://github.com/nextcloud/docker/issues/327

### DIFF
--- a/12.0/apache/entrypoint.sh
+++ b/12.0/apache/entrypoint.sh
@@ -34,8 +34,9 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ]; then
     fi
 
     if version_greater "$image_version" "$installed_version"; then
-        echo -n "running uprade from $installed_version to $image_version ..."
+        echo -n "Initializing version $image_version ... "
         if [ "$installed_version" != "0.0.0.0" ]; then
+            echo -ne "\nRunning upgrade from $installed_version ... "
             run_as 'php /var/www/html/occ app:list' | sed -n "/Enabled:/,/Disabled:/p" > /tmp/list_before
         fi
         if [ "$(id -u)" = 0 ]; then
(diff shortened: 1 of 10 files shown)


rdxmb:
What's the problem with the Travis Check above?

tilosp:
@rdxmb you need to run `update.sh` und commit the changes.

rdxmb:
what now?

tilosp:
We should fix the ci problems before merging. This currently causes the [`override-cmd`](https://github.com/docker-library/official-images/blob/master/test/tests/override-cmd/run.sh) test to fail. It fails because of the additional version output. And as far as i know this test is required for this to get merged into [docker-library/official-images](https://github.com/docker-library/official-images).

One possible solution would be to not print anything in the startup script if the current version is `0.0.0`

12.0/apache/entrypoint.sh
```diff
@@ -33,7 +33,9 @@ if version_greater "$installed_version" "$image_version"; then
 fi
 
 if version_greater "$image_version" "$installed_version"; then
+    echo -n "Initializing version $image_version ... "
```
tilosp: the `-n` option, which you use to keep this and `finished.` in one line, does not work if you upgrade from a previous nextcloud version. Because the `occ upgrade` command prints information and is run in betwen this echo and the `finished.` echo.

rdxmb:
Just simplified the output and removed the `-n` . We could also run rsync with `-v` - to have an output what is running with `docker logs` . What do you think?

rdxmb:
Can somebody help to get this ready? What's the problem with the ci-checks here?

tilosp:
@rdxmb once #459 is merged i will rebase this onto master, which should fix the ci.

tilosp:
rebased to master

tilosp:
Thanks for your contribution!

rdxmb:
thanks for merging.

Issue #337, opened 2018-05-16
PR: Add in `.examples` an example for LAN solutions with HTTPS without letsencrypt


lil5:
I've written one with [paulczar/omgwtfssl](https://hub.docker.com/r/paulczar/omgwtfssl/).

https://gist.github.com/lil5/b7263a59f385f1296fcfdb5fbf816a4b

pierreozoux:
Do you want to PR?

lil5:
@pierreozoux yeah sure

PR #338, opened 2018-05-16
Add example for LAN solutions with HTTPS without letsencrypt #337

lil5:
- [x] :warning: ~This is un-tested: will test and debug now.~


lil5:
Tested and it works.

PR #355, opened 2018-06-01
Do "pecl install" in series so it fails properly

J0WI:
See matomo-org/docker@88a8be3


tilosp:
Thanks for your contribution!

PR #358, opened 2018-06-04
Adjust gpg code to kill daemons, cutting down on race conditions

J0WI:
Backporting this (see https://github.com/matomo-org/docker/issues/104#issuecomment-394511396)


tilosp:
Thanks a lot for your work!

PR #382, opened 2018-07-05
Remove the `/var/www/html` volume

vfreex:
Auto-upgrade procedure in `docker-entrypoint.sh` is also removed
since it won't be necessary if `/var/www/html` is not a volume.


tilosp:
This would break most existing setups using this image.

PR #392, opened 2018-07-15
add example for mount order for nginx work

### LINKED ISSUE #391: all apps does not work on example docker-compose (fpm+nginx)
steps to reproduce

* get https://github.com/nextcloud/docker/blob/master/.examples/docker-compose/insecure/postgres/fpm/docker-compose.yml
* docker-compose build
* docke-compose up
* login and register
* install random app (for example mail, calendar, mind,map, etc

and you get

> Refused to execute script from localhost:8080
> because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
> 

in Chrome Dev Console

and icon for apps will be disabled

default

allustin:
add to example fpm mount example for custom-dirs like it describe in documentation

https://github.com/nextcloud/docker#persistent-data

fix #391

### DIFF
--- a/.examples/docker-compose/insecure/postgres/fpm/docker-compose.yml
+++ b/.examples/docker-compose/insecure/postgres/fpm/docker-compose.yml
@@ -13,6 +13,11 @@ services:
     image: nextcloud:fpm
     restart: always
     volumes:
+      - nextcloud-data:/var/www/html/data  
+      - nextcloud-config:/var/www/html/config  
+      - nextcloud-apps:/var/www/html/custom_apps
+      - nextcloud-theme:/var/www/themes/mytheme
+      # full html dir must be last in mount order
       - nextcloud:/var/www/html
     environment:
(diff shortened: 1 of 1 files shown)


.examples/docker-compose/insecure/postgres/fpm/docker-compose.yml
```diff
@@ -27,10 +32,20 @@ services:
     ports:
       - 8080:80
     volumes:
+      # volumes must copy-past volume section from app with readonly 
+      - nextcloud-data:/var/www/html/data:ro
+      - nextcloud-config:/var/www/html/config:ro
```
J0WI: nginx does not need acces to `nextcloud-data` or `nextcloud-config`. `nextcloud-theme` is deprecated.
Could you move this to a section in the README? I think by default only one volume should be used, but you could leaf them commented as an example.

J0WI:
orphaned PR

PR #399, opened 2018-07-26
Nextcloud 14.0 Beta

tilosp:


tilosp:
Sticking to php 7.1 for now, because php 7.2 fails to build.

PR #416, opened 2018-08-07
 Switch to php 7.2 for nextcloud 13.0 and greater 

### LINKED ISSUE #196: Update base image to php 7.2 (Nextcloud 13)
The current base image `php:7.1-fpm` manifests this bug: https://github.com/nextcloud/server/issues/7380

Please update the base image to one with `php:7.1-fpm`. This can be somewhat problematic as I see none of the `7.2` images are based on Debian anymore.

Any plans to move this image to be based on alpine or other common lightweight bases?

tilosp:
Closes #196

### DIFF
--- a/12.0/apache/Dockerfile
+++ b/12.0/apache/Dockerfile
@@ -43,7 +43,6 @@ RUN set -ex; \
         gd \
         intl \
         ldap \
-        mcrypt \
         opcache \
         pcntl \
         pdo_mysql \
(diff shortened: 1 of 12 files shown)


tilosp:
I had to remove mcrypt because it has been removed from php 7.2. (See: https://secure.php.net/releases/7_2_0.php)

But this shouldn't be a problem because its only an optional dependency. (See: https://docs.nextcloud.com/server/13/admin_manual/installation/source_installation.html#prerequisites-for-manual-installation)
And it was also removed from the nextcloud vm when they migrated to php 7.2. (See: https://github.com/nextcloud/vm/commit/d010c68844a8cc4f9894026e65cfd5d38357fc0e#diff-9a549a9e5acbbb9ce4941ef32f8949c3L226)

PR #429, opened 2018-08-20
Readme: add docker-compose down in update process

rubo77:
Signed off by Ruben Barkow <github@r.z11.de>


SnowMB:
Why should we need that? `docker-compose up` already stops and removes any old container.

rubo77:
sorry, missed that. I think in older docker-compose versions this was needed.

rubo77:
sorry, missed that. I think in older docker-compose versions this was needed.

PR #438, opened 2018-08-27
Allow the UID and GID to be changed

### LINKED ISSUE #359: How to change UID/GID in NextCloud container?
I have a CIFS/SMB volume mounted on my docker host, accessed by many of my containers. For all the other containers, I can specify a UID/GID in the docker-compose file so that they have read/write access to this volume.
How do you specify the UID/GID for NextCloud? As it stands, it can read the volume but doesn't have write access.
Thanks

tilosp:
Closes #359

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -19,6 +19,13 @@ run_as() {
     fi
 }
 
+if [ -n "${NEXTCLOUD_UID+x}" ]; then
+    usermod -u "$NEXTCLOUD_UID" www-data
+fi
+if [ -n "${NEXTCLOUD_GID+x}" ]; then
+    groupmod -g "$NEXTCLOUD_GID" www-data
+fi
+
 installed_version="0.0.0.0"
 if [ -f /var/www/html/version.php ]; then
     # shellcheck disable=SC2016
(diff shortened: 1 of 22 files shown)


docker-entrypoint.sh
```diff
@@ -19,6 +19,13 @@ run_as() {
     fi
 }
 
+if [ -n "${NEXTCLOUD_UID+x}" ]; then
+    usermod -u "$NEXTCLOUD_UID" www-data
+fi
+if [ -n "${NEXTCLOUD_GID+x}" ]; then
+    groupmod -g "$NEXTCLOUD_GID" www-data
```
J0WI: group is always set to root in https://github.com/nextcloud/docker/blob/35fd42d0429ed0deaf2081f8210f31ce3758ab40/docker-entrypoint.sh#L40

tianon:
Can this not be implemented transparently via Docker's `--user` option as in https://github.com/docker-library/docs/blob/master/php/README.md#running-as-an-arbitrary-user?

https://github.com/docker-library/wordpress/pull/249/files is an example of what was necessary in PHP (I can get other examples later if you want, on my phone ATM 😅)

PR #449, opened 2018-09-07
Auto update indices on NextCloud update

LouisMT:
After upgrading my NextCloud Docker from `13.0.6` to `14.0.0` the control panel was showing a warning that my database was missing indices.

Currently the `README.md` states that updating the NextCloud Docker is as easy as pulling the new image and recreating the container, and the upgrade process will start automatically. So I think there are 2 options:

* Automatically add missing indices during the upgrade process (as I've done in this PR)
* Add a note to the `README.md` that you may have to manually add missing indices after upgrading

----

If you're looking for a way to manually add missing indices, just run:

```shell
docker exec -it --user www-data $CONTAINER_NAME bash
./occ db:add-missing-indices
```

### DIFF
--- a/12.0/apache/entrypoint.sh
+++ b/12.0/apache/entrypoint.sh
@@ -51,6 +51,7 @@ if version_greater "$image_version" "$installed_version"; then
 
     if [ "$installed_version" != "0.0.0.0" ]; then
         run_as 'php /var/www/html/occ upgrade'
+        run_as 'php /var/www/html/occ db:add-missing-indices'
 
         run_as 'php /var/www/html/occ app:list' | sed -n "/Enabled:/,/Disabled:/p" > /tmp/list_after
         echo "The following apps have beed disabled:"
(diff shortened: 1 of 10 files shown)


tilosp:
Adding the missing indices is intentionally not done when running `occ upgrade` because this is a very slow operation if you have a large database.
This step is also optional. It will improve the performance but is not required for nextcloud to run.
Because of this it also doesn't make sense to tell users that is is required after each update.

I think the warning in the admin page is enough. And it should also link the user to the docs, where the user can read how to add the indices.

robclay:
I tried that, (and I assuming $CONTAINER_NAME is my 'app' one, and not the Maria one, yes?) and get:

```
 docker exec -it --user www-data $ncinit_app_1 bash ./occ db:add-missing-indices
./occ: line 2: ?php: No such file or directory
./occ: line 3: /bin: Is a directory
./occ: line 4: syntax error near unexpected token `('
./occ: line 4: ` * Copyright (c) 2013 Thomas MÃ¼ller <thomas.mueller@tmit.eu>'
```
Did I choose the wrong container?

robclay:
HA!  I am an idiot.  Those were separate commands.
First
docker exec -it --user www-data $ncinit_app_1 bash
THEN
./occ db:add-missing-indices

robclay:
Thanks!

PR #450, opened 2018-09-07
Allow Referrer-Policy header to be configured

LouisMT:
This PR adds the ability to configure the `Referrer-Policy` header using the Docker image by setting an environment variable.

Example:

```shell
docker run -e APACHE_REFERRER_POLICY=no-referrer nextcloud
```

This is needed because I'm now getting the following warning in my NextCloud 14 control panel and there's no clean way of fixing this right now (see nextcloud/server#9122):

> The "Referrer-Policy" HTTP header is not set to "no-referrer", "no-referrer-when-downgrade", "strict-origin" or "strict-origin-when-cross-origin". This can leak referer information. See the W3C Recommendation ↗.

Note that this header will **not** be added to the response if this environment variable is not set, so this is not a breaking change. Without explicitly setting the variable, the security message will still appear.

### DIFF
--- a/12.0/apache/Dockerfile
+++ b/12.0/apache/Dockerfile
@@ -95,14 +95,18 @@ RUN { \
 
 VOLUME /var/www/html
 
-RUN a2enmod rewrite remoteip ;\
+RUN a2enmod rewrite remoteip env headers ;\
     {\
      echo RemoteIPHeader X-Real-IP ;\
      echo RemoteIPTrustedProxy 10.0.0.0/8 ;\
      echo RemoteIPTrustedProxy 172.16.0.0/12 ;\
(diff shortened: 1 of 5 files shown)


tilosp:
I think it would make sense to have this header by default.
@pierreozoux @J0WI what do you think?

J0WI:
Yes, they should be there by default. But I don't want to make https://github.com/nextcloud/server/issues/8207#issuecomment-363839953 worse.

LouisMT:
Do you mean that this header should be added in the `server` repository using PHP, so it's all in one place and easier to refactor later?

J0WI:
The header is now present in NC 15.

Issue #448, opened 2018-09-07
setup warning of security scan in Nextcloud 14


suntorytimed:
I am using a slightly modified docker-compose setup of .examples/docker-compose/with-nginx-proxy/mariadb-cron-redis/fpm/ for my server.

Since Nextcloud 14 the security scan complains about:
```
There are some warnings regarding your setup.
The "Referrer-Policy" HTTP header is not set to "no-referrer", "no-referrer-when-downgrade", "strict-origin" or "strict-origin-when-cross-origin". This can leak referer information. See the W3C Recommendation ↗.
```

Is this linked to the nginx setup inside of the container or is it a general issue with the Nextcloud fpm container?

LouisMT:
You need to set this header in your web server. For example:

```
server {
    # Configuration
    add_header Referrer-Policy no-referrer always;
}
```

suntorytimed:
@LouisMT thanks. I think this should be added to the nginx.conf files in the .examples

opgailey:
@LouisMT What file exactly is it that needs to be edited to include that header?

LouisMT:
It's your web server's configuration file. Its location and syntax depend on the web server you're using.

suntorytimed:
@LouisMT Thanks for submitting the patch. My server is all green now :slightly_smiling_face:

Nottt:
From Google here.

Can confirm this works, but it's weird because according to their [forum](https://help.nextcloud.com/t/the-referrer-policy-http-header-is-not-set-to-no-referrer/36613/5) they say to add it to `.htaccess` instead and it did not fix it for me.

Now I have both 2 "fixes" applied, no apparent issues yet. Is this because we are using docker or what? 

Thanks!

PR #451, opened 2018-09-07
Add Referrer-Policy header examples for nginx

LouisMT:
Add an `Referrer-Policy` header to the example nginx configuration files as requested in #448.

I'm not sure if `no-referrer` should be the default here, I think it's a safe option but it may break things? I can also just add a commented version of this line.


tilosp:
@LouisMT the nginx config is base on documentation.
Could you open a pr in the docs repo to change it there? (See: https://github.com/nextcloud/documentation/blob/b627e8e855d5b9d51ec882fc0ac7ae486bdd0c1a/admin_manual/installation/nginx.rst)
Once it is merged into the docs i will merge this pr.

LouisMT:
See #852 (has been merged).

LouisMT:
I'm sorry I missed the second header - should be all good now! :smile:

PR #460, opened 2018-09-12
Add custom memory limit for PHP CLI (fix #447)

### LINKED ISSUE #447: PHP memory limit to 128 ?
Hello, i'm using the latest image (with nextcloud 14)
 and when i run any php occ command. i get this message
`The current PHP memory limit is below the recommended value of 512MB.`
i haven't touch the image. it's the official image from docker hub vanilla (with clean installation to mariadb)

J0WI:


J0WI:
Not sure how this will work on hosts with less than 512MB memory

PR #474, opened 2018-09-29
[WIP] Install smbclient dependencies to enable SMB/CIFS External storage

xtrasimplicity:
I've simply added `smbclient` as a dependency to each `Dockerfile` (using the templates, of course), to allow SMB/CIFS shares to be used under the `External Storage` configuration.

### DIFF
--- a/12.0/apache/Dockerfile
+++ b/12.0/apache/Dockerfile
@@ -114,7 +114,7 @@ RUN set -ex; \
         dirmngr \
     "; \
     apt-get update; \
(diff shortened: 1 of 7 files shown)


tilosp:
See: https://github.com/nextcloud/docker/pull/22

xtrasimplicity:
Ahh, thanks @tilosp. I must admit, I didn't see the `. examples` folder until after I submitted this PR.

Would it be worth adding some of the example dockerfiles, or at least the `full` example, to Docker Hub, as well?

PR #477, opened 2018-10-01
Add all files in one copy instruction

tilosp:
This saves one image layer and allows us to more easily add new files later.


tilosp:
> I'm not sure about [2d45a16](https://github.com/nextcloud/docker/commit/2d45a1675918378316251e4d9d76a1a43bb95287). It looks quite error-prone and doesn't add any benefits.
> Could you explain why you want to change this?

@J0WI I don't like the idea of adding a new `COPY` instruction (and thereby a new image layer) for each file we want to add to the image.

@tianon what would you recommend?

tianon:
The main issue I've seen with `COPY xxx /` is that it tramples permissions pretty aggressively (and any attempts to fix the resulting permissions mess in later layers will cause duplication and thus even more layer bloat).

As a matter of personal taste, I find it annoying that it creates a hierarchy of directories on the host that have to be navigated in order to see what's included, whereas individual files are very obvious, and the individual `COPY` instructions make it very clear what's being copied and where they end up, but that's all merely personal preference.

The layers issue shouldn't matter too much, especially since they'll all be relatively small layers and Docker will usually transmit them in parallel (so technically, it could actually end up being _faster_ to transmit lots of smaller layers than a smaller number of large layers, given the way Docker downloads them).

tianon:
(Doing it with one big directory that gets `COPY`'d also encourages _more_ file additions, while we typically try to strive for the opposite since any files necessary at this level really probably ought to be considered for inclusion in the official software distribution upstream.)

tilosp:
ok i will close this then

PR #478, opened 2018-10-01
Use exclude file for the updater

tilosp:
This makes it easier for child images to customize the updater.
For example adding another app dir. (Like [David-Development/nextcloud-dev-docker](https://github.com/David-Development/nextcloud-dev-docker))

### DIFF
--- a/12.0/apache/Dockerfile
+++ b/12.0/apache/Dockerfile
@@ -135,7 +135,7 @@ RUN set -ex; \
     apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false $fetchDeps; \
     rm -rf /var/lib/apt/lists/*
 
-COPY *.sh /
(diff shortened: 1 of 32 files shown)


J0WI:
I'm not sure about 2d45a16. It looks quite error-prone and doesn't add any benefits.
Could you explain why you want to change this?

tilosp:
> I'm not sure about [2d45a16](https://github.com/nextcloud/docker/commit/2d45a1675918378316251e4d9d76a1a43bb95287). It looks quite error-prone and doesn't add any benefits.
> Could you explain why you want to change this?

@J0WI let's discuss that part in #477

tilosp:
@J0WI i have removed [2d45a16](https://github.com/nextcloud/docker/commit/2d45a1675918378316251e4d9d76a1a43bb95287)

PR #480, opened 2018-10-02
added retry loop and decreased sleep time

SnowMB:
Discussion in https://github.com/docker-library/official-images/pull/4897


yosifkit:
This looks good to me! Thanks for adding this.

PR #487, opened 2018-10-06
Default data dir v1

JonasAlfredsson:
I have two improvement proposals and this is v1. Please read v2 as well. 

The admin manual recommends to place the data directory outside the web root as noted in [this post](https://help.nextcloud.com/t/admin-manual-recommends-placing-data-directory-outside-web-root-why/22225). Such a directory was created in pull request #307, but is never used. 

This pull request defaults the data directory on install to the folder in #307, which requires the created volume to be mounted already at /var/www to make sure nothing is lost. 

This requires only one volume but does not entirely solve the web root issue. Furthermore, mounting the entire /var/www folder might cause compatibility trouble for users running multiple docker services behind one nginx proxy.

The v2 solutions handles it with two volumes, but is a bit messier.

### DIFF
--- a/.examples/docker-compose/insecure/mariadb-cron-redis/apache/docker-compose.yml
+++ b/.examples/docker-compose/insecure/mariadb-cron-redis/apache/docker-compose.yml
@@ -22,7 +22,7 @@ services:
     ports:
       - 8080:80
     volumes:
-      - nextcloud:/var/www/html
+      - nextcloud:/var/www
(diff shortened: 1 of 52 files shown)


JonasAlfredsson:
This is actually bad. 
After some further testing of this solution I see now that this breaks backwards compatibility with the current dockerfile as users with already created named volumes will fail when mounting the new path in docker-compose.

PR #500, opened 2018-10-12
add a mechanism to set trusted domains on install

### LINKED ISSUE #469: set nextcloud trusted domain via environment variable
Hello

I'm struggling to setup nextcloud to be accessible from the outside world because I can not setup trusted_domains within config.php file in my docker instance.

I've tried several combinations mounting remote routes (nfs) but no luck I end up having exceptions related to logging system which seems to be happening on several versions with no clear explanation about what's wrong.

said that, do you guys thing would it be possible to create (if it does not exist already) the configuration required using an environment variable?

something like: NEXTCLOUD_HOSTNAME=xxx.yyy.zzz:port 

thanks!

blizzz:
Resolves #469 

Optionally, `NEXTCLOUD_TRUSTED_DOMAINS` can be specified. After sucessfull install, occ is being used to set the provided domains names. 

for example:

```
$ sudo docker run -e SQLITE_DATABASE=foobar -e NEXTCLOUD_ADMIN_USER=adm -e NEXTCLOUD_ADMIN_PASSWORD=dma -e NEXTCLOUD_TRUSTED_DOMAINS=foo.bar\ barfoo.com f7469ede5323
Initializing nextcloud 14.0.2.2 ...
Initializing finished
New nextcloud instance
Installing with SQLite database
starting nexcloud installation
creating sqlite db
Nextcloud was successfully installed
setting trusted domains…
System config value trusted_domains => 1 set to string foo.bar
System config value trusted_domains => 2 set to string barfoo.com
AH00558: apache2: Could not reliably determine the server's fully qualified domain name, using 172.17.0.2. Set the 'ServerName' directive globally to suppress this message
AH00558: apache2: Could not reliably determine the server's fully qualified domain name, using 172.17.0.2. Set the 'ServerName' directive globally to suppress this message
[Fri Oct 12 12:46:54.210310 2018] [mpm_prefork:notice] [pid 1] AH00163: Apache/2.4.25 (Debian) PHP/7.2.10 configured -- resuming normal operations
[Fri Oct 12 12:46:54.210354 2018] [core:notice] [pid 1] AH00094: Command line: 'apache2 -D FOREGROUND'
```

```
$ sudo docker exec -it c0cfa5a94c19 /bin/bash
root@c0cfa5a94c19:/var/www/html# cat config/config.php 
<?php
$CONFIG = array (
…
  'trusted_domains' => 
  array (
    0 => 'localhost',
    1 => 'foo.bar',
    2 => 'barfoo.com',
  ),
…
);
```

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -88,7 +88,7 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ]; then
                     install_options=$install_options' --database pgsql --database-name "$POSTGRES_DB" --database-user "$POSTGRES_USER" --database-pass "$POSTGRES_PASSWORD" --database-host "$POSTGRES_HOST"'
                     install=true
                 fi
-                
+
                 if [ "$install" = true ]; then
                     echo "starting nexcloud installation"
                     max_retries=10
@@ -102,6 +102,15 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ]; then
                     if [ "$try" -gt "$max_retries" ]; then
                         echo "installing of nextcloud failed!"
                         exit 1
+                    fi
+                    if [ -n "${NEXTCLOUD_TRUSTED_DOMAINS+x}" ]; then
+                        echo "setting trusted domains…"
+                        NC_TRUSTED_DOMAIN_IDX=1
+                        for DOMAIN in $NEXTCLOUD_TRUSTED_DOMAINS ; do
+                            DOMAIN=$(echo "$DOMAIN" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+                            run_as "php /var/www/html/occ config:system:set trusted_domains $NC_TRUSTED_DOMAIN_IDX --value=$DOMAIN"
+                            NC_TRUSTED_DOMAIN_IDX=$(($NC_TRUSTED_DOMAIN_IDX+1))
+                        done
                     fi
                 else
                     echo "running web-based installer on first connect!"
--- a/12.0/apache/entrypoint.sh
+++ b/12.0/apache/entrypoint.sh
@@ -88,7 +88,7 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ]; then
                     install_options=$install_options' --database pgsql --database-name "$POSTGRES_DB" --database-user "$POSTGRES_USER" --database-pass "$POSTGRES_PASSWORD" --database-host "$POSTGRES_HOST"'
                     install=true
                 fi
-                
+
                 if [ "$install" = true ]; then
                     echo "starting nexcloud installation"
                     max_retries=10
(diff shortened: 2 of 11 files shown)


SnowMB:
hey thanks for your contribution. Unfortunately the files you edited aren't meant to be edited directly. They are auto generated from the `docker-entrypoint.sh` in the root directory via the `update.sh` script. You should apply your changes in that file and run `./update.sh` afterwards.

blizzz:
ah, good to know :) I'll adjust next week, thanks for the pointer!

blizzz:
@SnowMB done

blizzz:
> There is a version bump in the nextcloud 14 image.

14.0.3 is the currently released version.  I don't know how updates are picked up here, manually or (semi-)automatically?

tilosp:
A simple rebase onto master should fix this (You branched before https://github.com/nextcloud/docker/commit/4d2dbd6ea0faaffe64cb5e560c6a383b87e7b261)

blizzz:
> A simple rebase onto master should fix this (You branched before 4d2dbd6)

Check!

docker-entrypoint.sh
```diff
@@ -103,6 +103,15 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ]; then
                         echo "installing of nextcloud failed!"
                         exit 1
                     fi
+                    if [ -n "${NEXTCLOUD_TRUSTED_DOMAINS+x}" ]; then
+                        echo "setting trusted domains…"
+                        NC_TRUSTED_DOMAIN_IDX=1
+                        for DOMAIN in $NEXTCLOUD_TRUSTED_DOMAINS ; do
+                            DOMAIN=$(echo "$DOMAIN" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+                            run_as "php /var/www/html/occ config:system:set trusted_domains $NC_TRUSTED_DOMAIN_IDX --value=$DOMAIN"
```
tilosp: This will break if you put special characters into this var. (like `"` or `$`)
But it shouldn't relay matter because a valid domain name can't contain these characters.

tilosp:
Thanks for your contribution :)

mvillarejo:
thank you for this contribution!

bernd-protofy:
my configuration does not work, could somebody help me?

this is my docker-compose:
`version: '3'

services:
  db:
    image: mariadb
    command: --transaction-isolation=READ-COMMITTED --binlog-format=ROW
    restart: always
    volumes:
      - db:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=supersecurepassword
    env_file:
      - db.env

  app:
    image: nextcloud:apache
    restart: always
    ports:
      - 80:80
      - 443:443
    volumes:
      - nextcloud:/var/www/html
      - /Volumes/Second/nextData:/data
    environment:
      - MYSQL_HOST=db
    env_file:
      - db.env
    depends_on:
      - db

volumes:
  db:
  nextcloud:`

and my db.env

`MYSQL_PASSWORD=superstrongpassword
MYSQL_DATABASE=nextcloud
MYSQL_USER=nextcloud
NEXTCLOUD_DATA_DIR=/data
NEXTCLOUD_TRUSTED_DOMAINS=mydomain.tld`

I see the following lines in the terminal with a fresh install:
`Initializing nextcloud 14.0.3.0 ...
Initializing finished
New nextcloud instance`

the install script (https://github.com/nextcloud/docker/blob/424364e2e10a9d6e1a31e6659e2149aac1f1c772/14.0/apache/entrypoint.sh) does not go to:
`if [ -n "${NEXTCLOUD_TRUSTED_DOMAINS+x}" ]; then
                        echo "setting trusted domains…"`

is mariadb not supported?
thx for any help!

Issue #491, opened 2018-10-08
custom "full" build


XtraLarge:
Hello,

since version 14.1 a custom buld using the full example didn't work anymore. In the "full" example the CMD was changed from "apache-forground" to "/usr/bin/supervisord" so the installation or update process in the "/entrypoint.sh" script will never be triggered.
For my opinion the must be a third if clause so "supervisord" is also a trigger in the "/entrypoint.sh" script next "apache" and "php-fpm".

my solution is: (please be critical by checking it) file: [/entrypoint.sh]
```
#!/bin/bash
....
if [[ "$1" == *apache* ]] || [[ "$1" == *supervisord* ]] || [[ "$1" == *php-fpm* ]]; then
```

best regards
Willi

J0WI:
Thanks for the hint, so we have to update our examples.
I would prefer using separate containers for `cron` and `libreoffice` instead of `supervisord`.

SnowMB:
Great catch.

Modifying the entry point is not really practical because we can't add every other solution that users use to it. I'm thinking adding an environment variable like FORCE_AUTO_INSTALL might work.

Cc @tilosp.

tilosp:
Yes a variable makes sense.
I would call the var something like NEXTCLOUD_UPDATER, and you can set it to true or false. If it's not set we can use the already existing detection based on the command.
@SnowMB calling it AUTO_INSTALL is kind of misleading because it would also enable or disable the update script.

SnowMB:
Sure. I just realized that the latest change might have broken the update for existing deployments. That's really bad and should be fixed asap. 

For the solution I am stuck here:

Introducing a variable and if it's 
not present but the command is apache or php-fpm still doing updates is confusing. 

Maybe relying on the environment variable and taking out the check would be clearer. It's also easier to document.

But on the other hand this would brake the update for all users. So not really an option.

Going the other way round and doing an opt out style would break the docker library echo test again.

We might have to move the checks for existing instalation and the used command a bit.

Any thoughts?

XA21X:
Just got caught out by this. I thought the 14.0.1 image was broken but after 14.0.2 also didn't take effect, I discovered the change in the entrypoint script. Note that some of the included examples use supervisord.

SnowMB:
@tilsop yeah I came to the conclusion that this might be the best solution after all. If nobody beats me to it I'll try to fix this this weekend.

benwaffle:
[...]

PR #501, opened 2018-10-12
run upgrades when using supervisord

benwaffle:
All of the `cron` and `full` examples in `.examples` run `supervisord` as the command.

Also, not sure if this is an issue, but using `|` with `expr` is a GNU extension. The test seems to run fine on alpine linux

### DIFF
--- a/12.0/apache/entrypoint.sh
+++ b/12.0/apache/entrypoint.sh
@@ -19,7 +19,7 @@ run_as() {
     fi
 }
 
-if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ]; then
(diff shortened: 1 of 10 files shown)


tilosp:
See: #491
We have decided to implement a environment variable to fix this.

Aneurin:
I was about to submit a PR for my much simpler solution to this problem - if this is going to be solved differently then I won't bother, but it's a lot more straightforward: in supervisord.conf, `command=apache2-foreground` should be `command=/entrypoint.sh apache2-foreground`. This way the supervisord invocation matches the standard invocation, without needing any special-casing.

Issue #503, opened 2018-10-13
Deadlock Problem with mariadb:latest


Tockra:
Hi,

this is related with #494 . 
There seems to be problems with the latest mariadb image. 

If you use mariadb:10.3 there is no problem, if you use mariadb:latest you get errors like in #494 

Example Sync-Error:
```
 13.10.2018 11:56:10,                     Dreamedit/vusolose/userbouquet.dbe00.tv_org,            Nextcloud\Projekte,Error transferring http://homemedia/remote.php/dav/files/Tockra/Documents/Dreamedit/vusolose/userbouquet.dbe00.tv_org - server replied: Internal Server Error (An exception occurred while executing 'INSERT INTO `filecache` (`mimepart`,`mimetype`,`mtime`,`size`,`etag`,`storage_mtime`,`permissions`,`parent`,`checksum`,`path_hash`,`path`,`name`,`storage`) SELECT ?,?,?,?,?,?,?,?,?,?,?,?,? FROM `filecache` WHERE `storage` = ? AND `path_hash` = ? HAVING COUNT(*) = 0' with params ["3", "15", 1539424572, 223, "5ebc52ee3af39eb7ed1bdc98b38b0580", 1539424572, 27, 172, "", "a123a247800228731b5c866c1faf6075", "files\/Documents\/Dreamedit\/vusolose\/userbouquet.dbe00.tv_org", "userbouquet.dbe00.tv_org", 2, 2, "a123a247800228731b5c866c1faf6075"]:

SQLSTATE[40001]: Serialization failure: 1213 Deadlock found when trying to get lock; try restarting transaction)
[...]

PR #505, opened 2018-10-15
Update insecure/mariadb-cron-redis/fpm/docker-compose.yml

Tockra:
- Changed version to 10.3 because: #503
- Added "--log-bin=mysqld-bin" caused of startup Warning.


pierreozoux:
Could you put #505 in this PR too? Thanks.

PR #514, opened 2018-10-15
Update with-nginx-proxy/mariadb-cron-redis/fpm/docker-compose.yml

Tockra:
- Changed version to 10.3 because: #503
- Added "--log-bin=mysqld-bin" caused of startup Warning.


pierreozoux:
Can you put #505 here too? Thanks!

Tockra:
Okay, sorry I'm not so familar with git and the problem at first was I added the changes via github webpanel. 
I hope I fixed it right ;)

pierreozoux:
No ptoblems, we are all learning ;)

I recommend reading https://nvie.com/posts/a-successful-git-branching-model/
In this case it is a feature branch (or maybe a bug branch)

Coming back to our PR:
Pinning the version is always a good thing, but it is time consuming to maintain.
Except if we have automation to always update to latest and check it is working as expected.
We already have CI integration with docker, but it merely test the docker image itself.
We could have a CICD for docker compose too.
It looks like one more argument to have it in its own repo.

This bin log argument, can you point me to the documents that states that it was not needed before and is needed now? I'm curious.

Thanks a lot for your contribution.

I'm curious about the thoughts of other maintainers?
(On mobile, I can't ping people..)

Tockra:
> 
> 
> No ptoblems, we are all learning ;)
> 
> I recommend reading https://nvie.com/posts/a-successful-git-branching-model/
> In this case it is a feature branch (or maybe a bug branch)
> 
> Coming back to our PR:
> Pinning the version is always a good thing, but it is time consuming to maintain.
> Except if we have automation to always update to latest and check it is working as expected.
> We already have CI integration with docker, but it merely test the docker image itself.
> We could have a CICD for docker compose too.
> It looks like one more argument to have it in its own repo.
> 
> This bin log argument, can you point me to the documents that states that it was not needed before and is needed now? I'm curious.
> 
> Thanks a lot for your contribution.
> 
> I'm curious about the thoughts of other maintainers?
> (On mobile, I can't ping people..)

If you don't use log-bin you get this warning:
` [Warning] You need to use --log-bin to make --binlog-format work.
`

pierreozoux:
Ok, so this fixes the warning, sounds legit to me.
About the pinning of the mysql version, it is just because the upper version doesn't work?
Do you know the cause?

Tockra:
> Do you know the cause?

I don't know why it doesn't work. But I think to pin the version until it is fixed is a good idea.

Tockra:
Merge or not merge that is the question!?

J0WI:
`10.3` is just an alias for `latest`:
https://github.com/docker-library/official-images/blob/db644c0d1376cbbc12fb1c3ffbacc268b77947f3/library/mariadb#L12

Tockra:
> 
> 
> `10.3` is just an alias for `latest`:
> https://github.com/docker-library/official-images/blob/db644c0d1376cbbc12fb1c3ffbacc268b77947f3/library/mariadb#L12

This isn't true: https://hub.docker.com/_/mariadb

J0WI:
Yes, meanwhile `10.4` is out, so now it's an alias for that version (which is compatible with NextCloud).

Tockra:
> 
> 
> Yes, meanwhile `10.4` is out, so now it's an alias for that version (which is compatible with NextCloud).

I remember that at the moment I posted this Issue 10.3 wasn't equal latest.
Are you sure you are right? Did you try to sync 1 GB or more small files via Nextcloud with a fast connection (1Gbit/s (LAN/Ethernet)) to your server? In this cases I got Lockes even before 1 month

J0WI:
`10.3`was stable since https://github.com/docker-library/official-images/commit/858b20ebe5f116362e8ace0e3c4e71d2d9070114#diff-0a3cf4f4c38ee0f991485508f73e7249 including https://github.com/docker-library/official-images/blob/165fd39e893eaff933eebbf67d18160f3edaed7b/library/mariadb#L7

Are you using Redis for file locking as recommend?
https://docs.nextcloud.com/server/stable/admin_manual/configuration_files/files_locking_transactional.html

PR #515, opened 2018-10-15
Add NEXTCLOUD_UPDATE for containers started with non-default commands 

### LINKED ISSUE #491: custom "full" build

SnowMB:
Todo:
- [x] Add environment variable
- [x] Add description to readme
- [x] Fix examples

fixes #491

### DIFF
--- a/.examples/README.md
+++ b/.examples/README.md
@@ -13,7 +13,7 @@ The Dockerfiles use the default images as base image and build on top of it.
 
 Example | Description
 ------- | -------
-[cron](https://github.com/nextcloud/docker/tree/master/.examples/dockerfiles/cron) | uses supervisor to run the cron job inside the container (so no extra container is needed). This image runs `supervisord`. To enable auto install and update run images like this with an additional environment variable:  `NEXTCLOUD_UPDATE=1`. See [here](https://github.com/nextcloud/docker/blob/master/README.md#auto-configuration-via-environment-variables) for more information.
+[cron](https://github.com/nextcloud/docker/tree/master/.examples/dockerfiles/cron) | uses supervisor to run the cron job inside the container (so no extra container is needed). This image runs `supervisord` to start nextcloud and cron as two seperate processes inside the container.
 [imap](https://github.com/nextcloud/docker/tree/master/.examples/dockerfiles/imap) | adds dependencies required to authenticate users via imap
 [smb](https://github.com/nextcloud/docker/tree/master/.examples/dockerfiles/smb) | adds dependencies required to use smb shares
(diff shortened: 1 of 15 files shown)


.examples/README.md
```diff
@@ -13,7 +13,7 @@ The Dockerfiles use the default images as base image and build on top of it.
 
 Example | Description
 ------- | -------
-[cron](https://github.com/nextcloud/docker/tree/master/.examples/dockerfiles/cron) | uses supervisor to run the cron job inside the container (so no extra container is needed).
+[cron](https://github.com/nextcloud/docker/tree/master/.examples/dockerfiles/cron) | uses supervisor to run the cron job inside the container (so no extra container is needed). This image runs `supervisord`. To enable auto install and update run images like this with an additional environment variable:  `NEXTCLOUD_UPDATE=1`. See [here](https://github.com/nextcloud/docker/blob/master/README.md#auto-configuration-via-environment-variables) for more information.
```
tilosp: It would also make sense to set it for the user in the Dockerfile

SnowMB: Can it then be overwritten by explicitly setting another value?

tilosp: I am not sure. I would only add it to the supervisor example. And if a user doesn't want it he can simply comment it out before building his custom image.

tilosp: And the supervisor example is basically useless without the updater working.

SnowMB:
@tilosp done. Changed the wording in the readme a little bit to clarify the use of the variable.

tilosp:
Is there some reason why this is still wip?

tilosp:
https://github.com/nextcloud/docker/blob/3838458843b9115179791f693e7afa65d6e22eed/.travis/test-example-dockerfiles.sh#L16
this is why.

But we can simply remove the official-images test for the examples. It's not really necessary.

SnowMB:
Ah. Does that mean this also runs into the echo test and fails because of additional outputs?

tilosp:
> Ah. Does that mean this also runs into the echo test and fails because of additional outputs?

yes

SnowMB:
Ok removed it. Thanks for your help.

PR #518, opened 2018-10-16
Change set criteria for environment variables

JonasAlfredsson:
Appending "+x" to the environment variables, in these cases, allows them to bet set to empty strings.
The if-statements here would then return true for empty variables ([examples](https://serverfault.com/a/382740)) which does not make sense at the places I changed them.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -60,29 +60,29 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
             if [ -n "${NEXTCLOUD_ADMIN_USER+x}" ] && [ -n "${NEXTCLOUD_ADMIN_PASSWORD+x}" ]; then
                 # shellcheck disable=SC2016
                 install_options='-n --admin-user "$NEXTCLOUD_ADMIN_USER" --admin-pass "$NEXTCLOUD_ADMIN_PASSWORD"'
-                if [ -n "${NEXTCLOUD_TABLE_PREFIX+x}" ]; then
+                if [ -n "${NEXTCLOUD_TABLE_PREFIX}" ]; then
                     # shellcheck disable=SC2016
                     install_options=$install_options' --database-table-prefix "$NEXTCLOUD_TABLE_PREFIX"'
                 else
                     install_options=$install_options' --database-table-prefix ""'
                 fi
-                if [ -n "${NEXTCLOUD_DATA_DIR+x}" ]; then
+                if [ -n "${NEXTCLOUD_DATA_DIR}" ]; then
                     # shellcheck disable=SC2016
                     install_options=$install_options' --data-dir "$NEXTCLOUD_DATA_DIR"'
                 fi
 
                 install=false
-                if [  -n "${SQLITE_DATABASE+x}" ]; then
+                if [  -n "${SQLITE_DATABASE}" ]; then
                     echo "Installing with SQLite database"
                     # shellcheck disable=SC2016
                     install_options=$install_options' --database-name "$SQLITE_DATABASE"'
                     install=true
-                elif [ -n "${MYSQL_DATABASE+x}" ] && [ -n "${MYSQL_USER+x}" ] && [ -n "${MYSQL_PASSWORD+x}" ] && [ -n "${MYSQL_HOST+x}" ]; then
+                elif [ -n "${MYSQL_DATABASE}" ] && [ -n "${MYSQL_USER}" ] && [ -n "${MYSQL_PASSWORD}" ] && [ -n "${MYSQL_HOST}" ]; then
                     echo "Installing with MySQL database"
                     # shellcheck disable=SC2016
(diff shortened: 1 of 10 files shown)


docker-entrypoint.sh
```diff
@@ -60,29 +60,29 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
             if [ -n "${NEXTCLOUD_ADMIN_USER+x}" ] && [ -n "${NEXTCLOUD_ADMIN_PASSWORD+x}" ]; then
                 # shellcheck disable=SC2016
                 install_options='-n --admin-user "$NEXTCLOUD_ADMIN_USER" --admin-pass "$NEXTCLOUD_ADMIN_PASSWORD"'
-                if [ -n "${NEXTCLOUD_TABLE_PREFIX+x}" ]; then
+                if [ -n "${NEXTCLOUD_TABLE_PREFIX}" ]; then
```
SnowMB: An emtpy table prefix is valid. It even is the default if nothing is present (see else clause).

Also this changes do not work for optional parameters if the variable is not set:

```shell
/entrypoint.sh: 63: /entrypoint.sh: NEXTCLOUD_TABLE_PREFIX: parameter not set
```

docker-entrypoint.sh
```diff
@@ -60,29 +60,29 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
...
                     # shellcheck disable=SC2016
                     install_options=$install_options' --database-table-prefix "$NEXTCLOUD_TABLE_PREFIX"'
                 else
                     install_options=$install_options' --database-table-prefix ""'
                 fi
-                if [ -n "${NEXTCLOUD_DATA_DIR+x}" ]; then
+                if [ -n "${NEXTCLOUD_DATA_DIR}" ]; then
```
SnowMB: Also optional so does not work here.

docker-entrypoint.sh
```diff
@@ -60,29 +60,29 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
...
                     # shellcheck disable=SC2016
                     install_options=$install_options' --data-dir "$NEXTCLOUD_DATA_DIR"'
                 fi
 
                 install=false
-                if [  -n "${SQLITE_DATABASE+x}" ]; then
+                if [  -n "${SQLITE_DATABASE}" ]; then
```
SnowMB: Also optional, as sqlite might not be used.

pierreozoux:
Thanks for the PR, we usually do this in 2 commints.
 1. the change in the automation (here docker-entrypoint.sh
 2. run the automation to change all the other files.
If you could update your PR, this would be amazing :)

Thre PR looks good to me.

JonasAlfredsson:
My apologies for not doing the "update.sh" script in a separate commit, that is so obvious now that you mention it. :)

However, I am still not convinced that it is a bad idea to be more strict already in the entrypoint. Doing something like the compose below (maybe by mistake) will ignore the fully configured mariadb and go directly for the SQLite database install. Requiring you to explicitly set the environment variables will also allow you to create an "[.env](https://docs.docker.com/compose/env-file/)" file in the [top directory](https://docs.docker.com/compose/environment-variables/#the-env-file) where you can easily list all the possible environment variables you can set, without having to find them all in the documentation. This might also allow sane default values for people just wanting to have something up and running quickly. 


``` Dockerfile
version: '3'

services:
  next:
    image: nextcloud:fpm
    environment:
      - SQLITE_DATABASE=
      - MYSQL_DATABASE=example
      - MYSQL_USER=user
      - MYSQL_PASSWORD=password
      - MYSQL_HOST=db:3306
      - NEXTCLOUD_ADMIN_USER=nc_user
      - NEXTCLOUD_ADMIN_PASSWORD=pass
```

tilosp:
I don't think that treating an empty string as not set is a good idea.
Especially if it sometimes makes sense to be able to set an empty string.

If you want to list all possible variables in some env file you can already do this by putting a `#` before the unused ones. This makes it also clear to someone else looking at the files which vars are in use.

@SnowMB @J0WI @pierreozoux  what do you think?

JonasAlfredsson:
I see your line of thinking that if a user takes the time to actually type it out they should know what they are doing. It was just me feeling a bit queasy accepting empty strings at some places. 
Furthermore, I do believe it would cause more confusion than good if you would implement different acceptance levels inside the entryscript, better to keep it the same for all variables throughout. 

Thank you for your time discussing this.

SnowMB:
I see that it is not one hundred percent foolproof working with the different database connections. Right now sqlite is prioritized if more than one connection is present. But I think this rarely is an issue. I feel fine to just sit on this and wait if there are really some users getting this wrong.

What could be a little bit cleaner is explicitly setting the desired database type through another environment variable. (like the maintenance:install command does) 

For the other variables this is not really an issue. Like I said an empty string might be a valid value and if it is not the auto install fails. If variables are missing the auto install is not triggered.

PR #292, opened 2018-03-23
Config redis and smtp using env vars.

rodrigoaguilera:
Closes #194

PR #522, opened 2018-10-24
Set Redis via environment variables

marceljd:
Copy of the Redis part of #292 without the SMTP part. Redis has to be set at boot time, SMTP can be set through the Nextcloud webinterface. Tested with Apache / 14.0.3.

@rodrigoaguilera @unteem

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ The install and update script is only triggered when a default command is used (
 
 - `NEXTCLOUD_UPDATE` (default: _0_)
 
-If you want to use Redis you have to create a seperate Redis container in your setup / in your docker-compose file. To inform Nextcloud about the Redis container add:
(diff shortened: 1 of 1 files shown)


marceljd:
Did the test fail because of trying to change a file that was not merged / approved yet? I changed the readme.md and updated later with a link to the Redis official repository.

rodrigoaguilera:
If I remember right you need to run update.sh to modify the images for all the version variants

SnowMB:
@rodrigoaguilera is right. You have to run `./update.sh` so that all the images get updated.

marceljd:
How do I run ./update.sh? Download the entire repository into linux and run it, then upload everything again? Or is there a better / quicker way? Sorry I have to ask, I could not find out how to do this...

Issue #401, opened 2018-07-27
Run Nextcloud from Subdirectory


studio315b:
How do you tell nextcloud that it's in a sub-directory? Currently each service on my machine is in its own docker instance, and I use Traefik to expose them to different sub-directories on my server:

mydomain/service1
mydomain/service2

## Current Behaviour
all links are "/index.php/*"

## Desired Behavior
all links are "/nextcloud/index.php/*"

datenfahrt:
if I understood you correctly - you have to adjust your nextcloud configuration:

`'overwritewebroot'`

Look this:

https://docs.nextcloud.com/server/13/admin_manual/configuration_server/reverse_proxy_configuration.html#overwrite-parameters

regards

somkun:
This is what I was looking for. Is there a way to set this setting via docker, or do I need to modify my version of the image?

udl:
Depending on how you run the docker container. If you mount your filesystem in /var/www/html, for example, you don't need to rebuild the container. You can find the config file in the mounted directory.

naramski:
There are two parameters to change this : overwritehost and overwriteprotocol.

I had the same problem and I have created a fix for this issue (Pull request : #527).

Hope it helps,

naramski:
I had a similar problem and I have created a fix for this issue (Pull request : #527) that contains the an ENV variable NEXTCOULD_OVERWRITEWEBROOT to help define "subdirectories". 

Hope it helps,

omontigny:
Hi,
I am sorry maybe i didn’t get something but in 16.0.3 version , I still have troubles to access to nextcloud with subdirectories as http://myserver/nextcloud/login

anything a do and every variable i use ( like
overwrite.cli.url
'htaccess.RewriteBase' ...)
 
i get a redirection to http://myserver/login

except if a use the variable 
'overwritewebroot' => '/nextcloud’

but in this case i have ERR_TOO_MANY_REDIRECTS error

Is there something i missed ?

Thank you
Best regards

johnjore:
Hi all!
I have the same issue. Migrating to using docker and need nextcloud to use the URL /nextcloud as I have multiple web services all using different folders.

I too get the ERR_TOO_MANY_REDIRECTS error

I've also tried fiddling with the 
`'htaccess.RewriteBase' => '/nextcloud',`
setting as well as RewriteBase / in .htaccess no success. (On some occasions all files were deleted when docker container was started)

[...]

PR #527, opened 2018-10-26
Add two parameters when Nextcloud is behind a proxy

naramski:
Added support for params overwriteprotocol and overwritehost when Nextcloud is behind a proxy and fails to detect hostname and protocol. 

(see https://docs.nextcloud.com/server/12/admin_manual/configuration_server/reverse_proxy_configuration.html)


marceljd:
I tried the same thing 2 days ago: you have to run the update.sh script before the commit, then the script will run to test the other images.

naramski:
Thanks @marceljd it worked :)

J0WI:
You can also just set some headers, e.g. in nginx:
```nginx
proxy_set_header Host $host;
proxy_set_header X-Real-IP $remote_addr;
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header X-Forwarded-Proto $scheme;
```

J0WI:
@naramski does https://github.com/nextcloud/docker/pull/527#issuecomment-434492082 work for you?

naramski:
@J0WI well I don't use Nginx and I prefer to have a standalone configuration for the nextcloud container itself.

naramski:
So I guess it work, but it doesn't solve my problem. I still think that adding the new parameters doesn't hurt.

J0WI:
@naramski NC 12 is no longer supported, but NC 15 have been added. Could you rebase your changes?
@tilosp this addresses a bunch of (#401, #528, #577) reasonable use cases. Do you agree with the new ENVs?
https://github.com/docker-library/docs/blob/master/nextcloud/content.md needs to be updated afterwards.

naramski:
@J0WI I tried to rebase, but it seems that I have messed something :(.

J0WI:
Would you like to open a new, clean PR?

tilosp:
looks sane to me :+1: 
@J0WI are you sure that this fixes #528?
I am not sure if nextcloud actually updates the apache config if you set the overwriteprotocol to https.

J0WI:
I haven't verified this so far. It would be great to get some feedback from @jkaberg or @funkypenguin.

pgera:
I had to manually add 'overwriteprotocol' => 'https' for a different issue (https://github.com/nextcloud/server/issues/6767).

marceljd:
I have investigated these issues and to solve a these proxy related problems on a new standard Apache installation this should be added. I have tried this with my Traefik proxy, can anyone test this on Nginx?

in entrypoint.sh (this will resolve the .well-known errors):
```
sed -i "s/caldav /caldav https:\/\/%{SERVER_NAME}/" /var/www/html/.htaccess
sed -i "s/carddav /carddav https:\/\/%{SERVER_NAME}/" /var/www/html/.htaccess
sed -i "s/webfinger /webfinger https:\/\/%{SERVER_NAME}/" /var/www/html/.htaccess
```

add a new file proxy.config.php in the .config (no HTTPS error in Password app and get the real IP adresses instead of the proxy IP). This is faster and easier than with OCC commands.
```
<?php
if (getenv('PROXY_IP')) {
    $CONFIG = array (
        'trusted_proxies'   => getenv('PROXY_IP'),
        'overwriteprotocol' => 'https',
    );
}
```

There will be one extra environment variable: PROXY_IP with the ip address of the proxy.

marceljd:
Some more tests and some more conclusions:

**Apache running on 443 / SSL**
The root cause of many issues is that Apache is configured for http (port 80) only. Nextcloud is designed to run on https (SSL port 443) and all built in security checks are a kind of expecting that it runs on SSL, and it should. In my test-environment I could get rid of the .well-known and HSTS security warnings this way. Besides that the apps that require SSL (Password, don't know if there are others) will work without warnings. The solutions I provided before are fighting the symptoms, the root cause is the Apache configuration. 

I will make a seperate PR for this.

**Proxy IP**
The second change is to introduce an environment variable PROXY_IP (or IP's, could be multiple). This will tell Nextcloud that there is a proxy active. NC will autodiscover and automatically set the hostname in the config.php file, and (not tested) probably the path if NC is installed in a subdirectory. NC will also see the real IP address of the request, for Traefik there is no extra configuration necessary.

`'trusted_proxies' => array('203.0.113.45', '198.51.100.128'),`

There is another enviroment variable that could tell NC what headers to expect, perhaps this is needed for Nginx or HAproxy. Traefik works without this setting.

`'forwarded_for_headers' => array('HTTP_X_FORWARDED', 'HTTP_FORWARDED_FOR'),`

I will make a seperate PR for this too. 

With both solutions the Apache versions should work out-of-the-box. I do think this needs some testing on different setups to be sure it works on all proxies and nothing breaks on existing installations.

marceljd:
See #644 for the SSL on Apache PR. I will create the trusted proxy PR next week, with both of them in place the SSL and proxy related errors are gone.

GuyPaddock:
I know that this PR is focused on a solution at application-level, but in case it's useful to anyone using NGinx we just implemented a fix for this issue. In our fix, NGinx switches to relying on a new variable that is based on the `X-Forwarded-Proto` header (if it is available). 

Here's the relevant commit:
https://github.com/GuyPaddock/inveniem-nextcloud-azure/commit/8f1619259a6e1fd1ed3c6e06ffc4c1877923b028

In our case, the proxy is the NGinx ingress controller for Kubernetes.

15.0/fpm/Dockerfile
```diff
@@ -101,7 +101,11 @@ RUN { \
 VOLUME /var/www/html
 
 
+<<<<<<< HEAD:15.0/fpm/Dockerfile
```
bencomp: This is a git conflict.

15.0/apache/Dockerfile
```diff
@@ -109,7 +109,11 @@ RUN a2enmod rewrite remoteip ;\
     } > /etc/apache2/conf-available/remoteip.conf;\
     a2enconf remoteip
 
+<<<<<<< HEAD:15.0/apache/Dockerfile
```
bencomp: This is a git conflict.

15.0/fpm-alpine/Dockerfile
```diff
@@ -90,7 +90,11 @@ RUN { \
 VOLUME /var/www/html
 
 
+<<<<<<< HEAD:15.0/fpm-alpine/Dockerfile
```
bencomp: This is a git conflict.

GuyPaddock:
Following-up on [my earlier comment](https://github.com/nextcloud/docker/pull/527#issuecomment-469474365), we now also have a solution for Apace-based Nextcloud that does not require environment variables to explicitly override the protocol and host.

Here's the Apache version:
https://github.com/GuyPaddock/inveniem-nextcloud-azure/commit/d09ee0cf0b50fca726cd87147a3fea782aa33ce1#diff-754ac7f0be091966820161830749036fR35

J0WI:
I'll close this in favor of #819

PR #535, opened 2018-11-13
Launch crond in background from entrypoint.sh

rcdailey:
This is so that the crontab is executed on a regular schedule in the background
without a secondary container to execute it.

Relates to #134

### DIFF
--- a/12.0/apache/cron.sh
+++ b/12.0/apache/cron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -l 0 -L /dev/stdout
+exec busybox crond -b -l 0 -L /dev/stdout
--- a/12.0/apache/entrypoint.sh
+++ b/12.0/apache/entrypoint.sh
@@ -129,4 +129,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     fi
 fi
 
+/cron.sh
+
 exec "$@"
--- a/12.0/fpm-alpine/cron.sh
+++ b/12.0/fpm-alpine/cron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -l 0 -L /dev/stdout
+exec busybox crond -b -l 0 -L /dev/stdout
--- a/12.0/fpm-alpine/entrypoint.sh
+++ b/12.0/fpm-alpine/entrypoint.sh
@@ -129,4 +129,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     fi
 fi
 
+/cron.sh
+
 exec "$@"
--- a/12.0/fpm/cron.sh
+++ b/12.0/fpm/cron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -l 0 -L /dev/stdout
+exec busybox crond -b -l 0 -L /dev/stdout
--- a/12.0/fpm/entrypoint.sh
+++ b/12.0/fpm/entrypoint.sh
@@ -129,4 +129,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     fi
 fi
 
+/cron.sh
+
 exec "$@"
--- a/13.0/apache/cron.sh
+++ b/13.0/apache/cron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -l 0 -L /dev/stdout
+exec busybox crond -b -l 0 -L /dev/stdout
--- a/13.0/apache/entrypoint.sh
+++ b/13.0/apache/entrypoint.sh
@@ -129,4 +129,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     fi
 fi
 
+/cron.sh
+
 exec "$@"
--- a/13.0/fpm-alpine/cron.sh
+++ b/13.0/fpm-alpine/cron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -l 0 -L /dev/stdout
+exec busybox crond -b -l 0 -L /dev/stdout
--- a/13.0/fpm-alpine/entrypoint.sh
+++ b/13.0/fpm-alpine/entrypoint.sh
@@ -129,4 +129,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     fi
 fi
 
+/cron.sh
+
 exec "$@"
--- a/13.0/fpm/cron.sh
+++ b/13.0/fpm/cron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -l 0 -L /dev/stdout
+exec busybox crond -b -l 0 -L /dev/stdout
--- a/13.0/fpm/entrypoint.sh
+++ b/13.0/fpm/entrypoint.sh
@@ -129,4 +129,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     fi
 fi
 
+/cron.sh
+
 exec "$@"
--- a/14.0/apache/cron.sh
+++ b/14.0/apache/cron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -l 0 -L /dev/stdout
+exec busybox crond -b -l 0 -L /dev/stdout
--- a/14.0/apache/entrypoint.sh
+++ b/14.0/apache/entrypoint.sh
@@ -129,4 +129,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     fi
 fi
 
+/cron.sh
+
 exec "$@"
--- a/14.0/fpm-alpine/cron.sh
+++ b/14.0/fpm-alpine/cron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -l 0 -L /dev/stdout
+exec busybox crond -b -l 0 -L /dev/stdout
--- a/14.0/fpm-alpine/entrypoint.sh
+++ b/14.0/fpm-alpine/entrypoint.sh
@@ -129,4 +129,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     fi
 fi
 
+/cron.sh
+
 exec "$@"
--- a/14.0/fpm/cron.sh
+++ b/14.0/fpm/cron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -l 0 -L /dev/stdout
+exec busybox crond -b -l 0 -L /dev/stdout
--- a/14.0/fpm/entrypoint.sh
+++ b/14.0/fpm/entrypoint.sh
@@ -129,4 +129,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     fi
 fi
 
+/cron.sh
+
 exec "$@"
--- a/15.0-beta/apache/cron.sh
+++ b/15.0-beta/apache/cron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -l 0 -L /dev/stdout
+exec busybox crond -b -l 0 -L /dev/stdout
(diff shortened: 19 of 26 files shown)


marceljd:
You have to run update.sh before you commit your change, then the checks will pass.

LorbusChris:
IMO this is not the right way to do this.
When running the container on Kubernetes, one could add a CronJob for this.
For normal docker operation, you could start another container to do this.

Please do not force this into everybody's entrypoint.

rcdailey:
I'm not forcing anything, this is a PR. It's up to you if you want to merge it.

What's the purpose of the background `cron.php` that runs? Why have I been hearing that some folks want it, but others don't? What does it actually do every 15 minutes? And why does Nextcloud warn in big red text that the cron jobs aren't running, as if it's a required part of the functionality of the app?

Everything I've seen so far indicates that the cronjob is required as part of a normal running instance of Nextcloud. If this is not true, then I'd like to understand that in detail.

I hear a lot of opinions on how this isn't the right way to do it, but no one can objectively explain why. From my standpoint, it seems like just someone else's personal preference. For a large part of the community, it takes a lot of effort to set up Nextcloud in docker which is unnecessary. If there's something we can do to make that process easier, with less of a learning curve, we should be doing it. It's very frustrating that people are so adamant to not have the cron daemon run this task inherent to the container, when there are only positives to doing so as far as I've learned.

LorbusChris:
@rcdailey I'm sorry if I came across harsh. You are absolutely right, there are a lot of opinions floating around on this and having the reasoning behind the current setup explained properly would indeed be great. Having the cron job run in the same container doesn't bode well with the one-process-per-container mantra, though. For me personally the usage on Kubernetes is most important, and as stated above there are other ways to accomplish this on that platform.

I would like to see topic #134 reopened. Let's get all the facts and find a solution that suits all usecases.

rcdailey:
> Having the cron job run in the same container doesn't bode well with the one-process-per-container mantra, though

I think plenty of people have debunked this argument. It's a guideline, not a hard rule. Again, I think we should be focused on what's practical, rather than principles that don't benefit anyone.

> For me personally the usage on Kubernetes is most important, and as stated above there are other ways to accomplish this on that platform

Can't say I'm familiar with that environment, but I can't imagine why it wouldn't work. Maybe for my benefit you can explain it, as a learning opportunity.

Regardless, though, I agree this needs some formal attention. It would be great to have the developers step in with a clear, detailed, and unambiguous explanation. Leaving this up to the open source community hasn't worked out well so far; there needs to be some formal direction here...

fuse314:
You could make the start of /cron.sh configurable through a new environment variable. (let's call it "NEXTCLOUD_ENABLE_CRON")
This leaves the behavior unchanged for current installations that have already solved this in another way (second "cron" container/supervisord/web cron/kubernetes cron/...) but allows others to enable this feature within the same container.
In my personal opinion, running a second docker container just for the cron functionality is a waste of resources, that's why I chose to use supervisord in my installation (there are several examples floating around).
I would welcome a built-in official solution to this.

rcdailey:
> In my opinion this PR cannot be merged in the current state because it forces the cron process onto everyone and that's not acceptable.

Yet again an opinion is shared but no explanations or facts. No one has answered the questions I have on why this is "unacceptable".

> So for everyone who does not want to use both processes in a single container things might break or have unwanted side effects.

What breaks? What are some unwanted side effects? Where are the examples and facts?

> Using two containers is equally easy

As I said before, this is demonstrably false. It took me days to figure out what I had to do. The setup in docker compose was complex, there's zero documentation officially on this, and a lot of different and incorrect information on Google. Also the up/down behavior of compose is much slower now with so many unnecessary containers.

> keeps the processes nice and clean away from each other.

Again I need to see something factual. I disagree anything about the status quo is clean right now for reasons I've already explained. Why is it important to keep the processes separated? And if your answer is because of that docker "one process per container", that is not acceptable. There are some very highly voted opinions in the community that this is only a guideline and not a hard rule. Plenty of folks agree that cron is a fair exception and plenty of other containers do things this way.

> But on the other hand the frequent questions and discussion show that quite a few users are confused or misguided despite our efforts to make things simple. 

What efforts to make things simple? I think it's fairly obvious that so far no one has made any effort to make things simple. Just look at all the great ideas out there that are constantly denied and refuted. This PR is yet another example. You're all still focused on multiple containers which in no way is easier. There is zero effort I've seen to make this solution more turn key.

And most importantly, I still don't have any technical explanation on why it's bad to have the cron process running in this container. And why the cron work itself is treated as optional as some but not by others. And why the tool warns very critically that the cron jobs aren't running. 

If the cron job is responsible for automatic maintenance of the instance, I consider that a required piece of the software that users shouldn't be directly familiar with since it's a complex implementation detail. It should be in the Nextcloud container for everyone. Again, hoping someone please provides some facts here. This issue desperately needs more facts and less opinions. I think we need developer involvement at this point.

marceljd:
I am curious why the Ajax version is the default in an image that contains a full webserver. Nextcloud is running much better (quicker) when the background jobs are not triggered by a page request. If you provide Nextcloud as a web-application (the original upstream version) I can understand the default choice because you cannot influence the webserver beneath it directly, but in a Docker container I would go for the easiest and best implementation. That includes cron and Redis, and both can be added through environment variables. With both enabled you do not need to create your own image, you do not have problems with file locking and the pages are loaded very quickly. I would build both options into the default image. In the update-script you need to detect the setting, for new installations I would enable them by default and give the option to disable each option. For most users Nextcloud will be running out-of-the-box with the best user experience possible. Isn't that the highest goal?

rcdailey:
I am not familiar with redis, and I see a lot of performance issues and locking problems with Nextcloud that I don't know how to solve. I agree the end goal is something that works out of the box and efficiently and configuration-free as possible. @marceljd could you share your setup?

SnowMB:
> And most importantly, I still don't have any technical explanation on why it's bad to have the cron process running in this container. And why the cron work itself is treated as optional as some but not by others. And why the tool warns very critically that the cron jobs aren't running.
> 
> If the cron job is responsible for automatic maintenance of the instance, I consider that a required piece of the software that users shouldn't be directly familiar with since it's a complex implementation detail. It should be in the Nextcloud container for everyone. Again, hoping someone please provides some facts here. This issue desperately needs more facts and less opinions. I think we need developer involvement at this point.


If you want to know what the cron job is doing have a look in the official documentation:
https://docs.nextcloud.com/server/14/admin_manual/configuration_server/background_jobs_configuration.html?highlight=cron

For nextcloud it is not necessary to run a standalone process that runs the cron.php script. For small instances with just a few users it is sufficient to do these task as an additional payload when a user is connected. That's whats Ajax is for. No cronjob required. That's the __default__ configuration. 

 However certain apps and configuration require the backround jobs to be run periodically and this is where cron (or any other solution that can run a script periodically) comes into play. The documentation gives a default of 15 minutes between each run but this is a value that might be customized by the administrators of the instances. 

> What efforts to make things simple? I think it's fairly obvious that so far no one has made any effort to make things simple. Just look at all the great ideas out there that are constantly denied and refuted. This PR is yet another example. You're all still focused on multiple containers which in no way is easier. There is zero effort I've seen to make this solution more turn key.

To make things easy for the user the docker image already contains the cron executable, a startup script `cron.sh` and the crontab with default values (as you noticed yourself). A user just has to customize the image to spawn a second process (what your change or the supervisord examples do) or spawn a second container with the `cron.sh` script as entrypoint. Which both is documented in the examples section.


> Again I need to see something factual. I disagree anything about the status quo is clean right now for reasons I've already explained. Why is it important to keep the processes separated? And if your answer is because of that docker "one process per container", that is not acceptable. There are some very highly voted opinions in the community that this is only a guideline and not a hard rule. Plenty of folks agree that cron is a fair exception and plenty of other containers do things this way.

If you simply spawn cron as a background process consider what happens if the process crashes. Who handles that? Here monitoring your docker containers would not help, because the container is still running with the apache/php-fpm task as foreground.

> As I said before, this is demonstrably false. It took me days to figure out what I had to do. The setup in docker compose was complex, there's zero documentation officially on this, and a lot of different and incorrect information on Google. Also the up/down behavior of compose is much slower now with so many unnecessary containers.

I am very sorry that using he image has caused you troubles. But please keep in mind that this is still a community maintained project where nobody gets paid to work on. I am glad that you are getting involved in the discussion and try to make things better for other users :+1: 


> I am curious why the Ajax version is the default in an image that contains a full webserver. Nextcloud is running much better (quicker) when the background jobs are not triggered by a page request. If you provide Nextcloud as a web-application (the original upstream version) I can understand the default choice because you cannot influence the webserver beneath it directly, but in a Docker container I would go for the easiest and best implementation. That includes cron and Redis, and both can be added through environment variables. With both enabled you do not need to create your own image, you do not have problems with file locking and the pages are loaded very quickly. I would build both options into the default image. In the update-script you need to detect the setting, for new installations I would enable them by default and give the option to disable each option. For most users Nextcloud will be running out-of-the-box with the best user experience possible. Isn't that the highest goal?

We try to find the balance between easy of use and still be open enough for customizations. There are others actively maintaining more "complete" images but for the official image I don't think we should include to much. For example users with an external redis server might not want the overhead of a redis installation inside of the container. We had some discussions in the past about what to include in the image (smb and imap support for example) and came to the conclusion that a) we either get to many image variants, that this gets unmaintainable or b) we force overhead on the users. We kind of settled with c) trying to find a senseful base and adding more documentation on how to build derived images on top of the official nextcloud image and more examples. 

I am sure that there is room to improve both on the image and on the documentation side. But please keep in mind, that all the work is done by volunteers that don't get paid. The docker image is not developed by the nextcloud employees.

marceljd:
I think the main goal should be an easy to implement Nextcloud image that causes as little problems as possible. That way everyone can start and use Nextcloud without too much technical stuff. That is what Docker is all about. I had a few encounters with file locking, and it was hard to solve these. It would be a reason to stop if I did not have any tech knowledge. And I was annoyed by the slow page loading, especially when I did not use the webinterface for some time. The Ajax background jobs are not triggered for a long time, that causes a longer wait for my page and I do not think it is helping the NC instance. Background jobs that should run every 15 minutes to keep everything ok should not run every few days or even weeks.

marceljd:
> I am sure that there is room to improve both on the image and on the documentation side. But please keep in mind, that all the work is done by volunteers that don't get paid. The docker image is not developed by the nextcloud employees.

That is clear, and I do understand that this is not something that is ready tomorrow. If I can I would be glad to help, but only after the idea get the commitment it needs. Otherwise I would invest time that will lead to nothing...

rcdailey:
@SnowMB Thank you for your feedback. And I apologize for letting my frustration show. I've struggled to get Nextcloud running efficiently, but I'm seeing a lot of issues and this PR represents just one small piece of that struggle. I thought I was doing something that most people wanted, but I can see based on your response it doesn't align with the direction the developers are setting. I think that's fine.

I think something else that might be acceptable is if there's a drop-in docker compose script + dependent files that people can just copy & run. Right now, I have my own dedicated nginx reverse proxy for other services, and I found it a struggle to make it tie nicely into the `nextcloud:fpm` image, since that means nginx needs to map to the data volume of Nextcloud itself. So there's a difficult to manage tight coupling between my reverse proxy container (which isn't in my nextcloud docker compose file) and the nextcloud container itself.

Does such a shared docker compose exist for homelab instances with less than 5 users? I think what's important is a reliable configuration that performs well. I see too many behavioral inefficiencies today, and I effectively had to map a php.ini, www-data config, and other things to "optimize" my instance, but now it just feels like it's hogging machine resources by spawning tons of worker processes.

I'm still struggling here. If we decline this PR, that's OK, but I would like to finally address these issues.

SnowMB:
Like I said, I certainly see room for improvement on that issue. For me the solution @fuse314 proposed seems a good compromise and the documentation should be improved.

For your own instance, I can see that the setup you are aiming for is not the easiest. I for myself (I am maintaining 2 instances, one with 4 users and one with about 10) switched back from nginx+php-fpm to using the apache image with the supervisord approach for cron. On the 10 user instance I use postresql and an additional redis container for caching, on the smaller just mariadb. On neither instance I have any performance issues.

Please note that while nginx is widely used as a web server for nextcloud and there are many  examples, apache is still the main target for the nextcloud devs and all of the default configuration is targeted for apache. So naturally you have to expect more work when going the nginx route.

If I was in your position and wanted to use nginx and php-fpm I probably would let the reverse proxy just do the proxying and handling of the certificates and add a nginx behind that one, that acts as the web server for nextcloud and shares the data volume. That adds an additional container to run, but it decreases configurational complexity through clear separation of concerns. You could then follow the examples closely.

rcdailey:
Thanks. I'm find with this setup:

    Nginx -> Apache -> FPM

Assuming the Apache & FPM part are the `nextcloud:apache` container, and Nginx is a reverse proxy I already have for many other services. The problem is, most examples of nginx are for the FPM directly, as in this setup:

    Nginx -> FPM

But this requires volume sharing between those 2 containers and is more complex to set up. The benefit, though, is that there are working examples of this out there (even though they are complex).

Is there any inefficiency in using the apache + fpm container and using a simple reverse proxy for nginx? This way nginx doesn't know anything about PHP, and it's just pure HTTP. I'd have to figure out how to configure nginx for this scenario though, as I haven't seen any examples.

Also I see there is a [example docker compose](https://github.com/nextcloud/docker/tree/master/.examples/docker-compose/insecure/mariadb-cron-redis/apache) for apache + redis + mariadb (I'm using mysql container so hopefully they are interchangeable). Should I start there?

It's unclear to me how to set up redis, but I can look into it and learn a bit.

marceljd:
I am using the Apache image only (not FPM) and am perfectly happy with it. Traefik as proxy handling certificates (not only for NC). Database is MySQL and Redis for caching (and preventing file locking). There are 3 containers: NC (Apache), database (MySQL) and Redis. Volumes exist for Apache and MySQL.

I made a PR to setup Redis through an environment variable, it is not that difficult. You have to start a Redis container (see the docker-compose file you found, use the standard port 6379) and modify the config.php with a few lines. For new installations it could be set through the env vars, to change an existing installation you have to change the config.php manually, you can use these lines and copy them into your config.php:

```
'memcache.locking' => '\\OC\\Memcache\\Redis',
    'redis' => array(
      'host' => 'redis', _(insert your Redis hostname here if you chose something else in your compose file)_
      'port' => 6379,
   ),
```

PR #538, opened 2018-11-15
Use Alpine Linux in examples and use default user on nginx containers

J0WI:


J0WI:
We could also drop all the custom Dockerfiles and just mount the configs, but I'm not sure how this behaves with relative paths.

SnowMB:
The custom Dockerfiles were introduced because of problems with docker for windows and the virtual machine running docker. I don't know what's the current status on this.

J0WI:
I could revert the `fpm-alpine` change, but I think the examples are for new users to play with the image. People should not relay on the examples for production usage without making adjustments for their personal use case (I think they are forced to set at least a database password). Otherwise we could never modernize the examples.

gwarf:
I've been using (and probably not the only one) the `.examples/docker-compose/with-nginx-proxy/mariadb-cron-redis/fpm` examples very happily to run my nextcloud instance, I'm used to `git stash`, `git pull` and `git stash pop` to keep/update/re-apply my small required changes (ie. only `MYSL_ROOT_PASSWORD`, `VIRTUAL_HOST`, `LETSENCRYPT_HOST` and `LETSENCRYPT_EMAIL`). Is it possible to get some advice on how to handle the changes that this PR introduced?

J0WI:
@gwarf in your case, you need to chown `gid` and `uid` to `82` in your `nextcloud` volume.

PR #542, opened 2018-11-21
 Update APCu to version 5.1.14

tilosp:
https://pecl.php.net/package-changelog.php?package=APCu&release=5.1.14


J0WI:
Should we automate this in #543?

PR #543, opened 2018-11-21
Update pecl extensions in update.sh

J0WI:


update.sh
```diff
@@ -24,10 +24,37 @@ declare -A extras=(
 	[fpm-alpine]=''
 )
 
+apcu_version="$(
+	git ls-remote --tags https://github.com/krakjoe/apcu.git \
```
tilosp: Are you sure that the version is available on pecl as soon as it's tagged on github?

tilosp:
I generally like the idea of automating this.

But this would cause a lot of confusing commits.
For example if the redis version is updated, we would get mutiple commits from the jenkins server that runs update.sh.
And they would be called `Update to 14.0.3`, `Update to 13.0.7` and `Update to 12.0.12`.

The code that generates the commits is here https://github.com/docker-library/oi-janky-groovy/blob/master/update.sh/target-pipeline.groovy

J0WI:
Yes, that would be the tradeoff. It's hard to change `target-pipeline.groovy` in a way that does not affects other repositories and I don't know if @tianon would accept such a specific change request.
See also https://github.com/docker-library/openjdk/issues/246

tilosp:
We could also use a different solution for running the update.sh script.
Especially because the current jenkins script does not handle deleted file and updates to travis.yml. See: https://github.com/docker-library/oi-janky-groovy/pull/8

Maybe something similar to what i am doing for `generate-stackbrew-library.sh`.

tilosp:
Btw there were also some wired commits in this repo https://github.com/nextcloud/docker/compare/ab3b62bd1024eeb819533b97eeb0b51c14449e5c...f16c8572317d9f04d2bd6e23db8341c9abf9a1eb

update.sh
J0WI: Nope, but it's hard to fetch the version from PECL.

tilosp: have you tried using the rss feed? https://pecl.php.net/feeds/pkg_APCu.rss

tilosp:
@J0WI update.sh is now done by @tilosp-bot (see for example https://github.com/nextcloud/docker/pull/662) so this should be possible now

update.sh
J0WI: I tried to use xmllint, but it's still harder than using git tags.

tilosp: I think git tags a fine for now

if a lot of builds end up failing, we can still change it to something hacky like this:
```bash
curl -fsSL https://pecl.php.net/feeds/pkg_APCu.rss |tac|tac| grep -oE '<title>APCu [[:digit:]]+(\.[[:digit:]]+){2}</title>' | grep -oE '[[:digit:]]+(\.[[:digit:]]+){2}'
```

update.sh
```diff
@@ -25,11 +25,47 @@ declare -A extras=(
...
+imagick_version="$(
+	git ls-remote --tags https://github.com/mkoppanen/imagick.git \
+		| cut -d/ -f3 \
+		| grep -viE '[a-z]' \
+		| tr -d '^{}' \
+		| sort -V \
+		| tail -1
+)"
+
 declare -A pecl_versions=(
```
tilosp: is there any reason to keep this array? we can use the vars directly

J0WI: Maybe we have to use multiple versions for different versions of PHP/NextCloud, like we did for PHP 5.

tilosp: maybe but how would you do this with the current data structure?
for php 5 there where two templates, see: https://github.com/nextcloud/docker/tree/10

J0WI: The current structure does indeed not use `php_version`, but you can easily expand it if necessary.

PR #555, opened 2018-11-27
Match image version with corresponding Server Administration Manual

mmarzantowicz:
Currently all `Dockerfile`s refer to Server Administration Manual version 12. This patch allows to base this version on specific image version.


J0WI:
I'd prefer using `stable` or `latest`.
Please run `update.sh` after modifying the templates.

Issue #345, opened 2018-05-25
Unable to connect to Postgres with docker-compose


cclloyd:
When trying to set up nextcloud it keeps telling me that it can't connect to the database, or that password authentication failed for user 'postgres'.  (I can connect via psql so the username/password is correct)

    version: '3.5'
  
    services:
        db:
            image: postgres
            restart: always
            ports:
              - "5432:5432"
            volumes:
              - ./db/nextcloud.sql:/docker-entrypoint-initdb.d/nextcloud.sql
              - postgres-data:/var/lib/postgresql/data
        nextcloud:
            image: nextcloud
            ports:
              - 8080:80
            depends_on:
              - db
            volumes:
              - nextcloud:/var/www/html
            restart: always
            environment:
              POSTGRES_DB: 'nextcloud'
              POSTGRES_USER: 'postgres'
              POSTGRES_PASSWORD: 'postgres'
              POSTGRES_HOST: 'db:5432'
    
    volumes:
      nextcloud:
      postgres-data:
      db:

J0WI:
Your containers are not linked together or connected to a bridge. So `db` is an unknown host.

beenje:
`POSTGRES_DB: 'nextcloud'` is incorrect. It should be `POSTGRES_DB: 'db'`.
"db" is the name of your postgres service.

SnowMB:
Just ran into the same issue, thanks for the explanation :grin:

SnowMB:
This issue is apparently present in our examples section and comes up a couple of times. I will join the bug requests in this issue.

lqhuang:
@SnowMB Hey, Marc. I just saw your commits to fix this bug with set `POSTGRES_DATABASE=not_used` few hours ago. But I had tried new configuration, and the same problem is still there. I'm not very familiar with docker repo postgres. Hence, I want to know what do you think or what do you want to do to let `db` be initialized by `nextcloud-app` instead of be initialized by itself? Thanks for your efforts.

victornoel:
@SnowMB for the record, with the same docker-compose file, I have machines where I have this problem and others where I don't.

I have no idea what is the difference between those (they use the same version of docker, and those failing are using docker-compose 1.21.1 while the working one is using 1.22.0 but I don't think this is the source of the problem...).

This is with `postgres:10-alpine` and `nextcloud:fpm-alpine`.

The docker-compose is:
```
version: '3'

networks:
  proxytanet:
    external: true

services:
  db:
[...]

Issue #104, opened 2017-06-17
[example docker-compose] Mounting nginx.conf fails with remote docker


ptitjes:
I try to use your example docker-compose. I just modified the domain, passwords and email to all use variables. I did put the `nginx.conf` file next to the `docker-compose.yml` file (and checked the permissions of those file).

But when I do
```
DOCKER_HOST=tcp://...:2376 DOCKER_TLS_VERIFY=1 \
DOMAIN="..." PASSWORD="..." EMAIL="..." docker-compose up
```
every container launches correctly except the web container:
```
ERROR: for web  Cannot start service web: oci runtime error: container_linux.go:247: starting container process caused "process_linux.go:359: container init caused \"rootfs_linux.go:54: mounting \\\"/home/didier/Code/docker/nextcloud/nginx.conf\\\" to rootfs \\\"/var/lib/docker/overlay2/770c4e28e5f59e3b7e1ff3fcbacecf011bffcbffda9853a95b412557f6c026f4/merged\\\" at \\\"/var/lib/docker/overlay2/770c4e28e5f59e3b7e1ff3fcbacecf011bffcbffda9853a95b412557f6c026f4/merged/etc/nginx/nginx.conf\\\" caused \\\"not a directory\\\"\""
: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
ERROR: Encountered errors while bringing up the project.
```

Do you have any idea on how I can fix this ? I found this error a lot on the web but no working fix.

SnowMB:
The error indicates that docker is trying to mount a single file (your `nginx.conf`) as a directory inside your container. Double check the docker command you used or the docker-compose file. (look for trailing slash or backslash).

ptitjes:
@SnowMB OK, I figured it out. docker-compose cannot bind-mount local files to a remote docker.

I fix it by doing a nginx build with Dockerfile:
```
FROM nextcloud:fpm
COPY config.php /var/www/html/config/config.php
```

I'm now fighting docker because it assigns user root:root to named volume and thus nextcloud can't read in /var/www/html/config...

SnowMB:
Yeah, we are aware of that poblem. For now I would recommend using only one volume for nextcloud: The whole `/var/www/html` folder. 

If you want to use multiple volumes you can change the permissions by hand. See #75.

ptitjes:
[...]

PR #559, opened 2018-11-29
map configs via volumes instead of build new image

avaika:
Wasn't sure if this will be accepted, thus did the change only for proxy-mariadb and proxy-postgres images. Can do the rest as well, but want to be sure that it won't be just waste of time.

PS. Also added a workaround for https://github.com/nextcloud/docker/issues/345


J0WI:
This has already been discussed here: https://github.com/nextcloud/docker/pull/538#issuecomment-439074124

@SnowMB could you give more details about this issue and how we can test this?

J0WI:
:ping_pong: @SnowMB

SnowMB:
Would love to help in this regard, but I can't. I have no windows machine to test this out. But I can point you to the corresponding issues / commits:

Initial problems came from #104. And I discussed this with @tilosp in https://github.com/nextcloud/docker/pull/125#issuecomment-32217644

I don't know if this is still relevant as I see many other images mounting config files, too.

J0WI:
@tilosp are you fine with dropping the homebrew proxy containers?

J0WI:
@avaika could you rebase your PR?

tilosp:
This will not work if the docker client is on a different machine than the docker host.
Because the docker host needs to mount the config files but they are only available on the client machine.
This is for example the case on docker for windows (in Linux container mode). There the client runs under windows and the docker host is a linux vm.

I don't see any other benefits to this pr other than not having to run `docker-compose build`.
Am i missing something?

So we have to decide whether we want to support remote docker hosts or not.

PR #525, opened 2018-10-26
Set Redis via environment vars

marceljd:
Copy of the Redis part of #292 without the SMTP part. Redis has to be set at boot time, SMTP can be set through the Nextcloud webinterface. Tested with Apache / 14.0.3.

New PR after running update.sh

@rodrigoaguilera @unteem

PR #573, opened 2018-12-10
Use PHP 7.3 for NC 15

J0WI:


J0WI:
Blocked by https://github.com/php-memcached-dev/php-memcached/issues/408

update.sh
```diff
@@ -3,7 +3,7 @@ set -eo pipefail
 
 declare -A php_version=(
 	[default]='7.2'
-	[12.0]='7.1'
+	[15.0]='7.3'
 )
```
tilosp: I think you should make 7.3 the default and set 13 and 14 to 7.2.
Because if nextcloud 16 is ready we would want it to use php 7.3 and not 7.2

J0WI:
We could just drop memcached from the image in favor of https://github.com/nextcloud/docker/pull/525#issuecomment-446599418

tilosp:
> We could just drop memcached from the image in favor of [#525 (comment)](https://github.com/nextcloud/docker/pull/525#issuecomment-446599418)

Yes dropping it should be fine. I don't think it is used that much, but we have no way of knowing for sure.

But we might want to wait a bit more before moving to php 7.3, because there are some apps that do not support it right now. (For example https://apps.nextcloud.com/apps/audioplayer/releases?platform=15#15)

J0WI:
> Yes dropping it should be fine. I don't think it is used that much, but we have no way of knowing for sure.

Oh, there is one: https://github.com/nextcloud/survey_server/issues/21

J0WI:
> Oh, there is one: nextcloud/survey_server#21

@SnowMB could you help out here?

J0WI:
`memcached-3.1.2` is PHP 7.3 compatible and https://github.com/Rello/audioplayer/commit/ffba9fb1ef1cb2a2f8efaab888eb6dca1cc7df3a adds compatibility to the audio player app.

J0WI:
Previous build failed because https://github.com/eduardok/libsmbclient-php/releases/tag/1.0.0 was not yet released.

J0WI:
Rebased

J0WI:
@tilosp do you think it's safe to ship this now? Otherwise I'll cherry-pick the pecl updates from here.

bestlibre:
the mail app is not yet compatible with 7.3.

PR #575, opened 2018-12-11
Add imagick php extension

### LINKED ISSUE #263: Add ImageMagick PHP extension to support "Theming of icons"
The current docker image doesn't provide the dependencies for "Theming of icons":

https://docs.nextcloud.com/server/13/admin_manual/configuration_server/theming.html#theming-of-icons

Please add the ImageMagick PHP extension and ImageMagick SVG support to the container to support custom favicons in settings/admin/theming of Nextcloud.

### LINKED ISSUE #574: Nextcloud 15 recommends 'imagick' to be installed
I get this message using Nextcloud 15 apache container:

> This instance is missing some recommended PHP modules. For improved performance and better compatibility it is highly recommended to install them.
>
> * imagick

Can this be added to the image?

rcdailey:
There is a warning issued in Nextcloud 15 about 'imagick' php extension not
being present. They highly recommend that it be installed.

Fixes #574 Fixes #263

### DIFF
--- a/.examples/dockerfiles/full/fpm/Dockerfile
+++ b/.examples/dockerfiles/full/fpm/Dockerfile
@@ -4,7 +4,7 @@ RUN mkdir -p /usr/share/man/man1 \
     && apt-get update && apt-get install -y \
         supervisor \
         ffmpeg \
-        libmagickwand-dev \
+        libbz2-dev \
         libgmp3-dev \
         libc-client-dev \
         libkrb5-dev \
@@ -15,8 +15,8 @@ RUN mkdir -p /usr/share/man/man1 \
     && docker-php-ext-configure imap --with-kerberos --with-imap-ssl \
     && ln -s "/usr/include/$(dpkg-architecture --query DEB_BUILD_MULTIARCH)/gmp.h" /usr/include/gmp.h \
     && docker-php-ext-install bz2 gmp imap \
-    && pecl install imagick smbclient \
-    && docker-php-ext-enable imagick smbclient \
+    && pecl install smbclient \
+    && docker-php-ext-enable smbclient \
     && mkdir /var/log/supervisord /var/run/supervisord
 
 COPY supervisord.conf /etc/supervisor/supervisord.conf
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -32,6 +32,7 @@ RUN set -ex; \
         libpng-dev \
         libpq-dev \
         libxml2-dev \
+        libmagickwand-dev \
     ; \
     \
     debMultiarch="$(dpkg-architecture --query DEB_BUILD_MULTIARCH)"; \
@@ -53,11 +54,13 @@ RUN set -ex; \
     pecl install APCu-%%APCU_VERSION%%; \
     pecl install memcached-%%MEMCACHED_VERSION%%; \
     pecl install redis-%%REDIS_VERSION%%; \
+    pecl install imagick-%%IMAGICK_VERSION%%; \
     \
     docker-php-ext-enable \
         apcu \
         memcached \
         redis \
+        imagick \
     ; \
     \
 # reset apt-mark's "manual" list so that "purge --auto-remove" will remove all build dependencies
--- a/.examples/README.md
+++ b/.examples/README.md
@@ -32,11 +32,6 @@ The required steps for each optional/recommended package that is not already in 
 #### PHP Module bz2
 `docker-php-ext-install bz2`  
 
-#### PHP Module imagick
-`apt install libmagickwand-dev`  
-`pecl install imagick`  
-`docker-php-ext-enable imagick`  
-
 #### PHP Module imap
 `apt install libc-client-dev libkrb5-dev`  
 `docker-php-ext-configure imap --with-kerberos --with-imap-ssl`  
--- a/.examples/dockerfiles/full/apache/Dockerfile
+++ b/.examples/dockerfiles/full/apache/Dockerfile
@@ -4,7 +4,7 @@ RUN mkdir -p /usr/share/man/man1 \
     && apt-get update && apt-get install -y \
         supervisor \
         ffmpeg \
-        libmagickwand-dev \
+        libbz2-dev \
         libgmp3-dev \
         libc-client-dev \
         libkrb5-dev \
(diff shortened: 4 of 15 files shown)


Dockerfile-debian.template
```diff
@@ -32,6 +32,8 @@ RUN set -ex; \
         libpng-dev \
         libpq-dev \
         libxml2-dev \
+        libmagickwand-dev \
+        libmagickcore-dev \
```
J0WI: `libmagickwand-dev` already depends on `libmagickcore-dev`, so this is not needed.

J0WI:
@tilosp are you fine with adding this to NC 13/14 too?

Dockerfile-debian.template
J0WI: No, Alpine does not split the package. This does only affect Debian.

Dockerfile-debian.template
```diff
@@ -50,14 +52,16 @@ RUN set -ex; \
     ; \
     \
 # pecl will claim success even if one install fails, so we need to perform each install separately
-    pecl install APCu-%%APCU_VERSION%%; \
-    pecl install memcached-%%MEMCACHED_VERSION%%; \
-    pecl install redis-%%REDIS_VERSION%%; \
+    pecl install APCu-5.1.14; \
+    pecl install memcached-3.0.4; \
+    pecl install redis-4.2.0; \
+    pecl install imagick; \
```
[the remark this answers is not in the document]
rcdailey: I'm not familiar with the update logic here, but this lacks the variable substitution that the Debian dockerfile has. Are these actually updated by the shell script? Or are these just hard-coded?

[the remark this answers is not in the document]
J0WI: See here:
https://github.com/nextcloud/docker/blob/master/update.sh#L27
https://github.com/nextcloud/docker/blob/master/update.sh#L80

rcdailey:
Changes:

* Commit is now signed-off, as well as GPG signed
* Using explicit version (3.4.3) for `imagick` (referred to [this](https://pecl.php.net/package/imagick) page for version numbering).

Note: I added the `%%IMAGICK_VERSION%%` variable to `Dockerfile-debian.template`, but the others were not using it. I figured you wanted the consistency instead of having to update version numbers in 2 files. I did not refactor the existing other 3 packages in that template file. I can do that if you'd like, but trying to keep my changes minimal. Let me know.

Dockerfile-debian.template
```diff
@@ -50,14 +51,16 @@ RUN set -ex; \
     ; \
     \
 # pecl will claim success even if one install fails, so we need to perform each install separately
-    pecl install APCu-%%APCU_VERSION%%; \
-    pecl install memcached-%%MEMCACHED_VERSION%%; \
-    pecl install redis-%%REDIS_VERSION%%; \
+    pecl install APCu-5.1.14; \
+    pecl install memcached-3.0.4; \
+    pecl install redis-4.2.0; \
+    pecl install imagick-%%IMAGICK_VERSION%%; \
```
J0WI: Something went wrong here...

rcdailey: What went wrong? When I ran `update.sh`, I noticed it is substituting this variable properly.

rcdailey: Never mind I see what you mean. Let me fix that.

J0WI: You replaced the variable with the hardcoded version in the Debian template. The Alpine template is fine.

tilosp:
> @tilosp are you fine with adding this to NC 13/14 too?

Yes adding it to nc 13 and 14 is fine.

tilosp:
We should also remove imagick from the full example

rcdailey:
> We should also remove imagick from the full example

Is that something I need to do? I apologize that I'm not as familiar with the structure of things. Let me know what I need to do and I'll see if I can do it. Thanks in advance.

tilosp:
@rcdailey it's not something you have to do. I can do it in a other pr.

But if you are interested there are two example Dockerfiles in this repo (https://github.com/nextcloud/docker/blob/master/.examples/dockerfiles/full/fpm/Dockerfile and https://github.com/nextcloud/docker/blob/master/.examples/dockerfiles/full/apache/Dockerfile)
They add a few additional php extensions to the base image, including imagick.
But if  imagick is already part of the base image we can remove it there.

rcdailey:
@tilosp Updated the examples for you. Also removed the relevant step in the README. Let me know if this is good enough.

rcdailey:
@tilosp I'm seeing some failures in the travis build. Should I be concerned about those? For example, [this one](https://travis-ci.org/nextcloud/docker/jobs/467071702).

tilosp:
@rcdailey the build of the example failed because libbz2-dev is missing.
is was there before because it is a dependency of libmagickwand-dev.

rcdailey:
So in other words, libbz2 is still needed for some other reason, i.e. a direct dependency on it? Which means the example still needs to include that package? I'll make the correction. The tests take some time to run so I'll keep an eye out for other issues.

rcdailey:
Looks like nextcloud-13:fpm is the one failing. Does Nextcloud need bzip in general?

tilosp:
no it should simply be the example image. the only builds that work are the fpm-alpine ones because the have no example dockerfiles (only apache and fpm have examples).

rcdailey:
Thanks for explaining. I am pushing fixup commits until I see all builds pass. Please allow me to squash them after tests pass, prior to merge.

.examples/dockerfiles/full/fpm/Dockerfile
```diff
@@ -4,19 +4,18 @@ RUN mkdir -p /usr/share/man/man1 \
     && apt-get update && apt-get install -y \
         supervisor \
         ffmpeg \
-        libmagickwand-dev \
         libgmp3-dev \
         libc-client-dev \
         libkrb5-dev \
         smbclient \
         libsmbclient-dev \
-#       libreoffice \
```
tilosp: please keep this line. it's referenced from the readme:
```md
NOTE: The Dockerfile does not install the LibreOffice package (line is commented), because it would increase the generated Image size by approximately 500 MB. In order to install it, simply uncomment the 13th line of the Dockerfile.
```

rcdailey: Good catch, I figured someone just forgot to delete the line. I've added it back. Line number was kept the same, as well. It looked like maybe the comment would break the escaped line return, but I guess blank lines in `RUN` commands don't get passed to the shell? I recall putting comments in the middle of escaped lines breaks the continuity of the shell command when working with normal bash scripts.

rcdailey:
@J0WI @tilosp I'm done and happy with all the changes at this point. If both of you have nothing else, I believe we can merge. Let me know if I'm needed for anything else. Thanks to both of you for the feedback & quick turnaround on this PR.

J0WI:
LGTM, thanks!

PR #586, opened 2018-12-19
Fix woff2 files in nginx examples

nicokaiser:
Since some versions of Nextcloud, fonts are served as woff2, which is not forwarded to index.php by the supplied nginx.conf. This leads to nginx returning the dynamic index page instead of the static fonts.

See https://docs.nextcloud.com/server/15/admin_manual/installation/nginx.html


J0WI:
LGTM, please signed off your commit.

J0WI:
:ping_pong: @nicokaiser

J0WI:
Thanks!

Issue #489, opened 2018-10-06
fpm-alpine fails with `Command "maintenance:install" is not defined.` after upgrade


TLATER:
Hi, I just upgraded from 13.0 to 14.0.1 with the fpm-alpine image. When starting with the new image I end up with the following repeating message in my logs:

```
<snip>
retrying install...
Nextcloud or one of the apps require upgrade - only >
You may use your browser or the occ upgrade command >

  Command "maintenance:install" is not defined.

Did you mean one of these?
      app:install
      maintenance:data-fingerprint
      maintenance:mimetype:update-db
      maintenance:mimetype:update-js
      maintenance:mode
      maintenance:repair
      maintenance:theme:update
      maintenance:update:htaccess

installing of nextcloud failed!
```

Any ideas? I didn't catch on to `stable-fpm-alpine` in time, I suppose.

MiGoller:
I have encountered the same error. Nextcloud will start but the Docker's entrypoint tries to install a few times; later on Nextcloud will be up and running.

If I do not provide environment variables for database setup, as I've done all the time up to 13.0.4, Nextcloud starts without any error messages.

Can anybody confirm that? It looks like the Nextcloud Docker container does not recognize that Nextcloud is already installed.

weikinhuang:
I am also seeing this upgrading from 14.0.0 to 14.0.3

SnowMB:
I can't really reproduce the issue. If you start the container, does it give you an installed version number?
Should start with something like this:

```
docker-compose logs -tf app
app_1  | 2018-11-07T18:36:13.906629237Z Initializing nextcloud 14.0.3.0 ...
app_1  | 2018-11-07T18:36:13.906661268Z Upgrading nextcloud from 13.0.7.2 ...
```

weikinhuang:
It thinks it's a new instance:

```
Initializing nextcloud 14.0.3.0 ...
Initializing finished
New nextcloud instance
Installing with MySQL database
starting nexcloud installation
Nextcloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade


  Command "maintenance:install" is not defined.

  Did you mean one of these?
      app:install
      maintenance:data-fingerprint
      maintenance:mimetype:update-db
      maintenance:mimetype:update-js
      maintenance:mode
      maintenance:repair
      maintenance:theme:update
      maintenance:update:htaccess
```

[...]

PR #588, opened 2018-12-20
fix version check without volume var/www/html

unteem:
it fixes https://github.com/nextcloud/docker/issues/489
At the moment the upgrade does not work if there is no volume for `/var/www/html/` as it checks `/var/www/html/version.php` to compare versions. This PR make it so it checks for the version in `var/www/html/config` which must be persisted and everyone is happy :)

### DIFF
--- a/13.0/apache/entrypoint.sh
+++ b/13.0/apache/entrypoint.sh
@@ -6,6 +6,11 @@ version_greater() {
     [ "$(printf '%s\n' "$@" | sort -t '.' -n -k1,1 -k2,2 -k3,3 -k4,4 | head -n 1)" != "$1" ]
 }
 
+# version_equal A B returns whether A = B
+version_equal() {
+    [ "$(printf '%s\n' "$@" | sort -t '.' -n -k1,1 -k2,2 -k3,3 -k4,4 | head -n 1)" = "$1" ]
+}
+
 # return true if specified directory is empty
 directory_empty() {
     [ -z "$(ls -A "$1/")" ]
@@ -21,9 +26,9 @@ run_as() {
 
 if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UPDATE:-0}" -eq 1 ]; then
     installed_version="0.0.0.0"
-    if [ -f /var/www/html/version.php ]; then
+    if [ -f /var/www/html/config/config.php ]; then
         # shellcheck disable=SC2016
-        installed_version="$(php -r 'require "/var/www/html/version.php"; echo implode(".", $OC_Version);')"
+        installed_version="$(awk '$1 ~ /version/ { print $3 }' /var/www/html/config/config.php | grep -oP "(?<=').*(?=')")"
     fi
     # shellcheck disable=SC2016
     image_version="$(php -r 'require "/usr/src/nextcloud/version.php"; echo implode(".", $OC_Version);')"
@@ -31,6 +36,22 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     if version_greater "$installed_version" "$image_version"; then
         echo "Can't start Nextcloud because the version of the data ($installed_version) is higher than the docker image version ($image_version) and downgrading is not supported. Are you sure you have pulled the newest image version?"
         exit 1
+    fi
+
+    if version_equal "$installed_version" "$image_version"; then
+        if [ "$(id -u)" = 0 ]; then
+            rsync_options="-rlDog --chown www-data:root"
+        else
+            rsync_options="-rlD"
+        fi
+        rsync $rsync_options --delete --exclude-from=/upgrade.exclude /usr/src/nextcloud/ /var/www/html/
+
+        for dir in config data custom_apps themes; do
+            if [ ! -d "/var/www/html/$dir" ] || directory_empty "/var/www/html/$dir"; then
+                rsync $rsync_options --include "/$dir/" --exclude '/*' /usr/src/nextcloud/ /var/www/html/
+            fi
+        done
+        echo "Initializing finished"
     fi
 
     if version_greater "$image_version" "$installed_version"; then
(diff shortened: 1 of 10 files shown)


J0WI:
This duplicates a lot of code from `version_greater`. Maybe a "greater or equal" function could cover both cases?

unteem:
@J0WI I agree that it looks like there are some duplicated code but the procedure is not the same if the version is equal or greater. 

If version is equal I just rsync the php code and do nothing whereas if the version is greater I rsync the php code and upgrade. 

What would you recommend to refactor it?

J0WI:
You can move the duplicated part to a new function that can be called in both cases.

13.0/apache/entrypoint.sh
```diff
@@ -6,6 +6,11 @@ version_greater() {
     [ "$(printf '%s\n' "$@" | sort -t '.' -n -k1,1 -k2,2 -k3,3 -k4,4 | head -n 1)" != "$1" ]
 }
 
+# version_equal A B returns whether A = B
+version_equal() {
+    [ "$(printf '%s\n' "$@" | sort -t '.' -n -k1,1 -k2,2 -k3,3 -k4,4 | head -n 1)" = "$1" ]
```
tilosp: This function is simple the inverse of the function above.
So either the name is wrong or the implementation is wrong.

weikinhuang:
Has there been any update on this, it is sorely needed when running in a kubernetes environment where I would like to avoid mounting the php codebase on a remote share.

J0WI:
If @unteem is not responding you're welcome to continue here.

timLoewel:
Does this really work?
If I I start the apache docker container with a `NEXTCLOUD_CONFIG_DIR` set, how does this check for the version then?
As far as I can tell new script only checks for the `config.php` in the default config directory?

And my second problem ist: even if I change the script to check in the `NEXTCLOUD_CONFIG_DIR`, there is no version number set in my config.php

it looks like this:
```
<?php
$CONFIG = array (
  'passwordsalt' => 'xxxxx',
  'secret' => 'xxxx',
  'trusted_domains' => 
  array (
    0 => 'localhost',
  ),
  'datadirectory' => '/mnt/nextcloudFiles/data',
  'dbtype' => 'mysql',
  'version' => '16.0.1.1',
  'overwrite.cli.url' => 'http://localhost',
  'dbname' => 'nextcloud',
  'dbhost' => '127.0.0.1',
  'dbport' => '',
  'dbtableprefix' => '',
  'dbuser' => 'nextclouduser',
  'dbpassword' => 'xxxxx',
  'installed' => true,
  'instanceid' => 'ocyldeyn9lq3',
); 
```

J0WI:
> there is no version number set in my config.php

`  'version' => '16.0.1.1',` :wink:

timLoewel:
:flushed: how blind I am. sorry..

but I stand by my first point. A custom config dir does not seem to be handled.

13.0/apache/entrypoint.sh
```diff
@@ -21,9 +26,9 @@ run_as() {
 
 if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UPDATE:-0}" -eq 1 ]; then
     installed_version="0.0.0.0"
-    if [ -f /var/www/html/version.php ]; then
+    if [ -f /var/www/html/config/config.php ]; then
```
timLoewel:
```suggestion
    if [ -f $NEXTCLOUD_CONFIG_DIR/config.php ]; then
```

13.0/apache/entrypoint.sh
```diff
@@ -21,9 +26,9 @@ run_as() {
...
         # shellcheck disable=SC2016
-        installed_version="$(php -r 'require "/var/www/html/version.php"; echo implode(".", $OC_Version);')"
+        installed_version="$(awk '$1 ~ /version/ { print $3 }' /var/www/html/config/config.php | grep -oP "(?<=').*(?=')")"
```
timLoewel:
```suggestion
            installed_version=$(CONFIG_DIR="$NEXTCLOUD_CONFIG_DIR" php -r 'require getenv("CONFIG_DIR") . "/config.php"; echo $CONFIG["version"];')
```

13.0/apache/entrypoint.sh
timLoewel:
```suggestion
    [ "$1" = "$2" ] 
```

sebastiansterk:
Is it planned to fix this issue also for v16?

J0WI:
Once this is ready we will apply it to all maintained versions.

jgeerds:
@unteem: Thank you very much for this patch. Saved me a lot of time

Is there any reason why this PR is not merged yet?

J0WI:
> Is there any reason why this PR is not merged yet?

The review comments above have not yet been resolved.

docker-entrypoint.sh
```diff
@@ -90,7 +111,7 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
                 fi
 
                 if [ "$install" = true ]; then
-                    echo "starting nextcloud installation"
+                    echo "starting nexcloud installation"
```
LoganBarnett, line 114: Typo?
```suggestion
                    echo "starting nextcloud installation"
```

13.0/apache/entrypoint.sh
```diff
@@ -90,7 +111,7 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
                 fi
 
                 if [ "$install" = true ]; then
-                    echo "starting nextcloud installation"
+                    echo "starting nexcloud installation"
```
RusmanCool, line 114: typo 'nexcloud '

15.0/fpm/entrypoint.sh
```diff
@@ -90,7 +111,7 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
                 fi
 
                 if [ "$install" = true ]; then
-                    echo "starting nextcloud installation"
+                    echo "starting nexcloud installation"
```
RusmanCool, line 114: same typo here

PR #596, opened 2019-01-05
Add Alpine variant to Dockerfile examples

J0WI:


J0WI:
I've updated the docs.

Greek64:
@J0WI 
I have a small question regarding this PR.
Why are you creating the `usr/share/man/man1` directory? Is this needed?

J0WI:
It is not needed and can probably be removed.

PR #600, opened 2019-01-11
Set Redis through environment variables

marceljd:
@J0WI added memcache.distributed

### DIFF
--- a/.examples/docker-compose/insecure/mariadb-cron-redis/apache/docker-compose.yml
+++ b/.examples/docker-compose/insecure/mariadb-cron-redis/apache/docker-compose.yml
@@ -25,6 +25,7 @@ services:
       - nextcloud:/var/www/html
     environment:
       - MYSQL_HOST=db
+      - REDIS_HOST=redis
     env_file:
       - db.env
     depends_on:
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ The install and update script is only triggered when a default command is used (
 
 If you want to use Redis you have to create a seperate [Redis](https://hub.docker.com/_/redis/) container in your setup / in your docker-compose file. To inform Nextcloud about the Redis container add:
 
-- 'REDIS_HOST' (not set by default) Name of Redis container
+- `REDIS_HOST` (not set by default) Name of Redis container
 
 The use of Redis is recommended to prevent file locking problems. See the examples for further instructions.
 
--- a/README.md
+++ b/README.md
@@ -133,8 +133,8 @@ The install and update script is only triggered when a default command is used (
 If you want to use Redis you have to create a seperate [Redis](https://hub.docker.com/_/redis/) container in your setup / in your docker-compose file. To inform Nextcloud about the Redis container add:
 
 - 'REDIS_HOST' (not set by default) Name of Redis container
-- 'REDIS_PORT' (optional, default:_6379_) Port number of Redis container. Use only if you use a non-standard port.
 
+The use of Redis is recommended to prevent file locking problems. See the examples for further instructions.
 
 # Running this image with docker-compose
 The easiest way to get a fully featured and functional setup is using a `docker-compose` file. There are too many different possibilities to setup your system, so here are only some examples what you have to look for.
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ The install and update script is only triggered when a default command is used (
 If you want to use Redis you have to create a seperate [Redis](https://hub.docker.com/_/redis/) container in your setup / in your docker-compose file. To inform Nextcloud about the Redis container add:
 
 - `REDIS_HOST` (not set by default) Name of Redis container
+- `REDIS_HOST_PORT` (default: _6379_) Optional port for Redis, only use for external Redis servers that run on non-standard ports.
 
 The use of Redis is recommended to prevent file locking problems. See the examples for further instructions.
 
--- a/README.md
+++ b/README.md
@@ -130,6 +130,10 @@ The install and update script is only triggered when a default command is used (
 
 - `NEXTCLOUD_UPDATE` (default: _0_)
 
+If you want to use Redis you have to create a seperate [Redis](https://hub.docker.com/_/redis/) container in your setup / in your docker-compose file. To inform Nextcloud about the Redis container add:
+
+- 'REDIS_HOST' (not set by default) Name of Redis container
+- 'REDIS_PORT' (optional, default:_6379_) Port number of Redis container. Use only if you use a non-standard port.
 
 
 # Running this image with docker-compose
--- a/.config/redis.config.php
+++ b/.config/redis.config.php
@@ -5,6 +5,7 @@ if (getenv('REDIS_HOST')) {
     'memcache.locking' => '\OC\Memcache\Redis',
     'redis' => array(
       'host' => getenv('REDIS_HOST'),
+      'port' => getenv('REDIS_HOST_PORT') ?: 6379,
     ),
   );
 }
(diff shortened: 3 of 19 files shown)


marceljd:
@J0WI Did you test it with v15.0 / Apache / FPM? I have to test the port settings again, it worked when I tried it a few months ago.

Do we need the port setting? The port is a default port (like 3306 for MySQL) and people who need to run their own Redis servers on different ports are probably running a non-standard version anyway.

J0WI:
> Did you test it with v15.0 / Apache / FPM?

I used `nextcloud:1{3,4,5}-apache` with `redis:alpine`.

> Do we need the port setting?

Yes, the port should be customizable.

marceljd:
@J0WI I have investigated this, and all Redis 5.0 Docker containers have a hardcoded port set (6379). Adding one variable (REDIS_HOST) is sufficient, there are no official Redis containers that use other ports. Shall I change everything to only that single variable? This is in all flavours of the Redis containers:

`EXPOSE 6379`

J0WI:
@marceljd your recent commits didn't made any changes

marceljd:
@J0WI I do not know why there are Dockerfile changes in here, I think they appeared because of the DCO correction I had to maken (force update). Is there any way to remove them? The rest should be ok: REDIS_HOST only because the port is hard-coded in the official Redis images. I also changed the examples and README files.

README.md
```diff
@@ -130,7 +130,11 @@ The install and update script is only triggered when a default command is used (
 
 - `NEXTCLOUD_UPDATE` (default: _0_)
 
+If you want to use Redis you have to create a seperate [Redis](https://hub.docker.com/_/redis/) container in your setup / in your docker-compose file. To inform Nextcloud about the Redis container add:
 
+- 'REDIS_HOST' (not set by default) Name of Redis container
```
SnowMB: change that from  `'` to back-ticks ` for better formatting.

marceljd:
Tried a few times to get the DCO error out of the way, but unfortunately the instructions given are not working...

J0WI:
> REDIS_HOST only because the port is hard-coded in the official Redis images.

This is only the case, if the nextcloud and redis container are in the same network. But you can expose port `6379` to any other port. We also support custom ports for databases.

> Tried a few times to get the DCO error out of the way, but unfortunately the instructions given are not working...

Use `git rebase --signoff master`.

SnowMB:
> I know that you can expose container ports to whatever port you want, but the default Redis containers do not support that directly. If you just want to start a default Redis container from the official image the way that is described in the examples than you do not have a choice. It runs on 6379 and this image does not support any environment variables out of the box. So that is why I think this is sufficient. MySQL has a documented way of changing the default port though an environment variable, I think that is a different case. For Redis you should use the Bitnami version instead. My solution is the easy way for the official Redis container, it just works.

Yeah but in most cases you would not expose your redis container on a network outside the host. When you use for example 2 servers (docker hosts) one running redis and the other one running nextcloud, nextcloud would connect to a port on the 2nd host. and there the redis port could be anything because of the required port mapping to the host.

It would just look like this:
```
services:
  redis:
    image: redis
    ports:
      - "11111:6379"
...
```

An other example would be someone running redis natively. 

The fix is fairly easy: Just introduce a 2nd variable called `REDIS_PORT` and default it to 6479 so for the user nothing will change but he gets the possibility to customize the port redis is running on.

In the config (untested):
```php
    'redis' => array(
      'port' => getenv('REDIS_PORT') ? getenv('REDIS_PORT')  : 6379
    ),
```

SnowMB:
Oh missed that. Maybe `REDIS_HOST_PORT` :grin:

marceljd:
>      'port' => getenv('REDIS_PORT') ? getenv('REDIS_PORT')  : 6379

I had this in the original config already (slightly different), but I am still wondering what the use case would be. In my opinion most users will start up 3 containers: SQL, webserver and redis. If you start them together they will connect on standard ports, no further config needed. The example configs and environment variables are not ment for expert users, they will change the config with custom Dockerfiles anyway. And a non-expert will not run a Redis server outside of the cluster. What you certainly do not want is non-expert users exposing host-ports to the outside world running a redis server. You want them to stay within the docker-compose boundaries, and exposing ports 80 and 443 to the outside world is more than enough. So I do not see the point of adding custom ports to this. Running Redis is faster and gets rid of file locks, so I think everyone should run 3 instead of only 2 containers, but adding custom ports will only be more complicated and insecure, especially if you are not an expert.

SnowMB:
Completely agree: *most* users will not ever use this and *most* users will run their setup on one host and *most* users may not have enough expertise to fiddle around with the details. (Including myself as I'm missing quite a bit of whats needed for proper server administration)

That said, I think in this case adding the **possibility** to change the redis port is very straight forward on both the documentation and the implementation side so that this feature is completely **opt-in**. For a normal user nothing changes and he never has to worry about setting this port to any value.

J0WI:
There is a great example at https://github.com/nextcloud/server/issues/13148#issuecomment-457654843

marceljd:
> 
> 
> There is a great example at [nextcloud/server#13148 (comment)](https://github.com/nextcloud/server/issues/13148#issuecomment-457654843)

I don't know, looks like the issue is a routing problem with CSFR errors. This should be solvable with the sticky settings in Kubernetes, I would not try to solve networking problems in an application. But perhaps I misunderstood the exact problem. Besides: they still use the default port.

@J0WI @SnowMB I added the port to the config files and the README, but I did not include them in the examples on purpose because I think these should be as simple as possible. Should I? And I still have trouble with the DCO check. I tried to rebase and signoff, but it still does not work. When it is necessary I think I need to create a new branch and change the files again, or do you have the golden hint?

J0WI:
Use `git rebase --signoff master`, fix any conflicts and `git push -f <your upstream branch>`

marceljd:
@J0WI The files are correct now but I cannot get rid of the commits that are in the branch already. I cannot get the signoff to work. I think the only possibility is to remove the entire fork and my local copies and start from scratch. I already lost too much time playing around with Git, is there any way to get this to work without trying for hours? The info on the internet was not really helpfull, tried just about everything except deleting everything and start over again.

marceljd:
I would be very happy! Next time I will be very careful not to create this kind of mess again :-)

Thank you very much!

marceljd:
@tilosp , is it safe to signoff with these commands now, or do you have to sign because you pushed?

`git rebase HEAD~8 --signoff`
`git push --force origin redis-env-vars`

tilosp:
@marceljd the commit mess is gone but there is still some sign off problem.
but this should be solvable by creating a new clone and running the commands that DCO bot recommends:
```bash
git clone https://github.com/marceljd/docker.git -b redis-env-vars
git rebase HEAD~8 --signoff
git push --force origin redis-env-vars
```

.examples/docker-compose/insecure/mariadb-cron-redis/apache/docker-compose.yml
```diff
@@ -25,6 +25,7 @@ services:
       - nextcloud:/var/www/html
     environment:
       - MYSQL_HOST=db
+      - REDIS_HOST=redis
```
tilosp: If the new config is used, the already existing one should be removed:
https://github.com/nextcloud/docker/blob/master/.examples/docker-compose/insecure/mariadb-cron-redis/apache/app/redis.config.php

J0WI:
@tilosp @SnowMB do you agree with the latest patch?

J0WI:
@tilosp could you merge this once RC are removed?

J0WI:
Please keep your fork until this is merged.

PR #632, opened 2019-01-31
Add proxy_cookie_path

danchal:
The __Host prefix mitigates cookie injection vulnerabilities within potential third-party software sharing the same second level domain.
This is one of the metrics used in the Nextcloud security scan.

This pull request adds the nginx proxy directive _proxy_cookie_path_ to the existing custom directives.
The Dockerfile COPY filename was changed from _uploadsize.conf_ to make it more generic in use.


J0WI:
Do you have any issues with cookies? The __host prefix is already supported by NextCloud and works with Docker.

danchal:
I found no cookie specific issues with the default setup, however, the Nextcloud security scan does include the __host prefix and it's not possible to achieve an A+ rating without this change.

Other than for attaining the A+ rating, I can't say if this change is necessary or not.

J0WI:
I cannot reproduce this issue using the latest image version.

PR #634, opened 2019-02-01
Fix #594: svg support for imagick on Debian

### LINKED ISSUE #594: svg support for imagick
Pull Request #575 has solved adding imagick manually to the container - thank you for this! - but imagick is missing SVG Support.
After changing my custom docker image back to the default (nextcloud:15-fpm-alpine), I am now stuck with the blue default favicons. My custom image showed the current app icon with the themed background.

Since I am new to this docker game, I cannot implement this change myself (I basically copy-pasted my code from the example docker files)

- What packages can I add manually to enable SVG support in imagick?
- What changes are needed in the official docker file to add support for SVG?
- Is SVG support something that should be added for everyone (i.E. official docker file) or should it be optional (i.E. add an example in the documentation) ?

J0WI:


J0WI:
I need to have a close look at this...

fuse314:
Is there a similar solution for alpine-based docker images?

J0WI:
`imagemagick` must be present:
https://github.com/nextcloud/docker/blob/bb105463275881d984e65ce482c141e80f2c5b5d/.examples/dockerfiles/full/fpm-alpine/Dockerfile#L7

PR #639, opened 2019-02-07
Remove app dockerfile

tilosp:
Remove the Dockerfiles that add the custom redis.config.php because the config file was removed in #600


tilosp:
> Same for the `cron` service.

should be fixed now

PR #644, opened 2019-02-08
create Apache SSL self signed certificates with environment variable

marceljd:
PR to enable SSL encryption with self-signed certificates on Apache. Multiple SSL related errors will disappear (.well.know autodiscovery, apps not working). For use behind a proxy or in private environments. For Traefik you need to set insecureSkipVerify: true in the traefik.yaml file so the self signed certificates are not rejected. The SSL encryption is activated by an environment variable to prevent existing installations to break. The certificates will be renewed every time the entrypoint script runs.

Please help with testing on different environments (behind Nginx proxy).

To do: the README and examples have to be rewritten.

### DIFF
--- a/.config/000-default.conf
+++ b/.config/000-default.conf
@@ -1,0 +1,15 @@
+<VirtualHost *:80>
+   Redirect / https://${SERVER_NAME}
+</VirtualHost>
+
+<VirtualHost *:443>
+    ServerAdmin webmaster@localhost
+    DocumentRoot /var/www/html
+    ErrorLog ${APACHE_LOG_DIR}/error.log
+    CustomLog ${APACHE_LOG_DIR}/access.log combined
+    SSLEngine on
+    SSLCertificateFile /etc/apache2/ssl/ca.crt
+    SSLCertificateKeyFile /etc/apache2/ssl/ca.key
+    Header always set Strict-Transport-Security "max-age=63072000; includeSubDomains"
+</VirtualHost>
+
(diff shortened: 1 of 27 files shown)


marceljd:
Travis error: 'override-cmd' [4/4].../entrypoint.sh: line 142: APACHE_SSL_SELFSIGNED: parameter not set

`if [ "${APACHE_SSL_SELFSIGNED}" = "true" ] ; then`

How do I make this work? The environment variable is not always set, but when it is set to true the then should kick in...

marceljd:
The environment variable you have to add to the docker-compose.yml is:
`APACHE_SSL_SELFSIGNED=true`

I will change the README and examples as soon as I have some time again. Please test this on different environments. Sorry for the pile of commits, I know how to prevent it now...

J0WI:
This is IMHO to much overhead to work around the problem. We should work with the configuration parameters provided by NextCloud. I hope we can make some progress in https://github.com/nextcloud/server/issues/13713

kesselb:
1) Documentation about service discovery: https://github.com/nextcloud/documentation/pull/1297 (IMHO redirects like this should be done by your reverse proxy)

2) A setup warning if you run nextcloud behind a reverse proxy without `trusted_proxies`: https://github.com/nextcloud/server/pull/14261

3) To be done: A warning if some required parameters not present (e.g. apache2 does not forward x-forwarded-proto while nginx and traefik does => overwriteproto is required for apache2).

PR #651, opened 2019-02-16
nginx-proxy: Set proxy_request_buffering to off. Fixes #650

### LINKED ISSUE #650: nginx proxy timeout problems
Hey,

I'm using "mariadb-cron-redis/fpm" with docker-compose and had various problems with the (linux) client timing out on uploads/downloads all the time and the upload status not updating regularly.

Adding "proxy_request_buffering off;" to the proxy config (uploadsize.conf for example) fixed all this.

Maybe it helps someone, or can be added as the default.

lazka:
Otherwise nginx buffers uploads which at least with the linux client results
in large files failing to upload and the upload stats not updating.


lazka:
Thanks!

Issue #617, opened 2019-01-23
Adding feasibility check for updates before replacing the nextcloud files


SnowMB:
I'm creating this issue to collect all the other issues related to broken updates that try to skip a major release.

Nextcloud does not support skipping a major release while updating. So going from `13->15` will not work without including version 14:  `13->14->15.` 

The problem is, that the entrypoint script correctly detects that the image version is greater, but does not check if the upgrade is feasible. So it overwrites the Nextcloud files (including `version.php`) and does not fail until running the occ upgrade command `run_as 'php /var/www/html/occ upgrade'`.

Upgrading fails and the container exits, but the user can't go back to his initial state because the Nextcloud source files are already replaced. He also can't trigger a proper update from `13->14` because the upgrade detection is broken (`version.php` now falsely claims version 15).

#597 in my opinion does not enough, but is a good start to at least document that Nextcloud might break when updating. However a check should be introduced to protect the Nextcloud install in that case. I think it is no problem, that running the new containers fails, but the Nextcloud instance should be unharmed.

related issues:
#616 
#605
#585

SnowMB:
To repair broken upgrades (a broken upgrade in that sense that a new nextcloud version is copied over the older one, but the ooc::upgrade failed), we could introduce a flag that allows downgrades (copy the older nextcloud version over the newer one in /var/www/html). So users could at least repair their setup by simply going back to their previous version or to the correct upgrade path.

nicbou:
### Fixed

I pieced a dozen old threads together and came up with a detailed solution: https://nicolasbouliane.com/blog/nextcloud-docker-upgrade-error

tl;dr: If you're stuck in that situation, do this:

1. Find your old version in config/config.php
2. Put that version everywhere in version.php
3. Roll back the docker image to that old version
4. Upgrade the image one major version at a time
5. Turn off maintenance mode

J0WI:
See also https://github.com/nextcloud/docker/issues/1129 and https://github.com/nextcloud/server/issues/2270

There is a workaround for it in https://github.com/nextcloud/docker/pull/1304

PR #660, opened 2019-02-26
Copy version.php last

### LINKED ISSUE #431: web root html dir is left half-empty on intializing the container.
I'm setting up a new docker app with k8 and `/var/www/html` on NFS mounts.

I don't know why, but sometimes some directories like `lib/` are not copied to the web root, yielding in an error like this on web, occ and status.php calls.:

```
Warning: require_once(/var/www/html/lib/versioncheck.php): failed to open stream: No such file or directory in /var/www/html/index.php on line 36

Fatal error: require_once(): Failed opening required '/var/www/html/lib/versioncheck.php' (include_path='.:/usr/local/lib/php') in /var/www/html/index.php on line 36
```

and in fact it is not there:

```bash
root@nextcloud:/var/www/html# ls /usr/src/nextcloud/
3rdparty  COPYING  config       core      custom_apps  index.html  lib  ocm-provider  ocs-provider  remote.php  robots.txt  status.php  version.php
AUTHORS   apps     console.php  cron.php  data         index.php   occ  ocs           public.php    resources   settings    themes
root@nextcloud:/var/www/html# ls /var/www/html/
3rdparty  AUTHORS  COPYING  console.php  cron.php  index.html  index.php  occ  public.php  remote.php  robots.txt  status.php  version.php
root@nextcloud:/var/www/html#
```

I could copy the missing files but I'm not sure what it causes. The directory has full (777) permissions. Here is a list of files that differ between `/var/www/html` and `/usr/src/nextcloud`:

```diff
 3rdparty
 AUTHORS
 COPYING
+apps
+config
 console.php
+core
 cron.php
+custom_apps
+data
 index.html
 index.php
-ls1
+lib
 occ
+ocm-provider
+ocs
+ocs-provider
 public.php
 remote.php
+resources
 robots.txt
+settings
 status.php
+themes
 version.php
```

Installation is done with empty directories.

I think the relevant line are the [rsync lines here](https://github.com/nextcloud/docker/blob/master/14.0-beta/apache/entrypoint.sh#L48):

```bash
    rsync $rsync_options --delete --exclude /config/ --exclude /data/ --exclude /custom_apps/ --exclude /themes/ /usr/src/nextcloud/ /var/www/html/

    for dir in config data custom_apps themes; do
        if [ ! -d "/var/www/html/$dir" ] || directory_empty "/var/www/html/$dir"; then
            rsync $rsync_options --include "/$dir/" --exclude '/*' /usr/src/nextcloud/ /var/www/html/
        fi
done
```

any ideas?
related: https://github.com/nextcloud/docker/issues/251

EDIT:
I tried with and without setting `NEXTCLOUD_DATA_DIR` because I did not see where this env is taken into account here.

Temporary fix: As I am initializing the web root, I skipped the exclude parameters like this:

```
rsync -rlDogv --chown www-data:root --delete /usr/src/nextcloud/ /var/www/html/
```

NOTE: could also be a timeout for the rsync.

tilosp:
Fixes #431

### DIFF
--- a/13.0/apache/entrypoint.sh
+++ b/13.0/apache/entrypoint.sh
@@ -51,6 +51,7 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
                 rsync $rsync_options --include "/$dir/" --exclude '/*' /usr/src/nextcloud/ /var/www/html/
             fi
         done
+        rsync $rsync_options --include '/version.php' --exclude '/*' /usr/src/nextcloud/ /var/www/html/
         echo "Initializing finished"
 
         #install
--- a/13.0/apache/upgrade.exclude
+++ b/13.0/apache/upgrade.exclude
@@ -2,3 +2,4 @@
 /data/
 /custom_apps/
 /themes/
+/version.php
--- a/13.0/fpm-alpine/entrypoint.sh
+++ b/13.0/fpm-alpine/entrypoint.sh
@@ -51,6 +51,7 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
                 rsync $rsync_options --include "/$dir/" --exclude '/*' /usr/src/nextcloud/ /var/www/html/
             fi
         done
+        rsync $rsync_options --include '/version.php' --exclude '/*' /usr/src/nextcloud/ /var/www/html/
         echo "Initializing finished"
 
         #install
--- a/13.0/fpm-alpine/upgrade.exclude
+++ b/13.0/fpm-alpine/upgrade.exclude
@@ -2,3 +2,4 @@
 /data/
 /custom_apps/
 /themes/
+/version.php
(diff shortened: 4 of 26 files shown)


SnowMB:
Just for clarification: 
This does not help with #617 and is just aimed at new instances, where the rsync fails to properly copy all files into the `/var/www/html` folder? To achieve what exactly?

I'm not really sure I understood what the problem in #431 is. Installing onto a nfs mount seems to fail sometimes, but the entrypoint script does not give an error? So some folders / files do not get copied properly? How would copying `version.php` last fix that?

Or can the error be described as a failing install on the first start of the container, but on second start there is no retry because `version.php` might already be present in the volume?

tilosp:
@SnowMB the problem is that rsync gets canceled (for example because the container stopped) 
and if the container is started again it wont run it again because version.php was already updated. this leads to an have upgrade or half initialized `/var/www/html` dir.
And yes you are right this does not help with #617

tilosp:
@J0WI simply kill the container while rsync is running (but this is hard to do on a fast machine)
then start it again, if version.php was already copied it will start the server with an half initialised webroot. With this pr it should run rsync again and copy the missing files.

Issue #620, opened 2019-01-25
Permission issue of mounted volume: Cannot write into directory


onny:
Hey, I'm trying to mount my app into the docker container with this command:
```
docker run -v /home/onny/projects/web-nextcloud-app-radio:/var/www/html/custom_apps/radio -d -p 8080:80 nextcloud
```
Unfortunately I get following error on the Nextcloud start page:
```
Error
Cannot write into "apps" directory
This can usually be fixed by giving the webserver write access to the apps directory or disabling the appstore in the config file. See https://docs.nextcloud.com/server/15/go.php?to=admin-dir_permissions
```
It is fixed after manually chown the custom_apps folder to www-data:root. But is there any other way to fix this issue?

Best regards,
Jonas

onny:
Looks like the issue is not the permission of the app folder itself, it is the custom_apps folder created as root. I guess we need a fix in the Dockerfile!
```
root@3c12c52ca62d:/var/www/html/custom_apps# ls -la
total 0
drwxr-xr-x  1 root     root  10 Mar  8 13:48 .
drwxrwxrwx  1 www-data root 428 Mar  8 13:48 ..
drwxr-xrwx 13     1000 1000 360 Mar  8 10:41 radio
```

onny:
This might fix the issue, haven't tested it yet though.

onny:
This pull request fixes the issue for me: https://github.com/nextcloud/docker/pull/686

PR #679, opened 2019-03-08
fix user permissions custom_apps directory

onny:
Fix issue https://github.com/nextcloud/docker/issues/620 mounting custom app with correct permissions


onny:
sorry this was the wrong fix

Issue #562, opened 2018-12-01
build of fpm image fails on gpg: keyserver receive failed: Cannot assign requested address


sonologic:
I'm trying to build a new image for nextcloud fpm 13.0.8 to upgrade my currently running 13.0.7. I've built the 13.0.7 image on this same system. The system is a raspberry pi 3b+

Somewhere along the way, the command `gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys 28806A878AE423A28372792ED75899B9A724937A` fails with `gpg: keyserver receive failed: Cannot assign requested address`.

Full output:

```
gmc@pi2:~/docker-compose/pi2/certbot $ docker --version
Docker version 18.09.0, build 4d60db4
gmc@pi2:~/docker-compose/pi2/certbot $ uname -a
Linux pi2.sonologic.net 4.14.79-v7+ #1159 SMP Sun Nov 4 17:50:20 GMT 2018 armv7l GNU/Linux
gmc@pi2:~/docker-compose/pi2/certbot $ cat /etc/debian_version 
9.6
gmc@pi2:~/docker-nextcloud/13.0/fpm $ docker build -t nextcloud:fpm-13.0.8 .
Sending build context to Docker daemon  19.97kB
Step 1/11 : FROM php:7.2-fpm-stretch
 ---> 85de92f4991c
Step 2/11 : RUN set -ex;         apt-get update;     apt-get install -y --no-install-recommends         rsync         bzip2         busybox-static     ;     rm -rf /var/lib/apt/lists/*;         mkdir -p /var/spool/cron/crontabs;     echo '*/15 * * * * php -f /var/www/html/cron.php' > /var/spool/cron/crontabs/www-data
 ---> Using cache
 ---> f887e514a620
[...]

PR #687, opened 2019-03-12
fix build image with gpg keyserver on ipv6 setups

onny:
This will fix https://github.com/nextcloud/docker/issues/562 building the image on IPv6 systems regarding gpg keyserver error

### DIFF
--- a/13.0/apache/Dockerfile
+++ b/13.0/apache/Dockerfile
@@ -126,6 +126,8 @@ RUN set -ex; \
         "https://download.nextcloud.com/server/releases/nextcloud-${NEXTCLOUD_VERSION}.tar.bz2.asc"; \
     export GNUPGHOME="$(mktemp -d)"; \
 # gpg key from https://nextcloud.com/nextcloud.asc
+    mkdir ~/.gnupg; \
+    echo "disable-ipv6" >> ~/.gnupg/dirmngr.conf; \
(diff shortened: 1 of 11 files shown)


tilosp:
why would disabling ipv6 fix problems on ipv6 systems? I would expect that it would break building on ipv6-only/nat64 hosts.

onny:
I got from here https://github.com/inversepath/usbarmory-debian-base_image/issues/9#issuecomment-466594168
They also use this workaround in the official roundcubemail docker container https://github.com/roundcube/roundcubemail-docker/pull/41

J0WI:
You could use `ipv4.pool.sks-keyservers.net` or `ipv6.pool.sks-keyservers.net`, but I think the issue is the missing fallback respectively the inconsistent dual stack support in the ha pool.
This problem is not limited to GnuPG, it does also affect some software repos NextCloud and GitHub themselves do not even support IPv6 (https://github.com/nextcloud/updater_server/issues/154).

We currently use the upstream https://github.com/tianon/pgp-happy-eyeballs to avoid build errors.

onny:
Okay this looks like a more appropriate workaround. So maybe we can document this in the wiki here or in the ``README.md``?

J0WI:
Meanwhile https://github.com/nextcloud/updater_server/issues/154 has been fixed. Fell free to add a hint to the docs for GnuPG.

PR #707, opened 2019-04-08
Add Redis session handler

J0WI:


tilosp:
@J0WI what advantage does this offer? Nextcloud works fine without it, so maybe performance?
Is this something officially supported? Otherwise it may cause namespace collisions between nextcloud and php inside the redis db.

J0WI:
On one hand it's a performance enhancement, on the other hand this further allows you to balance the load between multiple containers.
The default session handler is file based and would require a shared volume to achieve this.

This was previously documented [here](https://docs.nextcloud.com/server/11/admin_manual/installation/deployment_recommendations.html#session-storage), but now it has been [moved](https://portal.nextcloud.com/article/nextcloud-deployment-recommendations-7.html).

Issue #730, opened 2019-04-29
Broken Redis session handler with unix socket, can't login


epma01:
commit 83ea69d54bea72e1df9a2647ee101c73deb86d44 on Apr 26, 2019 (Add Redis session handler) is not valid if using redis with unix socket.

if setting environment

      - REDIS_HOST=/var/run/redis/redis.sock
      - REDIS_PORT=0

code in entrypoint.sh is setting a tcp configuration:

        echo "Configuring Redis as session handler"
        {
            echo 'session.save_handler = redis'
            echo "session.save_path = \"tcp://${REDIS_HOST}:${REDIS_HOST_PORT:=6379}\""
        } > /usr/local/etc/php/conf.d/redis-session.ini

this produce "Failed to start session" error when accessing to nextcloud:

{"reqId":"LnNxDDhkJ5TOCSOBw8PL","level":3,"time":"2019-04-29T19:56:23+00:00","remoteAddr":"192.168.2.10","user":"--","app":"base","method":"GET","url":"\/","message":{"Exception":"Exception","Message":"Failed to start session","Code":0,"Trace":[{"file":"\/var\/www\/html\/lib\/base.php","line":424,"function":"__construct","class":"OC\\Session\\Internal","type":"->","args":["ocmj3i1peglp"]},{"file":"\/var\/www\/html\/lib\/base.php","line":642,"function":"initSession","class":"OC","type":"::","args":[]},{"file":"\/var\/www\/html\/lib\/base.php","line":1056,"function":"init","class":"OC","type":"::","args":[]},{"file":"\/var\/www\/html\/index.php","line":40,"args":["\/var\/www\/html\/lib\/base.php"],"function":"require_once"}],"File":"\/var\/www\/html\/lib\/private\/Session\/Internal.php","Line":61,"CustomMessage":"--"},"userAgent":"Mozilla\/5.0 (Windows NT 10.0; Win64; x64; rv:66.0) Gecko\/20100101 Firefox\/66.0","version":""}

J0WI:
I assume you have a shared volume for the socket. Do you have any use case for this? The Docker concept usually makes heavy use of networks.

epma01:
Yes, I used a shared volume.

All the nexcloud service is made in a single composite container, and only fpm or proxy requires an assigned public port. Why not have all the communication inside the container (redis, mariadb and fpm) through a Unix socket and eliminate all the TCP stack overload?. Doing this just to try, it's not really a real case.

[...]

PR #734, opened 2019-05-01
Fix Broken Redis session handler with unix socket #730

epma01:
Fix Broken Redis session handler with unix socket #730

Without more patches redis can be used now in socket mode, setting REDIS_HOST to socket path and REDIS_HOST_PORT to 0

The RedisFactory.php server file must be patched if we want to avoid having to establish REDIS_HOST_PORT


epma01:
I close this pull, have a better propossal

PR #735, opened 2019-05-01
Redis patch

epma01:
Fix Broken Redis session handler with unix socket #730

Without more patches redis can be used now in socket mode, setting REDIS_HOST to socket path and REDIS_HOST_PORT to 0

The RedisFactory.php server file must be patched if we want to avoid having to establish REDIS_HOST_PORT

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -25,7 +25,11 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
         echo "Configuring Redis as session handler"
         {
             echo 'session.save_handler = redis'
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -25,7 +25,11 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
         echo "Configuring Redis as session handler"
         {
             echo 'session.save_handler = redis'
-            echo "session.save_path = \"tcp://${REDIS_HOST}:${REDIS_HOST_PORT:=6379}\""
+            if [[ ${REDIS_HOST:0:1} == "/" ]] ; then
+                echo "session.save_path = \"unix://${REDIS_HOST}?persistent=1&weight=1&database=0\""
```
J0WI: Are these extra arguments required for sockets?

J0WI:
#944

PR #756, opened 2019-05-21
Install imap php extension

varac:
In order to use the [user external](https://github.com/nextcloud/apps/tree/master/user_external) app with imap, the php imap extension needs to get installed.


J0WI:
Thanks for your contribution!
The image does not cover dependencies for all extensions. An `imap` variant is currently covered in our [examples](https://github.com/nextcloud/docker/tree/master/.examples#dockerfiles).

PR #772, opened 2019-06-07
WIP: introduce unprivileged images

### LINKED ISSUE #760: Running as non-root user (Apache Privileged Ports)
Hi everyone,

I came across this issue about running NextCloud on OpenShift as a non-root user:
https://github.com/nextcloud/docker/issues/458

That thread mentioned two PRs made to docker-library/php that were intended to fix the issue. Since the PRs were merged last year to docker-library/php, I figured that using nextcloud:latest would include these changes and there'd be some way to run the apache server on a non-privileged port (i.e., 8080).

A similar thread pointed me to documentation on running as an arbitrary user - where you'd set your RUN_APACHE_USER and RUN_APACHE_GROUP environment variables, but the container is still trying to run on port 80 even with those env variables set:
https://github.com/docker-library/docs/tree/master/php#running-as-an-arbitrary-user

When https://github.com/nextcloud/docker/issues/458 was closed, what was the resolution to get the nextcloud image running on OpenShift (as a non-root user, on a non-privileged port)?

Thanks in advance!

smueller18:
Following Docker best practices, all images should run as non-root user. This PR introduces
1. an unprivileged variant for each of the existing nextcloud images so that the existing images will not have breaking changes
2. replacement of `nginx` image with `nginxinc/nginx-unprivileged`
3. replacement of nextcloud images in the example docker-compose files with unprivileged variants

Todos:
- [ ] Use non-root images for postgres, mariadb and redis in example docker-compose files
- [ ] Test each example docker-compose
- [ ] Maybe update readme with unprivileged images
- [ ] Implement new image tags (e.g. `nextcloud:unprivileged`)

Closes #760


EWouters:
Nice work!

pierreozoux:
This is indeed nice.

Could you separate your PR on 2 commits, one with the changes and one with the update.sh?

This makes it easier to review.

Thanks :)

smueller18:
@pierreozoux I did a rebase on the latest master commit and seperated the changes into two seperat commits.

In the meantime, I changed my mind and think it is not worth the effort to support both, privileged and unprivileged container images. The default one should always be unprivileged. With that in mind, I stopped working on this PR because it is not target-oriented for me.

pierreozoux:
You are right,  everything should be unpriviledged and readonly, imho too :)

tilosp:
wouldn't it be enough to add a environment var, that changes the port? Running as an unprivileged user is already possible with `--user`.

smueller18:
Using the `--user` option is possible for the fpm image but not for the apache image because port 80 is binded by default. Replacing port 80 with 8080 during runtime in the entrypoint based on an environment variable is not possible because the responsible file `/etc/apache2/ports.conf` is owned by root. So either it has to be changed during build time (as introduced in this PR) or `ports.conf` must be mounted during runtime which is not user friendly.

jan-di:
> 
> 
> wouldn't it be enough to add a environment var, that changes the port? Running as an unprivileged user is already possible with `--user`.

Additionally to what @smueller18 said, there are a few more cases where nextcloud don't work right when you set the --user parameter. For example (at least, with the apache image):
- Redis cannot be used:
`/entrypoint.sh: 56: /entrypoint.sh: cannot create /usr/local/etc/php/conf.d/redis-session.ini: Permission denied`
- Apache IP-Rewrite cannot be disabled (as recommended when running behind a reverse proxy):
`Could not remove /etc/apache2/conf-enabled/remoteip.conf: Permission denied`

I think the only clean solution to this would be to make the entrypoint script aware for UID/GID env variables. The entrypoint could make the changes above as root and then start the services under the unprivileged UID/GID. I think this is described in #359

PR #785, opened 2019-06-26
Enable WEBP support

Flow86:
Enable WEBP support in GD module, so that processing webp files in nextcloud will work.

I created a small app https://github.com/Flow86/preview_webp to allow preview of webp files in nextcloud file browser.

### DIFF
--- a/14.0/apache/Dockerfile
+++ b/14.0/apache/Dockerfile
@@ -35,10 +35,11 @@ RUN set -ex; \
         libpq-dev \
         libxml2-dev \
         libmagickwand-dev \
+        libwebp-dev \
     ; \
     \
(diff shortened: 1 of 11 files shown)


J0WI:
This is a tricky one. Normally we use [example Dockerfiles](https://github.com/nextcloud/docker/blob/master/.examples/README.md#dockerfiles) to support additional apps.
But in this case we can't extend the already compiled GD extension.

I just asked for upstream support in https://github.com/nextcloud/server/issues/13552#issuecomment-505874797. @tilosp would you agree to extend the GD extension for WebP support?

Flow86:
> Please also add the corresponding dependency to the Alpine variant

Done in 11cc51e

J0WI:
Thanks!

Issue #792, opened 2019-06-30
SSL Reverse Proxy support?


lukasmrtvy:
I am running SSL Reverse Proxy (Traefik) in from of Nextcloud, but static content on context `/core/` is loaded over http (mixed-content errors), other content is loaded OK.

I am aware of this https://docs.nextcloud.com/server/15/admin_manual/configuration_server/reverse_proxy_configuration.html#overwrite-parameters

,but there should be option in https://github.com/nextcloud/docker/blob/master/docker-entrypoint.sh to enable reverse proxy support.

I dont want to precreate file and volume for enabling SSL Reverse Proxy support.

Will be revamped https://github.com/nextcloud/docker/pull/527 accepted?

# Run
```
docker run -d \
--name nextcloud \
--network mynetwork \
--restart=unless-stopped \
--label "traefik.enable=true" \
--label "traefik.basic.frontend.rule=Host:nextcloud.example.com" \
--label "traefik.basic.port=80" \
--label "traefik.frontend.redirect.permanent=true" \
--label "traefik.frontend.redirect.regex=https://(.*)/.well-known/(card|cal)dav" \
--label "traefik.frontend.redirect.replacement=https://$$1/remote.php/dav/" \
-e POSTGRES_DB=nextcloud \
-e POSTGRES_HOST=nextcloud-postgres \
-e POSTGRES_PASSWORD=nextcloud \
-e POSTGRES_USER=postgres \
-e NEXTCLOUD_ADMIN_PASSWORD=admin \
-e NEXTCLOUD_ADMIN_USER=admin \
-e REDIS_HOST=nextcloud-redis \
-e NEXTCLOUD_TRUSTED_DOMAINS=nextcloud.example.com \
nextcloud:latest
```

# Workaround
```
docker exec -it -u www-data nextcloud bash -c './occ config:system:set overwriteprotocol --
value="https"'
```

Heracles31:
Hey Luka,

Here is half a fix for you : 
If you enable HSTS in your SSL proxy engine, clients will take note of it and will never call back your server over HTTP again. Even should they receive a link that says HTTP, they will rewrite it themselves to HTTPS and so will call you back on HTTPS.

It is only half a fix because the server keeps sending the HTTP links and a browser that would not obey HSTS would still go after HTTP. There is also a risk should your certificate turns to not trusted for any reason or you re-install the server and need to access it clear text first time to configure it. Your own browser will refuse to connect to it because of the HSTS violation.

[...]

PR #819, opened 2019-07-26
Added support for proxy configuration using environment variables

clickkiddie:
Added environment variables NEXTCLOUD_OVERWRITEPROTOCOL, NEXTCLOUD_OVERWRITEHOST, NEXTCLOUD_OVERWRITEWEBROOT according to https://docs.nextcloud.com/server/13/admin_manual/configuration_server/reverse_proxy_configuration.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ One or more trusted domains can be set through environment variable, too. They w
 
 - `NEXTCLOUD_TRUSTED_DOMAINS` (not set by default) Optional space-separated list of domains
 
-Reverse proxy configuration according to [Admin Manual] (https://docs.nextcloud.com/server/13/admin_manual/configuration_server/reverse_proxy_configuration.html) can be achieved using:
+Reverse proxy configuration according to [Admin Manual](https://docs.nextcloud.com/server/13/admin_manual/configuration_server/reverse_proxy_configuration.html) can be achieved using:
 - `NEXTCLOUD_OVERWRITEPROTOCOL` (not set by default) Optional overwrite protocol (e.g. https)
 - `NEXTCLOUD_OVERWRITEHOST` (not set by default) Optional overwrite host (e.g. nextcloud.examle.com)
 - `NEXTCLOUD_OVERWRITEWEBROOT` (not set by default) Optional overwrite webroot (e.g. /cloud)
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ One or more trusted domains can be set through environment variable, too. They w
 - `NEXTCLOUD_TRUSTED_DOMAINS` (not set by default) Optional space-separated list of domains
 
 Reverse proxy configuration according to [Admin Manual](https://docs.nextcloud.com/server/13/admin_manual/configuration_server/reverse_proxy_configuration.html) can be achieved using:
-- `NEXTCLOUD_TRUSTED_PROXIES` (not set by default) set trusted_proxies  (e.g. 172.0.0.0/24)
+- `NEXTCLOUD_TRUSTED_PROXIES` (not set by default) Optional space-separated list of trusted_proxies  (e.g. 172.0.0.0/24)
 - `NEXTCLOUD_OVERWRITEPROTOCOL` (not set by default) Optional overwriteprotocol (e.g. https)
 - `NEXTCLOUD_OVERWRITEHOST` (not set by default) Optional overwritehost (e.g. nextcloud.examle.com)
 - `NEXTCLOUD_OVERWRITEWEBROOT` (not set by default) Optional overwritewebroot (e.g. /cloud)
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ One or more trusted domains can be set through environment variable, too. They w
 
 - `NEXTCLOUD_TRUSTED_DOMAINS` (not set by default) Optional space-separated list of domains
 
-Reverse proxy configuration according to [Admin Manual](https://docs.nextcloud.com/server/13/admin_manual/configuration_server/reverse_proxy_configuration.html) can be achieved using:
+Reverse proxy configuration according to [Admin Manual](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/reverse_proxy_configuration.html) can be achieved using:
 - `NEXTCLOUD_TRUSTED_PROXIES` (not set by default) Optional space-separated list of trusted_proxies  (e.g. 172.0.0.0/24)
 - `NEXTCLOUD_OVERWRITEPROTOCOL` (not set by default) Optional overwriteprotocol (e.g. https)
 - `NEXTCLOUD_OVERWRITEHOST` (not set by default) Optional overwritehost (e.g. nextcloud.examle.com)
--- a/README.md
+++ b/README.md
@@ -133,6 +133,11 @@ One or more trusted domains can be set through environment variable, too. They w
 
 - `NEXTCLOUD_TRUSTED_DOMAINS` (not set by default) Optional space-separated list of domains
 
+Reverse proxy configuration according to [Admin Manual] (https://docs.nextcloud.com/server/13/admin_manual/configuration_server/reverse_proxy_configuration.html) can be achieved using:
+- `NEXTCLOUD_OVERWRITEPROTOCOL` (not set by default) Optional overwrite protocol (e.g. https)
+- `NEXTCLOUD_OVERWRITEHOST` (not set by default) Optional overwrite host (e.g. nextcloud.examle.com)
+- `NEXTCLOUD_OVERWRITEWEBROOT` (not set by default) Optional overwrite webroot (e.g. /cloud)
+
 The install and update script is only triggered when a default command is used (`apache-foreground` or `php-fpm`). If you use a custom command you have to enable the install / update with
 
 - `NEXTCLOUD_UPDATE` (default: _0_)
--- a/README.md
+++ b/README.md
@@ -134,9 +134,10 @@ One or more trusted domains can be set through environment variable, too. They w
 - `NEXTCLOUD_TRUSTED_DOMAINS` (not set by default) Optional space-separated list of domains
 
 Reverse proxy configuration according to [Admin Manual](https://docs.nextcloud.com/server/13/admin_manual/configuration_server/reverse_proxy_configuration.html) can be achieved using:
-- `NEXTCLOUD_OVERWRITEPROTOCOL` (not set by default) Optional overwrite protocol (e.g. https)
-- `NEXTCLOUD_OVERWRITEHOST` (not set by default) Optional overwrite host (e.g. nextcloud.examle.com)
-- `NEXTCLOUD_OVERWRITEWEBROOT` (not set by default) Optional overwrite webroot (e.g. /cloud)
+- `NEXTCLOUD_TRUSTED_PROXIES` (not set by default) set trusted_proxies  (e.g. 172.0.0.0/24)
+- `NEXTCLOUD_OVERWRITEPROTOCOL` (not set by default) Optional overwriteprotocol (e.g. https)
+- `NEXTCLOUD_OVERWRITEHOST` (not set by default) Optional overwritehost (e.g. nextcloud.examle.com)
+- `NEXTCLOUD_OVERWRITEWEBROOT` (not set by default) Optional overwritewebroot (e.g. /cloud)
 
 The install and update script is only triggered when a default command is used (`apache-foreground` or `php-fpm`). If you use a custom command you have to enable the install / update with
 
--- a/README.md
+++ b/README.md
@@ -138,6 +138,7 @@ Reverse proxy configuration according to [Admin Manual](https://docs.nextcloud.c
 - `NEXTCLOUD_OVERWRITEPROTOCOL` (not set by default) Optional overwriteprotocol (e.g. https)
 - `NEXTCLOUD_OVERWRITEHOST` (not set by default) Optional overwritehost (e.g. nextcloud.examle.com)
 - `NEXTCLOUD_OVERWRITEWEBROOT` (not set by default) Optional overwritewebroot (e.g. /cloud)
+- `NEXTCLOUD_OVERWRITE_CLI_URL` (not set by default) Optional overwrite.cli.url (e.g. https://example.com/cloud)
 
 The install and update script is only triggered when a default command is used (`apache-foreground` or `php-fpm`). If you use a custom command you have to enable the install / update with
 
--- a/15.0/apache/entrypoint.sh
+++ b/15.0/apache/entrypoint.sh
@@ -116,27 +116,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
                         echo "installing of nextcloud failed!"
                         exit 1
                     fi
-                    if [ -n "${NEXTCLOUD_TRUSTED_DOMAINS+x}" ]; then
-                        echo "setting trusted domains…"
-                        NC_TRUSTED_DOMAIN_IDX=1
-                        for DOMAIN in $NEXTCLOUD_TRUSTED_DOMAINS ; do
-                            DOMAIN=$(echo "$DOMAIN" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
-                            run_as "php /var/www/html/occ config:system:set trusted_domains $NC_TRUSTED_DOMAIN_IDX --value=$DOMAIN"
-                            NC_TRUSTED_DOMAIN_IDX=$(($NC_TRUSTED_DOMAIN_IDX+1))
-                        done
-                    fi
-					if [ -n "${NEXTCLOUD_OVERWRITEPROTOCOL+x}" ]; then
-                        echo "setting overwriteprotocol"
-                        run_as "php /var/www/html/occ config:system:set overwriteprotocol --value=$NEXTCLOUD_OVERWRITEPROTOCOL"
-                    fi
-					if [ -n "${NEXTCLOUD_OVERWRITEHOST+x}" ]; then
-                        echo "setting overwritehost"
-                        run_as "php /var/www/html/occ config:system:set overwritehost --value=$NEXTCLOUD_OVERWRITEHOST"
-                    fi
-					if [ -n "${NEXTCLOUD_OVERWRITEWEBROOT+x}" ]; then
-                        echo "setting overwritewebroot…"
-                        run_as "php /var/www/html/occ config:system:set overwritewebroot --value=$NEXTCLOUD_OVERWRITEWEBROOT"
-                    fi
                 else
                     echo "running web-based installer on first connect!"
                 fi
@@ -152,6 +131,38 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
 
         fi
     fi
+fi
+
+#settings in config.php:
+if [ -n "${NEXTCLOUD_TRUSTED_DOMAINS+x}" ]; then
+    echo "setting trusted domains…"
+    NC_TRUSTED_DOMAIN_IDX=1
+    for DOMAIN in $NEXTCLOUD_TRUSTED_DOMAINS ; do
+        DOMAIN=$(echo "$DOMAIN" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+        run_as "php /var/www/html/occ config:system:set trusted_domains $NC_TRUSTED_DOMAIN_IDX --value=$DOMAIN"
+        NC_TRUSTED_DOMAIN_IDX=$(($NC_TRUSTED_DOMAIN_IDX+1))
+    done
+fi
+if [ -n "${NEXTCLOUD_TRUSTED_PROXIES+x}" ]; then
+    echo "setting trusted proxies…"
+    NC_TRUSTED_PROXY_IDX=1
+    for PROXY in $NEXTCLOUD_TRUSTED_PROXIES ; do
+        PROXY=$(echo "$PROXY" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+        run_as "php /var/www/html/occ config:system:set trusted_proxies $NC_TRUSTED_PROXY_IDX --value=$PROXY"
+        NC_TRUSTED_PROXY_IDX=$(($NC_TRUSTED_PROXY_IDX+1))
+    done
+fi
+if [ -n "${NEXTCLOUD_OVERWRITEPROTOCOL+x}" ]; then
+    echo "setting overwriteprotocol"
+    run_as "php /var/www/html/occ config:system:set overwriteprotocol --value=$NEXTCLOUD_OVERWRITEPROTOCOL"
+fi
+if [ -n "${NEXTCLOUD_OVERWRITEHOST+x}" ]; then
+    echo "setting overwritehost"
+    run_as "php /var/www/html/occ config:system:set overwritehost --value=$NEXTCLOUD_OVERWRITEHOST"
+fi
+if [ -n "${NEXTCLOUD_OVERWRITEWEBROOT+x}" ]; then
+    echo "setting overwritewebroot…"
+    run_as "php /var/www/html/occ config:system:set overwritewebroot --value=$NEXTCLOUD_OVERWRITEWEBROOT"
 fi
 
 exec "$@"
--- a/15.0/apache/entrypoint.sh
+++ b/15.0/apache/entrypoint.sh
@@ -164,5 +164,9 @@ if [ -n "${NEXTCLOUD_OVERWRITEWEBROOT+x}" ]; then
     echo "setting overwritewebroot…"
     run_as "php /var/www/html/occ config:system:set overwritewebroot --value=$NEXTCLOUD_OVERWRITEWEBROOT"
 fi
+if [ -n "${NEXTCLOUD_OVERWRITE_CLI_URL+x}" ]; then
+    echo "setting overwrite.cli.url"
+    run_as "php /var/www/html/occ config:system:set overwrite.cli.url --value=$NEXTCLOUD_OVERWRITE_CLI_URL"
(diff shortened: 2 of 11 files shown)


icasdri:
Is `overwrite.cli.url` needed as well?
Reference: https://help.nextcloud.com/t/stuck-in-a-login-loop-when-logging-in-from-sync-client-on-linux/47549

clickkiddie:
> Is `overwrite.cli.url` needed as well?
> Reference: https://help.nextcloud.com/t/stuck-in-a-login-loop-when-logging-in-from-sync-client-on-linux/47549

My setup runs with NEXTCLOUD_TRUSTED_DOMAINS, NEXTCLOUD_OVERWRITEPROTOCOL and NEXTCLOUD_OVERWRITEWEBROOT and I can access it using webfrontend, windows desktoop app and andoid app without issues. I don't think `overwrite.cli.url` is required. It isn't even mentioned in the nextcloud manual.

lukasmrtvy:
Fixes https://github.com/nextcloud/docker/issues/792

lukasmrtvy:
You should also consider to implement support for trusted_proxies `NEXTCLOUD_TRUSTED_PROXIES`
https://docs.nextcloud.com/server/16/admin_manual/configuration_server/reverse_proxy_configuration.html#defining-trusted-proxies

clickkiddie:
> You should also consider to implement support for trusted_proxies `NEXTCLOUD_TRUSTED_PROXIES`

`NEXTCLOUD_TRUSTED_PROXIES` is already included in master branch

lukasmrtvy:
where ? 
https://github.com/nextcloud/docker/blob/master/16.0/apache/entrypoint.sh

clickkiddie:
> where ?
> https://github.com/nextcloud/docker/blob/master/16.0/apache/entrypoint.sh

you are correct. It is missing. I mixed it up with 'trusted_domains'. 'trusted_proxies' should be set for a sane configuration.  I'll add it when I have time to test it

clickkiddie:
> > You should also consider to implement support for trusted_proxies `NEXTCLOUD_TRUSTED_PROXIES`
> 
> `NEXTCLOUD_TRUSTED_PROXIES` is already included in master branch

`NEXTCLOUD_TRUSTED_PROXIES`  has been added
additionally the config setup has been moved to a region where it is also done after install (so you can update container configuration while keeping the volumes)

following tests have been performed:
1. setting NEXTCLOUD_TRUSTED_PROXIES disables security warning in  admin panel -> PASS
2. setting reverse proxy config before install -> PASS
3. setting reverse proxy config after install -> PASS
4. setting incorrect trusted proxy triggers security warning-> FAIL (it seems that the config-check only checks if `trusted_proxies` is present, but not if it actually matches the proxy used)
5. checking IP-addresses in access log -> FAIL (Proxy IP is displayed there, expected Client-IP. I guess this is related to my proxy config. I haven't checked the actual passed headers, yet)

I would appreciate if somebody could recheck  5. and confirm if the feature works as expected.

Apache image for testing will be available on Dockhub (`clickkiddie/nextcloud:latest`) in about 30 min (upload takes time)

gobengo:
@clickkiddie Will you rebase/merge from master so this is mergeable without write conflicts?

clickkiddie:
> @clickkiddie Thanks for making this. I can't wait to use it! Will you rebase/merge from master so this PR is mergeable without write conflicts?

I'll have a look on that during the coming week

clickkiddie:
sorry, I recently messed things up during merge. Now it should be in sync again.
@gobengo please provide some feedback if that works for you, maybe it increases our chances to get this PR integrated by the maintainers

gobengo:
@J0WI @tilosp y'all have been merging PRs recently. Will you consider merging this or providing feedback? It would really help with using the docker container behind reverse proxies, which I think is pretty common!

Thank you :)

kesselb:
> additionally the config setup has been moved to a region where it is also done after install (so you can update container configuration while keeping the volumes)

Thanks :+1: 

>     4. setting incorrect trusted proxy triggers security warning-> FAIL (it seems that the config-check only checks if `trusted_proxies` is present, but not if it actually matches the proxy used)

Yes: https://github.com/nextcloud/server/pull/14606

>     6. checking IP-addresses in access log (trusted proxy incorrect) -> FAIL (real client IPs are shown; expected REAL-IP-HEADER is ignored)

https://github.com/nextcloud/server/blob/fdf4e1ebb2a458f767b57abddac337d1b92f45b6/lib/private/AppFramework/Http/Request.php#L642-L665

> is that intended behavior?

I don't think so. You should see the ip of the reverse proxy in this case. 

> I don't think overwrite.cli.url is required. It isn't even mentioned in the nextcloud manual.

It's good to set it but not required if overwriteprotocol and overwritehost are present. Nextcloud uses the value to generate absolute urls from cli (e.g cron, background job). https://github.com/nextcloud/server/blob/b8fcf6e9b3d492d19bbbdcd144428fef9bbf2324/lib/private/URLGenerator.php#L242-L250 `getBaseUrl` will do the fallback to overwriteprotocl and overwritehost.

> NEXTCLOUD_TRUSTED_PROXIES

Could you consider to set a default value? I prepared a pull request for it https://github.com/nextcloud/docker/pull/891 but with a different approach. This one (setting the values on container start) is better. Nginx and Traefik set X-Forwarded-Host and X-Forwarded-Protocol by default. If the request is coming from a trusted proxy overwriteprotocol and overwritehost are superfluous. Not sure if overwrite.cli.url is required then :thinking: (becaue the fallback to overwrite* is missing).

> wants to merge 39 commits into 

Could you try to cleanup the git history a bit? You can use rebase for it.

clickkiddie:
> > NEXTCLOUD_TRUSTED_PROXIES
> Could you consider to set a default value?
I don't think a default should be set. It might introduce security issues when used without reverse proxy.

> > 6. checking IP-addresses in access log (trusted proxy incorrect) -> FAIL (real client IPs are shown; expected REAL-IP-HEADER is ignored)
> > 
https://github.com/nextcloud/server/blob/fdf4e1ebb2a458f767b57abddac337d1b92f45b6/lib/private/AppFramework/Http/Request.php#L642-L665
> 
> I don't think so. You should see the ip of the reverse proxy in this case.

I am not able to reproduce this one anymore. Maybe I did something wrong during my tests. Code looks good on first glance, despite scenarios where multiple forwarded-for-headers are defined in config.

> Could you try to cleanup the git history a bit? You can use rebase for it.
done

kesselb:
> I don't think a default should be set. It might introduce security issues when used without reverse proxy.

With [private IPv4 addresses](https://en.wikipedia.org/wiki/Private_network)?

clickkiddie:
> > I don't think a default should be set. It might introduce security issues when used without reverse proxy.
> 
> With [private IPv4 addresses](https://en.wikipedia.org/wiki/Private_network)?

Sure. Many larger networks in Universities and Companies have their whole infrastructure - including workstations and wifi dhcp range- in these ranges. Having e.g. 10.x.x.x trusted per default would allow a huge number of clients to spoof being a proxy.

kesselb:
Convinced ;) `overwrite.cli.url` would be nice :see_no_evil:

clickkiddie:
> `overwrite.cli.url` would be nice 🙈

I added it, however I don't fully understand that paramter, so please do some testing.

As I understand it the logic is like that:
1. default: build urls from HTTP header
2. if overwriteprotocol/-host/-webroot is set, use these ones instread
3. if overwrite.cli.url is set replace url from 1 or 2. by that one

kesselb:
Thanks again :+1: 

> I don't fully understand that paramter,

No one does because the documentation is bad :see_no_evil: 

**Example 1: Nextcloud with Nginx or Traefik as reverse proxy**

1) Add the ip of the reverse proxy to trusted_proxies. Need this to read the correct value for the client ip (usually from X-Forwarded-For) for features like bruteforce detection, logging your client ip, etc. Nextcloud also accepts `X-Forwarded-Protocol` and `X-Forwarded-Host` from trusted proxies.

2) If you run the nextcloud cron via Ajax or Webcron you are done :tada: For [cron](https://github.com/nextcloud/docker/tree/master/.examples/dockerfiles/cron) `overwrite.cli.url` is required because the request is not coming from the reverse proxy and thus `X-Forwarded-Protocol` and `X-Forwarded-Host` not present. Nextcloud needs this value to generate notifications or emails from background jobs (not sure if this is the case somewhere).

**Example 2: Nextcloud with Apache2 as reverse proxy**

1) Add the ip of the reverse proxy to trusted_proxies. Need this to read the correct value for the client ip (usually from X-Forwarded-For) for features like bruteforce detection, logging your client ip, etc. Nextcloud also accepts `X-Forwarded-Protocol` and `X-Forwarded-Host` from trusted proxies.

2) Apache2 set `X-Forwarded-Host` but not `X-Forwarded-Protocol`. For this scenario you need `overwriteprotocol`.

3) If you run the nextcloud cron via Ajax or Webcron you are done :tada: For [cron](https://github.com/nextcloud/docker/tree/master/.examples/dockerfiles/cron) `overwrite.cli.url` is required because the request is not coming from the reverse proxy and thus `X-Forwarded-Protocol` and `X-Forwarded-Host` not present. Nextcloud needs this value to generate notifications or emails from background jobs (not sure if this is the case somewhere).

**But beware:** There is a [magic setup check ](https://github.com/nextcloud/server/blob/2d697332d37fb1e179555cd16f9eb189c5b95715/apps/settings/lib/Controller/CheckSetupController.php#L475-L486)(which runs if you login as administrator for example) that detects the current url and set it as `overwrite.cli.url` if the value is empty. 

**WTF?:**

Your domain is reachable at https://cloud.domain.com via reverse proxy (192.168.10.2)

1) Add `192.168.10.2` to trusted_proxies
2) Set `overwriteprotcol` to `https`
3) Set `overwritehost` to `cloud.domain.com`
4) Set `overwrite.url.cli` to `https://cloud.domain.com`

kesselb:
Turns out the situation with `trusted_proxies` is even more complex :see_no_evil: 

https://github.com/nextcloud/docker/blob/6d007ebd0bdedd444414d10ce804a416056dc54e/17.0/apache/Dockerfile#L105-L112

@domdorn pointed me to `remoteip` that is enabled by default for the apache image. If the reverse proxy is running in a private network and set `X-Real-IP` ([Nginx :heavy_check_mark:](https://www.nginx.com/resources/wiki/start/topics/examples/forwarded/), [Traefik :x:](https://docs.traefik.io/routing/entrypoints/#forwarded-header), [Apache2.4 :x:](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#x-headers)) the client ip is already correct but `x-forwarded-protcol` and `x-forwarded-host` do not work (because the client ip is not the ip of the trusted proxy). Added an example to my previous post.

domdorn:
I would suggest to remove the `/etc/apache2/conf-available/remoteip.conf` from the docker build, as it is unexpected behaviour in the current implementation (either all should work by default or nothing).

kesselb:
> I would suggest to remove the `/etc/apache2/conf-available/remoteip.conf` from the docker build, as it is unexpected behaviour in the current implementation (either all should work by default or nothing).

Prabably ;) But this is out of scope for this pull request (sorry for hijacking @clickkiddie :see_no_evil:) and will break many instances.

clickkiddie:
> Turns out the situation with `trusted_proxies` is even more complex 🙈
> 
> https://github.com/nextcloud/docker/blob/6d007ebd0bdedd444414d10ce804a416056dc54e/17.0/apache/Dockerfile#L105-L112
> 
> @domdorn pointed me to `remoteip` that is enabled by default for the apache image. If the reverse proxy is running in a private network and set `X-Real-IP` ([Nginx ✔️](https://www.nginx.com/resources/wiki/start/topics/examples/forwarded/), [Traefik ❌](https://docs.traefik.io/routing/entrypoints/#forwarded-header), [Apache2.4 ❌](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#x-headers)) the client ip is already correct but `x-forwarded-protcol` and `x-forwarded-host` do not work (because the client ip is not the ip of the trusted proxy). Added an example to my previous post.

Ah, that might explain, why I have seen Real-IPs in the logs when I set a wrong trusted proxy range.

In my test deployment I use the following config with traefik and PassHostHeader enabled:

> - NEXTCLOUD_TRUSTED_DOMAINS=<subdomain.domain.com>
      - NEXTCLOUD_TRUSTED_PROXIES=172.0.0.0/8
      - NEXTCLOUD_OVERWRITEPROTOCOL=https
      - NEXTCLOUD_OVERWRITEWEBROOT=/cloud

With this setup I observe the following behavior:
- nextcloud on https://<subdomain.domain.com>/cloud works via webfrontend and links displayed are correct (even when NEXTCLOUD_TRUSTED_PROXIES is not set)
- nextcloud filesync via windows and android app works
- nextcloud-cron (seperate container) seems to work (have not tested, advanced features such as mail-sending)
- command-line log shows always proxy IP (that was better in older versions; something seems to have changed)
- Log messages in the nextcloud-UI show REAL-IPs
- there are some complaints about reachability of .well-konwn-URLs from the config checker, but I dindn't experience problems with my file sync apps

summary:
In its current status, this PR can solve basic scenarios. the remote remoteip.conf-thing might be an issue for security and proxies in other address ranges.

README.md
```diff
@@ -133,6 +133,13 @@ One or more trusted domains can be set through environment variable, too. They w
 
 - `NEXTCLOUD_TRUSTED_DOMAINS` (not set by default) Optional space-separated list of domains
 
+Reverse proxy configuration according to [Admin Manual](https://docs.nextcloud.com/server/13/admin_manual/configuration_server/reverse_proxy_configuration.html) can be achieved using:
+- `NEXTCLOUD_TRUSTED_PROXIES` (not set by default) set trusted_proxies  (e.g. 172.0.0.0/24)
```
kesselb:
```suggestion
- `NEXTCLOUD_TRUSTED_PROXIES` (not set by default) Optional space-separated list of trusted_proxies  (e.g. 172.0.0.0/24)
```

README.md
```diff
@@ -133,6 +133,13 @@ One or more trusted domains can be set through environment variable, too. They w
 
 - `NEXTCLOUD_TRUSTED_DOMAINS` (not set by default) Optional space-separated list of domains
 
+Reverse proxy configuration according to [Admin Manual](https://docs.nextcloud.com/server/13/admin_manual/configuration_server/reverse_proxy_configuration.html) can be achieved using:
```
kesselb:
```suggestion
Reverse proxy configuration according to [Admin Manual](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/reverse_proxy_configuration.html) can be achieved using:
```

kesselb:
```
docker run -p 8080:80 -e NEXTCLOUD_TRUSTED_DOMAINS="test.test" nextcloud:docker-test 
Initializing nextcloud 17.0.0.9 ...
Initializing finished
New nextcloud instance
setting trusted domains…
Nextcloud is not installed - only a limited number of commands are available
                                                                   
  There are no commands defined in the "config:system" namespace. 
```

If you decide against the cli installer it does not work because nextcloud is not installed. 

```
diff --git a/docker-entrypoint.sh b/docker-entrypoint.sh
index 9514d88..1f759ce 100755
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -116,15 +116,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
                         echo "installing of nextcloud failed!"
                         exit 1
                     fi
-                    if [ -n "${NEXTCLOUD_TRUSTED_DOMAINS+x}" ]; then
-                        echo "setting trusted domains…"
-                        NC_TRUSTED_DOMAIN_IDX=1
-                        for DOMAIN in $NEXTCLOUD_TRUSTED_DOMAINS ; do
-                            DOMAIN=$(echo "$DOMAIN" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
-                            run_as "php /var/www/html/occ config:system:set trusted_domains $NC_TRUSTED_DOMAIN_IDX --value=$DOMAIN"
-                            NC_TRUSTED_DOMAIN_IDX=$(($NC_TRUSTED_DOMAIN_IDX+1))
-                        done
-                    fi
                 else
                     echo "running web-based installer on first connect!"
                 fi
@@ -140,6 +131,46 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
 
         fi
     fi
+
+    if [ "$installed_version" != "0.0.0.0" ]; then
+        #settings in config.php:
+        if [ -n "${NEXTCLOUD_TRUSTED_DOMAINS+x}" ]; then
+            echo "setting trusted domains…"
+            NC_TRUSTED_DOMAIN_IDX=1
+            for DOMAIN in $NEXTCLOUD_TRUSTED_DOMAINS ; do
+                DOMAIN=$(echo "$DOMAIN" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+                run_as "php /var/www/html/occ config:system:set trusted_domains $NC_TRUSTED_DOMAIN_IDX --value=$DOMAIN"
+                NC_TRUSTED_DOMAIN_IDX=$(($NC_TRUSTED_DOMAIN_IDX+1))
+            done
+        fi
+        if [ -n "${NEXTCLOUD_TRUSTED_PROXIES+x}" ]; then
+            echo "setting trusted proxies…"
+            NC_TRUSTED_PROXY_IDX=1
+            for PROXY in $NEXTCLOUD_TRUSTED_PROXIES ; do
+                PROXY=$(echo "$PROXY" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+                run_as "php /var/www/html/occ config:system:set trusted_proxies $NC_TRUSTED_PROXY_IDX --value=$PROXY"
+                NC_TRUSTED_PROXY_IDX=$(($NC_TRUSTED_PROXY_IDX+1))
+            done
+        fi
+        if [ -n "${NEXTCLOUD_OVERWRITEPROTOCOL+x}" ]; then
+            echo "setting overwriteprotocol"
+            run_as "php /var/www/html/occ config:system:set overwriteprotocol --value=$NEXTCLOUD_OVERWRITEPROTOCOL"
+        fi
+        if [ -n "${NEXTCLOUD_OVERWRITEHOST+x}" ]; then
+            echo "setting overwritehost"
+            run_as "php /var/www/html/occ config:system:set overwritehost --value=$NEXTCLOUD_OVERWRITEHOST"
+        fi
+        if [ -n "${NEXTCLOUD_OVERWRITEWEBROOT+x}" ]; then
+            echo "setting overwritewebroot…"
+            run_as "php /var/www/html/occ config:system:set overwritewebroot --value=$NEXTCLOUD_OVERWRITEWEBROOT"
+        fi
+        if [ -n "${NEXTCLOUD_OVERWRITE_CLI_URL+x}" ]; then
+            echo "setting overwrite.cli.url"
+            run_as "php /var/www/html/occ config:system:set overwrite.cli.url --value=$NEXTCLOUD_OVERWRITE_CLI_URL"
+        fi
+    fi
 fi
 
+
+
 exec "$@"
```

Above works for me for this special case. It's required to restart the container for these changes. We might need to print a hint about that: "Please visit your nextcloud instance, finish the setup and restart the docker container".

volker-raschek:
Hello, 
I believe I have the same problem with nginx as trusted proxy. See my post in the  [nextcloud forum](https://help.nextcloud.com/t/docker-nextcloud-setup-nginx-nginx-fpm-ip-forwarding-does-not-work/63291). 

To define the trusted proxy about his ip adress it's would be nice to use his dns name, because his ip address can be changed if the environment would be restarted.

clickkiddie:
> Hello,
> I believe I have the same problem with nginx as trusted proxy. See my post in the [nextcloud forum](https://help.nextcloud.com/t/docker-nextcloud-setup-nginx-nginx-fpm-ip-forwarding-does-not-work/63291).
> 
> To define the trusted proxy about his ip adress it's would be nice to use his dns name, because his ip address can be changed if the environment would be restarted.

You can specify an IP range using CIDR notation (e.g.  172.16.0.0/16 ) (see https://docs.nextcloud.com/server/15/admin_manual/configuration_server/reverse_proxy_configuration.html). The IP might change between restarts, but the network prefix should be always the same.

JonasBVS:
@clickkiddie can you complete the signoff?

it looks like thats what is missing for this PR to be completed!?

clickkiddie:
> @clickkiddie can you complete the signoff?
> 
> it looks like thats what is missing for this PR to be completed!?

did it. sorry for latency, i'm quite busy currently.

can somebody comment / verify the latest test findings from @kesselb ? Which configs work, which not? Do we need to improve the config script or is a doc update sufficient? Currently I have no test system available to work on it.

JonasBVS:
> > @clickkiddie can you complete the signoff?
> > it looks like thats what is missing for this PR to be completed!?
> 
> did it. sorry for latency, i'm quite busy currently.
> 
> can somebody comment / verify the latest test findings from @kesselb ? Which configs work, which not? Do we need to improve the config script or is a doc update sufficient? Currently I have no test system available to work on it.

@clickkiddie i can try verifying or commenting, but is there a procedure for this? which image should i use for testing?

besides the travis-ci builds seems to fail at this point!?

mschoettle:
I have exactly this problem (fpm-alpine, nginx & traefik).

How can this PR be tested?

J0WI:
> How can this PR be tested?

- clone this repo
- apply the patch locally
- run `docker build -t nextcloud:patch-819 17.0/fpm-alpine`
- run the specified version (`nextcloud:patch-819`) of the container

mschoettle:
Thanks!

Setting these environment variables on a new instance causes the following problem:

```
nextcloud-test-app | setting trusted proxies…
nextcloud-test-app | Nextcloud is not installed - only a limited number of commands are available
nextcloud-test-app | 
nextcloud-test-app |                                                                    
nextcloud-test-app |   There are no commands defined in the "config:system" namespace.  
nextcloud-test-app |                                                                    
nextcloud-test-app | 
nextcloud-test-app exited with code 1
```

Without the environment variables (to install it first) I first get "Access through untrusted domain" error when accessing the web interface. For some reason right now even if I add the domain manually in `config.php` it still tells me it is an untrusted domain.

mschoettle:
Setting `NEXTCLOUD_TRUSTED_DOMAINS` works in master.

There should probably be a check before setting the configs based on the environment variables to ensure Nextcloud is installed. However, at the end of the script `"$installed_version" != "0.0.0.0"` so might have to check with `occ status` (which will contain ` - installed: false` in the output) but maybe there's an easier way.

kesselb:
> maybe there's an easier way.

Not messing around with occ ;) I was thinking about putting all those reverse proxy configurations into a separate config file similar to https://github.com/nextcloud/docker/blob/master/.config/smtp.config.php. That's probably less performant but much easier. And works better with the weird way nextcloud handles configuration. But time ... :disappointed:

mschoettle:
> > maybe there's an easier way.
> 
> Not messing around with occ ;)

I meant to just "parse" the output of the `status` command to see whether Nextcloud is installed.

schildbach:
What's the state of this PR? Due to Nextcloud not honoring the `X-Forwarded-Proto` proxy header and also the `NEXTCLOUD_OVERWRITEPROTOCOL` not yet present, Nextcloud currently cannot be used within Docker and behind an SSL proxy. It throws around URLs that link back to unsecure HTTP.

kesselb:
> What's the state of this PR? Due to Nextcloud not honoring the `X-Forwarded-Proto` proxy header and also the `NEXTCLOUD_OVERWRITEPROTOCOL` variable not yet present, Nextcloud currently cannot be used within Docker and behind an SSL proxy. It throws around URLs that link back to unsecure HTTP.

You can set most configuration variables via ENV by prefixing `NC_`. `NC_overwriteprotocol` for that case. Hopefully it will replaced by `NEXTCLOUD_OVERWRITEPROTOCOL` sometime but for the time beeing.

schildbach:
`NC_overwriteprotocol` works. Thanks!

t-abraham:
Can anyone please tell me the ENV variables to required to set for the official Nextcloud:latest image so that my config.php have these two settings from the fresh install (initial run)

'overwriteprotocol' => 'https',
'filelocking.enabled' => false,

J0WI:
> Can anyone please tell me the ENV variables to required to set for the official Nextcloud:latest image so that my config.php have these two settings from the fresh install (initial run)

As you can see this PR is not yet merged, so there are currently no ENVs for these settings. You can mount additional config files in the config folders.

thannaske:
Is there any plan to complete this (imho very important PR)? Without those environment variables you are not able to properly deploy this image in a high-scale k8s-environment.

t-abraham:
> Is there any plan to complete this (imho very important PR)? Without those environment variables you are not able to properly deploy this image in a high-scale k8s-environment.

I don't see any movements for the time being. But on the other hand I developed a custom php to add any variables required just pass through the ENV file of the stack. Knock me stating your required variables, if you need one.

J0WI:
Recent progress have been made in #1048.

J0WI:
Fixed in #1048.

Kr4cker:
> Your domain is reachable at https://cloud.domain.com via reverse proxy (192.168.10.2)
> 
> Add 192.168.10.2 to trusted_proxies
> Set `overwriteprotcol` to https
> Set `overwritehost` to cloud.domain.com
> Set `overwrite.url.cli` to https://cloud.domain.com
> I hope this helps a little to understand this topic.

Just a small hint for the future people coming to this thread.
The [entry](https://github.com/nextcloud/docker/pull/819#issuecomment-541434657) of @kesselb helped me to solve my login issue.
I had a hard time at the beginning cause I just copy pasted the attribute names from here.
But the `overwriteprotocol` is missing an `o`. After fixing this typo, my Nextcloud login worked =)

yassinwallace:
@Kr4cker Thank you for your input. It helped me.
I think it is overwrite.cli.url and not overwrite.url.cli.
I'm a total newbie on the subject and this is my very first comment on Git. So I might be wrong.

Kr4cker:
@yassinwallace yes you are right. I checked the [Nextcloud documentation](https://docs.nextcloud.com/server/19/admin_manual/configuration_server/config_sample_php_parameters.html?#proxy-configurations).

Issue #593, opened 2018-12-30
Auto-configuration based on environment variables does not work


pasikarkkainen:
It seems auto-configuration using environment variables does not work when using docker image "nextcloud:15.0.0-apache".

docker-compose.yml:

```
version: "2"

services:

  nextcloud:
    image: nextcloud:15.0.0-apache
    container_name: nextcloud
    expose:
      - "80"
    links:
      - mariadb
    depends_on:
      - mariadb
    volumes:
      - "./data:/var/www/html:z"
    environment:
      - MYSQL_HOST=mariadb
      - MYSQL_DATABASE=nextcloud
      - MYSQL_USER=nextcloud
      - MYSQL_PASS=secretpassword
      - NEXTCLOUD_ADMIN_USER=ncadmin
      - NEXTCLOUD_ADMIN_PASSWORD=supersecretpassword
    restart: always
```

Results in the first-time wizard having empty admin user/password, sqlite as the chosen database, and when one chooses mysql the mysql infos aren't filled in. So for some reason all those environment variables are ignored by the first-time wizard.

If I start a bash shell in the container, I can see the environment variables there just fine.

chrisallenlane:
I am encountering the same issue.

J0WI:
I'm not able to reproduce this

darkdan21:
I'm having the same issue using the example docker container configuration (tried both docker-compose/with-nginx-proxy/mariadb/fpm and docker-compose/with-nginx-proxy/postgres/fpm)

ReneCD:
I'm having the same problem.  The variables are simply being ignored.  Any idea why so many of us have this problem even though you can't reproduce it ?

J0WI:
I used the example in https://github.com/nextcloud/docker/issues/593#issue-394863856

ReneCD:
Example in #593?  That is this thread ?!?!

dve:
I got the same problem, but only on an existing installation. Copy & Pasting the nextcloud parts into a new docker-compose file and starting it works like a charme...

SnowMB:
Auto configuration will not overwrite settings if you already have an existing installation. If you are migrating or already have installed nextcloud, you have to edit the config file manually by attaching to the container or copying additional files into the container/volume.

ReneCD:
[...]

PR #835, opened 2019-08-13
Change sleep time from 3s to 10s in all entrypoint.sh files

### LINKED ISSUE #833: Nextcloud installation fails, when mariadb is not ready: Sleep is too short
The installation of nextcloud (docker-compose with mariadb) sometimes fails, because the database is not ready.

The defaults of sleep (3s) and max_retries (10) in docker-entrypoint.sh are too short for some cases.

Increasing the sleep time to 10s solves the issue for me.

If the database is ready early, it only extends the installation-time a few seconds and if not, it could help others, where the database needs more time.

Nils160988:
Fixes #833, as mentioned there, in some cases the sleep timeout ist too short.

It is my first PR, please be patient and tell me, if my approach is not as desired ;)

### DIFF
--- a/.examples/docker-compose/insecure/postgres/apache/db.env
+++ b/.examples/docker-compose/insecure/postgres/apache/db.env
@@ -1,2 +1,3 @@
 POSTGRES_PASSWORD=
-POSTGRES_USER=postgres
+POSTGRES_DB=nextcloud
+POSTGRES_USER=nextcloud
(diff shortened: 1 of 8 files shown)


J0WI:
... or should we just change `max_retries`? @tilosp

Nils160988:
Changing max_retries was a solution I also considered. But I thought it will just fill the logs so that increasing sleep timeout is favorable.

J0WI:
On the other hand, an increased timeout will also consume more time if something is wrong.

J0WI:
Related issue: https://github.com/nextcloud/docker/issues/593#issuecomment-573260560
cc @tilosp

tilosp:
:+1: it makes sense to merge this
once the MariaDB bug is fixed we can revert back to 3s

J0WI:
Sorry, I messed up during rebase. Created #963 for it.

PR #863, opened 2019-09-14
Add gmp php extension

### LINKED ISSUE #854: Bookmarks is missing php-gmp
Bookmarks now requires manually installing a required dependency which isn't optimal for a container (see https://github.com/nextcloud/bookmarks/issues/792). Is it practical to include the `php-gmp` extension in the docker build?

The error reads:
> App "Bookmarks" cannot be installed because the following dependencies are not fulfilled: The library gmp is not available

marcelklehr:
fixes #854

### DIFF
--- a/15.0/apache/Dockerfile
+++ b/15.0/apache/Dockerfile
@@ -37,10 +37,13 @@ RUN set -ex; \
         libmagickwand-dev \
         libzip-dev \
         libwebp-dev \
+        libgmp-dev \
     ; \
     \
     debMultiarch="$(dpkg-architecture --query DEB_BUILD_MULTIARCH)"; \
+    if [ ! -e /usr/include/gmp.h ]; then ln -s /usr/include/$debMultiarch/gmp.h /usr/include/gmp.h; fi;\
     docker-php-ext-configure gd --with-freetype-dir=/usr --with-png-dir=/usr --with-jpeg-dir=/usr --with-webp-dir=/usr; \
+    docker-php-ext-configure gmp --with-gmp="/usr/include/$debMultiarch"; \
     docker-php-ext-configure ldap --with-libdir="lib/$debMultiarch"; \
     docker-php-ext-install -j "$(nproc)" \
         exif \
@@ -52,11 +55,12 @@ RUN set -ex; \
         pdo_mysql \
         pdo_pgsql \
         zip \
+        gmp \
     ; \
     \
 # pecl will claim success even if one install fails, so we need to perform each install separately
     pecl install APCu-5.1.17; \
-    pecl install memcached-3.1.3; \
+    pecl install memcached-3.1.4; \
     pecl install redis-4.3.0; \
     pecl install imagick-3.4.4; \
     \
--- a/.examples/dockerfiles/full/apache/Dockerfile
+++ b/.examples/dockerfiles/full/apache/Dockerfile
@@ -20,16 +20,13 @@ RUN set -ex; \
     apt-get install -y --no-install-recommends \
         libbz2-dev \
         libc-client-dev \
-        libgmp3-dev \
         libkrb5-dev \
         libsmbclient-dev \
     ; \
     \
     docker-php-ext-configure imap --with-kerberos --with-imap-ssl; \
-    ln -s "/usr/include/$(dpkg-architecture --query DEB_BUILD_MULTIARCH)/gmp.h" /usr/include/gmp.h; \
     docker-php-ext-install \
         bz2 \
-        gmp \
         imap \
     ; \
     pecl install smbclient; \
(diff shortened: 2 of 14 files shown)


marcelklehr:
Mmh. based on the travis matrix, it's not really clear to me, what's the problem. Debian fpm x86 always seems to fail, with other targets failing at random it seems :thinking:

maxmeyer:
@marcelklehr Did you try to build those fauling images locally?

J0WI:
Debian uses different `include` locations for each architecture

15.0/apache/Dockerfile
```diff
@@ -37,10 +37,13 @@ RUN set -ex; \
         libmagickwand-dev \
         libzip-dev \
         libwebp-dev \
+        libgmp-dev \
     ; \
     \
     debMultiarch="$(dpkg-architecture --query DEB_BUILD_MULTIARCH)"; \
+    test -f "/usr/include/gmp.h" || ln -s "/usr/include/$debMultiarch/gmp.h" /usr/include/gmp.h; \
     docker-php-ext-configure gd --with-freetype-dir=/usr --with-png-dir=/usr --with-jpeg-dir=/usr --with-webp-dir=/usr; \
+    docker-php-ext-configure gmp --with-gmp="/usr/include/$debMultiarch"; \
```
marcelklehr: @J0WI like this? :thinking:

marcelklehr:
@maxmeyer I tried yes, but it's not quite easy to do and they fail locally as well.

J0WI:
@tianon do you know any solution for the architecture relative library paths on Debian?

norweeg:
@J0WI does this help at all? https://wiki.debian.org/Multiarch/LibraryPathOverview
I may not understand all that's going on in this PR, but I am anxious to see this merged

norweeg:
Pretty sure I found the problem.  

@marcelklehr, the lines where you have 
`test -f "/usr/include/gmp.h" || ln -s "/usr/include/$debMultiarch/gmp.h" /usr/include/gmp.h; \`
or similar, i think it should be changed to 
`test -e "/usr/include/gmp.h" || ln -s "/usr/include/$debMultiarch/gmp.h" /usr/include/gmp.h; \`

test -f means the file exists AND is a regular file, which based on the fact that you create a symbolic link on failure means that this file will never be a regular file.  test -e is a check for simple existence of a file with no other conditions.  If you want to check if /usr/include/gmp.h exists and is specifically a symbolic link, then use test -L instead

marcelklehr:
Now, we are where I've been before. Still all debian i386 builds are failing. :/

norweeg:
@marcelklehr hmm... the build logs indicate it can't find gmp.h when building in i386... wish I could see where it was trying to look for it.  The fpm image is based on debian 'buster', and according to https://packages.debian.org/buster/i386/libgmp-dev/filelist the header file should be at /usr/include/i386-linux-gnu/gmp.h.  Wish I could see what `dpkg-architecture --query DEB_BUILD_MULTIARCH` returns on Debian 'buster' i386.  I suspect it isn't "i386-linux-gnu"

tilosp:
take a look at https://travis-ci.org/nextcloud/docker/jobs/603287089#L4009

docker uses `/bin/sh` to build which doesn't support `[[` `]]`, replacing it with `[` `]` should fix this

norweeg:
@tilosp literally popped up seconds before I clicked "comment" to say the same thing!

marcelklehr:
ooph, fingers crossed. Once the build passes I'll do a rebase to clean things up a bit... :sweat_smile:

tilosp:
It breaks the build of some of the `full` example [dockerfiles](https://github.com/nextcloud/docker/blob/master/.examples/dockerfiles/full/apache/Dockerfile) because they try to install gmp.
Simply removing it from those Dockerfiles should fix it.

marcelklehr:
Hah, if I only I had looked at the examples.

marcelklehr:
@tilosp @J0WI Have you thought about creating a CONTRIBUTING.md file for this repo, where people can learn how to contribute, how things are supposed to work, some basic info about the CI process? That would go a long way, I think :)

tilosp:
yeah having a `CONTRIBUTING.md` to help out new contributors makes a lot of sense.
I am not really into writing docs, and i am also not sure what info would be useful.
Maybe explaining which files to edit and which are auto-generated by `update.sh`.

@marcelklehr what info about the ci would haven been useful to you? Simply explaining which Dockerfiles get build?

btw could you also remove the gmp from the full/fpm-alpine example? It doesn't break the build, but building it twice still wastes resources.

matthiasbaldi:
When will this be released? I will test it then asap :) 👍

tilosp:
Once https://github.com/docker-library/official-images/pull/6867 is reviewed and merged it will be build and published by https://doi-janky.infosiftr.net/.

PR #865, opened 2019-09-17
Change upload_max_filesize

gruessung:
Per default only 2MB are allowed. This is not usable for webclients to add attachments. 
Added 1G as default value.

### DIFF
--- a/16.0/fpm-alpine/Dockerfile
+++ b/16.0/fpm-alpine/Dockerfile
@@ -83,6 +83,11 @@ RUN { \
     \
     echo 'apc.enable_cli=1' >> /usr/local/etc/php/conf.d/docker-php-ext-apcu.ini; \
     \
+    { \
+        echo 'upload_max_filesize = 1G'; \
+        echo 'post_max_size = 1G'; \
+    } > /usr/local/etc/php/conf.d/upload-max-filesize.ini; \
+    \
     echo 'memory_limit=512M' > /usr/local/etc/php/conf.d/memory-limit.ini; \
     \
     mkdir /var/www/data; \


J0WI:
There is no default value that suites all needs. Please use `.htaccess` or `.user.ini` to customize this value: https://docs.nextcloud.com/server/16/admin_manual/configuration_files/big_file_upload_configuration.html#configuring-upload-limits-within-the-gui

kesselb:
Hey @J0WI, would you mind to have a look again? 

nextcloud/server#13990
nextcloud/server#14430

Since Nextcloud 16 php settings are not overwritten via .htaccess or .user.ini. So the current max_upload_size for the docker image is 2MB. Probably something like below should be done for upload_max_filesize and post_max_size. 

@gruessung could you please sign off your commits. Check https://github.com/nextcloud/docker/pull/865/checks?check_run_id=224844488 for details. 1G is much (and higher than the memory limit). Could we pick a lower value? All nextcloud clients use chunking for uploads anyway. We probably need a higher value for something like davfs2.

DudleyDursley:
It might not be possible to have a default value that suites all needs but there could be one that suites most needs and something like 500MB or 1GB would IMHO be more practical than 2MB.

FelixSchwarz:
> There is no default value that suites all needs. Please use `.htaccess` or `.user.ini` to customize this value: …

@J0WI While I agree that is likely hard to find a value that suites all needs I think it should be possible to use Nextcloud with the official client and upload common files (e.g. smartphone images) out-of-the box. As the official clients use chunking (AFAIK) a pretty small value like 15 MB should be enough to achieve that.

Also uploading files is ... somewhat the core of Nextcloud. So there are quite a few users who need higher limits. For example they might be using use 3rd-party applications without chunking. Therefore the container should offer a way to override the setting easily when creating a new container.

> It might not be possible to have a default value that suites all needs but there could be one that suites most needs and something like 500MB or 1GB would IMHO be more practical than 2MB.

@DudleyDursley You don't need to go that high. 15 MB should be enough due to chunking.

ghoebilly:
docker-compose file workaround until this can be some variables with #873 :

```
app:
    image: nextcloud:fpm-alpine
    restart: always
    volumes:
     - type: bind
        source: /_docker-compose-dir_/upload-max-filesize.ini
        target: /usr/local/etc/php/conf.d/upload-max-filesize.ini
```

where `/_docker-compose-dir_/upload-max-filesize.ini` is, of course, an existing file on the filesystem containing the [upload_max_filesize](https://www.php.net/manual/en/ini.core.php#ini.upload-max-filesize) and [post_max_size](https://www.php.net/manual/en/ini.core.php#ini.post-max-size) directives.

For example:

```
upload_max_filesize = 15M
post_max_size = 30M
```

Verify with https://host/ocs/v2.php/apps/serverinfo/api/v1/info?format=json

_Notes: 
Works on 19.x
Fix Nextcloud Deck upload_max_filesize issue when uploading >2M attachments from android app._

J0WI:
Closed by #1379

PR #889, opened 2019-10-05
README: Mention how to enable STARTTLS

dbrgn:
This was non-obvious, I assumed the value would be `starttls`.


tilosp:
A note about starttls was added in #897

PR #891, opened 2019-10-08
Set default value for  trusted proxies

kesselb:
There are some reports at nextcloud/server about broken url generation. This change should improve the situation for docker users. 

Nginx and Traefik forward X-Forwarded-Host and X-Forwarded-Proto by default. Url generation should work out of the box with the ip of the reverse proxy listed in trusted_proxies.

### DIFF
--- a/.config/trusted_proxies.config.php
+++ b/.config/trusted_proxies.config.php
@@ -1,0 +1,7 @@
+<?php
+
+if ($trustedProxies = getenv('TRUSTED_PROXIES')) {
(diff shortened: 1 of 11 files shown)


kesselb:
Hmm. I see we use a different way for trusted domains. Adding the trusted proxies once on setup to config.php would save some getenv calls. Let me know which way do you prefer.

kesselb:
As discussed at https://github.com/nextcloud/docker/pull/819#issuecomment-541362298 its bad.

PR #894, opened 2019-10-10
Readme: improve migration from existing section

NeroBurner:
This PR consists of three parts to improve the migration subsection:

1. add postgresql commands to migration section
in the migration section it is assumed the previous installation was using MySQL. Make the MySQL specific statements explicit and add PostgreSQL specific instructions

2. copy commands copied into existing folders
When the docker-compose up is run the first time the nextcloud
directories are created. When after that the `docker cp` command copies
whole folders they are added as subfolders to the existing folders.

For example, when copying the data folder with
```
docker cp ./data/ nextcloud_app_1:/var/www/html/data
```

afterwards the data folder is in `/var/www/html/data/data`

3. and last mention custom_apps config
I had a hard time figuring out why my apps, which I copied to `custom_apps` folder were not recognized

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,7 +378,22 @@ You're already using Nextcloud and want to switch to docker? Great! Here are som
   -            "writable" => true,
   -    ),
   ```
-  3. Make sure your data directory is set to /var/www/html/data
+  3. Make sure to have the `apps` directory non writable and the `custom_apps` directory writable
+  ```php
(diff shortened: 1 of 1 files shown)


J0WI:
Could you signoff your commits?

NeroBurner:
signed off and rebased on current master

J0WI:
Thanks!

PR #909, opened 2019-10-29
configuration according CVE-2019-11043

nexudx:
suggested configuration from: https://nextcloud.com/blog/urgent-security-issue-in-nginx-php-fpm/


J0WI:
Thanks for the heads up. The examples are outdated anyway, so I'll close this in favor of #912.

PR #921, opened 2019-11-12
Ship 17.0.2 on stable channel

tilosp:
Depends on https://github.com/nextcloud/updater_server/pull/279

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -2,8 +2,8 @@
 set -Eeuo pipefail
 
 declare -A release_channel=(
-	[stable]='16.0.6'
-	[production]='16.0.6'
+	[stable]='17.0.2'
+	[production]='16.0.7'
 )
 
 self="$(basename "$BASH_SOURCE")"


generate-stackbrew-library.sh
```diff
@@ -2,7 +2,7 @@
 set -Eeuo pipefail
 
 declare -A release_channel=(
-	[stable]='16.0.6'
+	[stable]='17.0.1'
```
J0WI:
```suggestion
	[stable]='17.0.2'
```

generate-stackbrew-library.sh
```diff
@@ -2,7 +2,7 @@
...
 	[production]='16.0.6'
```
J0WI:
```suggestion
 	[production]='17.0.2'
```

tilosp:
```suggestion
 	[production]='16.0.7'
```
the production channel is still on 16

tilosp:
Sure

Issue #847, opened 2019-08-28
App container crashing using full + compose examples


mikecai:
Hi, I'm trying to setup my NextCloud using the docker compose example for nginx-proxy/mariadb-cron-redis/fpm.  For the NextCloud image, I'm building my own using the full/fpm-alpine example Dockerfile, because I would like to install LibreOffice.  However, this combination is causing the NextCloud app container to enter a crash loop when I execute `docker-compose up -d`.  The only change I made was to the docker-compose.yml to use my image for web and cron + setting the environment vars.  If I leave the docker-compose.yml as is where it is using the nextcloud:fpm-alpine image, it runs fine.  I've been at this for couple of hours now, any suggestions on what I maybe doing incorrectly would be really appreciated.  Here's what I see in the docker logs for the app container.

```
docker logs -ft nextcloud_app_1 
2019-08-28T01:53:55.609183422Z Configuring Redis as session handler
2019-08-28T01:53:55.839298076Z /usr/lib/python2.7/site-packages/supervisor/options.py:461: UserWarning: Supervisord is running as root and it is searching for its configuration file in default locations (including its current working directory); you probably want to specify a "-c" argument specifying an absolute path to a configuration file for improved security.
2019-08-28T01:53:55.839328773Z   'Supervisord is running as root and it is searching '
2019-08-28T01:53:56.851836938Z Configuring Redis as session handler
2019-08-28T01:53:57.096103131Z /usr/lib/python2.7/site-packages/supervisor/options.py:461: UserWarning: Supervisord is running as root and it is searching for its configuration file in default locations (including its current working directory); you probably want to specify a "-c" argument specifying an absolute path to a configuration file for improved security.
2019-08-28T01:53:57.096131614Z   'Supervisord is running as root and it is searching '
2019-08-28T01:53:57.099474899Z Unlinking stale socket /run/supervisord.sock
2019-08-28T01:53:58.260561818Z Configuring Redis as session handler
2019-08-28T01:53:58.492818686Z /usr/lib/python2.7/site-packages/supervisor/options.py:461: UserWarning: Supervisord is running as root and it is searching for its configuration file in default locations (including its current working directory); you probably want to specify a "-c" argument specifying an absolute path to a configuration file for improved security.
2019-08-28T01:53:58.492849843Z   'Supervisord is running as root and it is searching '
2019-08-28T01:53:58.496005263Z Unlinking stale socket /run/supervisord.sock
[...]

PR #925, opened 2019-11-17
Correct supervisord examples so that the container can start

lazerl0rd:
Inspired by #847, I selected the rootdir as the place for our configuration
since we already chuck a few files there as seen by https://github.com/nextcloud/docker/blob/master/17.0/fpm-alpine/Dockerfile#L122.

Suggested-by: mikecai <mikecai@us.ibm.com>


lazerl0rd:
Apologies for the multiple force-pushes, I had used an old signoff (which used my previous alias/pseudonym) by mistake.

PR #932, opened 2019-11-22
Avoid pecl prompt

aanno:
Building the docker image(s) on my own, I failed with pecl prompting for (interactive) user input. I found the very same problem at https://stackoverflow.com/questions/8141407/install-pecl-modules-without-the-prompts and fixed it.


J0WI:
I cannot reproduce this and so far our CI and builds work without this. How did you build them and on what version/system?

PR #939, opened 2019-12-11
Replace Travis CI with GitHub actions

tilosp:
This way ci should be a lot faster because GitHub actions runs up to 20 concurrent jobs instead of the 5 concurrent by travis ci.
The builds are still failing because of the unstable tests, but that was also happening on travis ci.

This pr makes use of the [generate jobs script](https://github.com/docker-library/bashbrew/blob/master/scripts/github-actions/generate.sh) provided by the docker-library team. This allows for a simpler `update.sh` script.


pierreozoux:
@tilosp instead of requiring PR to run update.sh, why don't we have a GitHub action that does run the script instead?

If you agree, I can PR that soemtimes :)

tilosp:
@pierreozoux Sure go ahead.
I was planning on replaceing @tilosp-bot with GitHub Actions in the future anyway.

But I am not sure how you want to remove the requirement on manually committing the `update.sh` result for each pr, since the the build actions won't work correctly without.

pierreozoux:
Here it is :)

https://github.com/nextcloud/docker/pull/1212

PR #991, opened 2020-02-09
Pass original request protocol to FPM to fix Content Security Policy errors

gpoole:
The Nginx config for the proxied FPM images does not set HTTPS as enabled by default. This caused an issue for me in the client login screen where clicking "Accept" would seem to time out. Chrome shows an error in the console like `Refused to send form data to 'http://cloud.domainname/' because it violates the following Content Security Policy directive: "form-action 'self'"`.

This change will pass the original request protocol via the FastCGI `HTTPS` param, so I think it should cover both HTTP and HTTPS.

The details of the error and suggested fix are described in nextcloud/server#17409.

### DIFF
--- a/.examples/docker-compose/with-nginx-proxy-self-signed-ssl/mariadb/fpm/web/nginx.conf
+++ b/.examples/docker-compose/with-nginx-proxy-self-signed-ssl/mariadb/fpm/web/nginx.conf
@@ -35,6 +35,11 @@ http {
         server app:9000;
     }
 
+    map $http_x_forwarded_proto $pass_https {
+        default off;
+        https on;
+    }
+
     server {
         listen 80;
 
@@ -119,7 +124,7 @@ http {
             include fastcgi_params;
             fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
             fastcgi_param PATH_INFO $path_info;
-            # fastcgi_param HTTPS on;
+            fastcgi_param HTTPS $pass_https;
 
             # Avoid sending the security headers twice
             fastcgi_param modHeadersAvailable true;
--- a/.examples/docker-compose/with-nginx-proxy/mariadb-cron-redis/fpm/web/nginx.conf
+++ b/.examples/docker-compose/with-nginx-proxy/mariadb-cron-redis/fpm/web/nginx.conf
@@ -35,6 +35,11 @@ http {
         server app:9000;
     }
 
+    map $http_x_forwarded_proto $pass_https {
+        default off;
+        https on;
+    }
+
     server {
         listen 80;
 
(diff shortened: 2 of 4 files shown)


J0WI:
See https://github.com/nextcloud/server/issues/17409#issuecomment-538685736

gpoole:
That comment suggests making a manual change to the nexcloud config to force https on. This PR makes manual user configuration unnecessary and works for both HTTP and HTTPS without any manual changes.

-------- Original Message --------

J0WI:
There's [an ENV for it](https://github.com/nextcloud/docker#using-the-apache-image-behind-a-reverse-proxy-and-auto-configure-server-host-and-protocol). We use the nginx config [proposed by upstream](https://docs.nextcloud.com/server/19/admin_manual/installation/nginx.html#nextcloud-in-the-webroot-of-nginx).

gpoole:
In the upstream config you've linked to HTTPS is turned on with `fastcgi_param HTTPS on`, whereas in the examples in this repo, which I've updated in this PR, [it's commented out](https://github.com/nextcloud/docker/blob/6b6c68cabfc745ca94e3fe64145f5e04dffeecf3/.examples/docker-compose/with-nginx-proxy/mariadb/fpm/web/nginx.conf#L122).

J0WI:
That's because the example is without https.

gpoole:
The example I linked to is using the letsencrypt companion to terminate HTTPS, it is using HTTPS.

gpoole:
I'm not saying this was intentional on your part, but as a first time contributor to this project your approach to this PR came across as a bit dismissive. I don't mean to waste your time arguing over trivial changes and I completely understand if you don't want to merge my change - it's your decision - but I put a small amount of time and effort into sharing a change I thought would be helpful with the community, so I would have appreciated a quick note in your closing like "thanks for the suggestion, we don't want to merge the change because we try to keep the configuration as close to the upstream one as possible and this is too different" in your original response. From my perspective it's difficult to know what your motivation for closing the PR is unless you explain it, so it also helps me figure out what to do or not do in the future or maybe learn something useful.

J0WI:
> The example I linked to is using the letsencrypt companion to terminate HTTPS, it is using HTTPS.

> In the upstream config you've linked to HTTPS is turned on with fastcgi_param HTTPS on, whereas in the examples in this repo, which I've updated in this PR, it's commented out.

You're right, something is a bit obscure here. I closed a bunch of issues that seems to be obsolete since #1048 and I'm sorry if I was to fast with this one. I think this change is not required if you're [using the ENV](https://github.com/nextcloud/docker#using-the-apache-image-behind-a-reverse-proxy-and-auto-configure-server-host-and-protocol) to tell Nextcloud to use HTTPS. Maybe `fastcgi_param HTTPS $https if_not_empty;` is more suitable for a generic example.

If I'm wrong or if there are common cases where this config is still useful, I'd like to ask you to open a PR in the [documenation repo](https://github.com/nextcloud/documentation/blob/master/admin_manual/installation/nginx.rst) first.

gpoole:
That's fair enough, I see what you're saying. Thanks for hearing me out and taking the time to respond in detail, I appreciate it.

PR #998, opened 2020-02-14
Precanned config

jarrod-lowe:
Adds a new environment variable - `NEXTCLOUD_CONFIG_BOOTSTRAP_FILE`, to allow the administrator to set arbitrary settings in config.php easily.

Only applies to new installations.


jarrod-lowe:
@tilosp I am getting failures in some checks deep in the image; that don't seem to be related to my change at all. Should I be worrying about them?

kesselb:
I think https://github.com/nextcloud/docker/pull/968 is the way to go. I don't see any advantage by adding such a feature.

PR #1010, opened 2020-02-22
Disable rewriteip for apache image

kesselb:
By default the ip address visible to Nextcloud is `X-Real-IP` if the request is coming from `10.0.0.0/8`, `172.16.0.0/12` or `192.168.0.0/16`. 

That's a problem because Nextcloud will ignore `HTTP_X_FORWARDED_HOST` / `HTTP_X_FORWARDED_PROTO` and people see invalid (absolute) urls.


J0WI:
There is another PR solving a similar issue: #1048
Do you have any preference or are both required?

kesselb:
> There is another PR solving a similar issue: #1048
> Do you have any preference or are both required?

I would like to disable mod rewriteip and not use the overwrite* variables. Without mod rewrite most setups behind nginx and traefik should work by setting trusted proxies out of the box. 

The other PR is not ready from my pov. If you defined OVERWRITEWEBROOT (or any other variable) once it's not possible to unset it again due the config merge Nextcloud usually does. Using the proper default variable (or null) is required to workaround that limitation.

J0WI:
@kesselb this needs rebase

PR #1023, opened 2020-03-09
add support for oauth in docker image

monperrus:
Fix https://github.com/cweiske/grauphel/issues/55

### DIFF
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -37,6 +37,7 @@ RUN set -ex; \
         libzip-dev \
         libwebp-dev \
         libgmp-dev \
+        php-oauth \
     ; \
     \
     debMultiarch="$(dpkg-architecture --query DEB_BUILD_MULTIARCH)"; \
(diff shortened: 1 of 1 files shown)


kesselb:
That looks like the wrong way to add a new php extension. Please check: https://github.com/docker-library/docs/blob/master/php/README.md#how-to-install-more-php-extensions

monperrus:
Thanks a lot for the pointer, fixed.

kesselb:
Still looks wrong. Afaik apt-get is not the right way to install a php package.

monperrus:
thanks for the feedback.

unclear to me whether `oauth` should be installed with `docker-php-ext-enable` or PECL/`docker-php-ext-enable`.

Here is a WIP version full PECL, WDYT?

J0WI:
Since this is not required by default apps I'd prefer moving this to our examples: https://github.com/nextcloud/docker/tree/master/.examples#dockerfiles

monperrus:
We really want to use grauphel but it's more important for us to depend on the official Docker image of Nextcloud.

So if we don't merge this feature, it's basically a NO-GO for grauphel and all plugins depending on oauth.

monperrus:
Here is how to activate oauth for grauphel.

Works for me in a Docker container running Nextcloud 23:

```
apt install libpcre3-dev
pecl install oauth
echo extension=oauth.so > /usr/local/etc/php/conf.d/oauth.ini
service apache2 restart
```

PR #1056, opened 2020-04-03
Fix incorrect nginx configuration

### LINKED ISSUE #1055: Trying to load certain files causes HTTP 302
As only a small number of files (identified by their extensions) are loaded directly or passed to `index.php` there are some apps such as the `firstrunwizard` which can't show files correctly as a `302` error is returned instead of the actual file.

## Steps to reproduce
1. Boot up the `.examples/docker-compose/insecure/mariadb/fpm/docker-compose.yml`-environment
2. If necessary set a password for the administrator
3. If necessary wait for the additional apps to install
4. Notice that the intro video of the `firstrunwizard` doesn't load
5. Copy the address of the video and open it in a new tab
6. Notice that you're redirected to `http://localhost/apps/files/`

## Expected Behavior
The files should load correctly.

## Actual behavior
Some files don't load and refer to `/apps/files/` instead.

## Workaround
I think the source of the problem might be here:
https://github.com/nextcloud/docker/blob/269a08113889884e3d3607df81ca1aba8f4dd408/.examples/docker-compose/with-nginx-proxy/mariadb/fpm/web/nginx.conf#L104-L106
and here:
https://github.com/nextcloud/docker/blob/269a08113889884e3d3607df81ca1aba8f4dd408/.examples/docker-compose/with-nginx-proxy/mariadb/fpm/web/nginx.conf#L167-L171

I found myself able to fix this issue by removing these sections and adding this piece of code instead:

```nginx
        location / {
            try_files $uri /index.php$request_uri @inexistent;
        }

        location @inexistent {
            rewrite ^ /index.php;
        }
```

manuth:
This PR fixes #1055.
Though - for instance `documentserver_community` still isn't working correctly but I assume this is this app's fault.

### DIFF
--- a/.examples/docker-compose/insecure/mariadb-cron-redis/fpm/web/nginx.conf
+++ b/.examples/docker-compose/insecure/mariadb-cron-redis/fpm/web/nginx.conf
@@ -97,6 +97,12 @@ http {
         #pagespeed off;
 
         location / {
+            try_files $uri /index.php$request_uri @nextcloud-fallback;
+            # Optional: Don't log access to other assets
+            access_log off;
+        }
+
+        location @nextcloud-fallback {
             rewrite ^ /index.php;
         }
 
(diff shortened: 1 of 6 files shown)


J0WI:
This example is based on the [official documentation](https://github.com/nextcloud/documentation). Could you please open a PR there?

PR #1073, opened 2020-04-12
Set Cache-Control header

SuperSandro2000:
This change sets the ``Cache-Control``-Header to the value of the [official documentation](https://docs.nextcloud.com/server/18/admin_manual/installation/nginx.html).
I run Nextcloud behind Traefik which can't set those headers properly and I think those changes benefit the default docker image.

I tested this on my private Nextcloud instance where is works flawlessly and reduced the transferred resources from 8.7MB to 52KB.

### DIFF
--- a/16.0/apache/Dockerfile
+++ b/16.0/apache/Dockerfile
@@ -106,7 +106,7 @@ RUN { \
 
 VOLUME /var/www/html
 
-RUN a2enmod rewrite remoteip ;\
+RUN a2enmod headers rewrite remoteip ;\
     {\
      echo RemoteIPHeader X-Real-IP ;\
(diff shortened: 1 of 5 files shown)


kesselb:
Those cache control headers are part of the .htaccess file. I think it's enough to enable mod headers unless .htacces is disabled.

SuperSandro2000:
I did not check the ``.htaccess`` file. This would have saved me a lot of tinkering. I removed the second apache conf again and just enabled the headers module.

kesselb:
Looks good to me ;) I guess those cache control headers are passed on by a reverse proxy.

SuperSandro2000:
> I guess those cache control headers are passed on by a reverse proxy.

I am using Traefik and it passes them trough.

PR #1107, opened 2020-05-05
Read postfix version tags

J0WI:
Restores 'beta', 'alpha' and 'rc' version tags

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -89,7 +89,7 @@ for version in "${versions[@]}"; do
 			versionPostfix="-$( echo "$fullversion_with_extension" | tr '[:upper:]' '[:lower:]' | grep -oE '(beta|rc|alpha)')"
 		fi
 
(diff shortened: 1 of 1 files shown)


J0WI:
@tilosp `nextcloud:beta` and `nextcloud:rc` haven't been updated since https://github.com/nextcloud/docker/commit/b7e9458648f6d8571ff0e3fcf6abf8aadb9dab35. Do we want them again?

tilosp:
I am not sure why I have removed them in the first place.
I see no reason not to have them, but I think your current PR could cause the same tag to be at multiple images. (For example if there are multiple RCs out)

J0WI:
Do we have `alpha` or multiple `beta` releases?

tilosp:
> Do we have `alpha` or multiple `beta` releases?

The were `alpha` releases. For example:  https://github.com/nextcloud/docker/tree/a1ed54243dbcf7e9b7ebdb77fc325a9d8ebdb3a5/16.0-alpha
I don't think that there were ever multiple betas at the same time.

PR #1108, opened 2020-05-05
Upgrade PHP to 7.4 for 19.0

J0WI:


J0WI:
@tilosp should we merge this before the final release of 19.0?

tilosp:
sure, seems fine
but I think 7.4 should be the default version, otherwise nextcloud 20 will be back on 7.3

J0WI:
I just pushed an update

PR #1115, opened 2020-05-09
Traefik 2.0  with sample docker-compose file

ismailyenigul:
Traefik 2.0 is much more native solution for reverse proxy in docker env. 
I created a sample docker-compose.yml file and updated README.


.examples/docker-compose/with-traefik2-postgresql-redis-single-network/README.md
```diff
@@ -0,0 +1,16 @@
+## Trafik Single Network Deployment
+
+1. Create a  network
+
+` # docker network  create  nextcloud `
+
+4. Edit `docker-compose.yml`
```
NeroBurner, line 7: why does it jump from 1. to 4.? And what does "Create a network" mean?

.examples/docker-compose/with-traefik2-postgresql-redis-multi-network/README.md
```diff
@@ -0,0 +1,23 @@
+## Trafik Multi Network Deployment
```
NeroBurner, line 1: what does "Multi Network" mean in this context, and why would I choose it over the single-network?

NeroBurner (commented):
some things unclear for me

.examples/docker-compose/with-traefik2-postgresql-redis-single-network/docker-compose.yml
```diff
@@ -0,0 +1,104 @@
...
+    restart: always
+    networks:
+      - nextcloud
+    depends_on:
+      - redis
+      - db
+    labels:
+      - traefik.http.routers.nextcloud.middlewares=nextcloud,nextcloud_redirect
+      - traefik.http.routers.nextcloud.tls.certresolver=myresolver
+      - traefik.http.routers.nextcloud.rule=Host(`nextcloud.mydomain.com`)
```
meonkeys, line 81: Just a nitpick: `nextcloud.example.com` may be a more appropriate placeholder.

.examples/docker-compose/with-traefik2-postgresql-redis-multi-network/README.md
isdnfan: Example shows 2 networks - one is "traefik" network which used to expose Nextcloud app to reverse proxy and the second one is "Nextcloud" network used for communication between Nextcloud app and DB.

"Multi network" is best practice in docker environment. Basically it means containers belonging to the same application connect to each other using isolated internal network (which is not accessible from outside). This increase security as only container within this network can talk to each other. it's a docker implementation of separated network VLAN.

joshtrichards:
Agreed. The examples maintained here in this repository are meant to give an idea of how these images can be combined with other services. We must take a fairly stringent perspective on limiting the scope of the examples to keep things from getting out of hand (e.g. because we can't realistically maintain, keep up-to-date with, or be experts at every possible service/add-on/configuration).

For Traefik, a possible spot for this documentation would be [contributing to the Nextcloud Server documentation itself](https://github.com/nextcloud/documentation/) and/or [the "how to" / wiki](https://help.nextcloud.com/c/how-to-faqs/9) resources maintained as part of [the help forum](https://help.nextcloud.com/). Another possibility is that someone maintains their own repos/gists of the example(s) of interest to them that they're volunteering to maintain for others. :)

PR #1131, opened 2020-05-28
Remove db prefix option

J0WI:
https://github.com/nextcloud/server/pull/21112
Closes: #1146


J0WI:
> I think we should also remove this:
> 
> https://github.com/nextcloud/docker/blob/0599d1022cc4e1192a19f1c8fcd83c6d195d6b96/.config/autoconfig.php#L26-L28

Fixed and also updated to include #1146.

kesselb:
Looks good to me :+1: Sorry I forgot the test it :see_no_evil:

PR #1167, opened 2020-06-28
update version to latest (18.0.6 and 17.0.7)

doc75:
Update the script to generate official image to latest version as they have been published a while ago.

Despite the tags present on docker hub, the **production** image seems to still be in version 17.0.6 (not checked for 18.0)


doc75:
Not sure why the build are failing. Any hint is welcome so that I can fix it (if possible)

Issue #1165, opened 2020-06-25
New instance's initialization removes everything in /var/www/html


myzhang1029:
# The issue
When I tried to use this this image, mounted /var/www/html with a directory previously used for nextcloud, I ended up realizing the image deleted everything in that directory. I felt lucky that I used a separate data directory for my original nextcloud files.  
This happens to be done at [entrypoint.sh:98](https://github.com/nextcloud/docker/blob/c135f7ef602a2dca6938a8b9aec295c1b6ca66f3/19.0/fpm/entrypoint.sh#L98)

# Steps to reproduce
```
mkdir nextcloud_data
touch nextcloud_data/foo
docker run --rm -itv $PWD/nextcloud_data:/var/www/html nextcloud
[wait until it says "Initializing finished"]
^C
[nextcloud_data/foo is removed]
```

# Possible fix
Maybe moving the original file to a backup directory is better.

doc75:
@myzhang1029 I think this is the expected behaviour. This folder is updated, but you have an exclusion list that prevent `data` directory inside `/var/www/html` to be modified. The fact that your `foo` file get deleted seems normal to me.

myzhang1029:
@doc75 Thanks for your reply. Yes, you're right. After reading the code I found `/upgrade.exclude` which did that job.
However, as this is the _official_ image, I guess there will be other silly people like me who ran into the same issue.
So I'll try to note that in the documentation.

PR #1168, opened 2020-06-28
Add note about not placing files inside /var/www/html

myzhang1029:
Existing files inside /var/www/html get deleted by the upgrade script. This update documents that. (#1165)


README.md
```diff
@@ -92,6 +92,8 @@ $ docker run -d \
 nextcloud
 ```
 
+Please note that anything inside `/var/www/html` will be removed by the installation script, except for `config`, `data`, `custom_apps`, `themes` and `version.php`. So make sure you do not leave other important files inside.
```
J0WI, line 95: Can you just refer to [upgrade.exclude](https://github.com/nextcloud/docker/blob/master/upgrade.exclude)?

README.md
```diff
@@ -479,6 +481,10 @@ You're already using Nextcloud and want to switch to docker? Great! Here are som
   ```php
   'datadirectory' => '/var/www/html/data',
   ```
+  5. Make sure your Nextcloud lives at the web root. Look for lines like
+  ```php
+  'overwrite.cli.url' => 'https://localhost/nextcloud',
+  ```
```
J0WI, line 487: This seems to be unrelated to this PR

README.md
myzhang1029: Yeah, I agree. That would probably be better.

README.md
myzhang1029: What I meant is if the user is migrating an existing instance to Docker, and the old one was not at the web root, the installation script would remove the files. So I guess it is related.

README.md
myzhang1029:
```suggestion
Please note that anything inside `/var/www/html` will be removed by the installation script, except for those listed in [upgrade.exclude](https://github.com/nextcloud/docker/blob/master/upgrade.exclude). So make sure you do not leave other important files inside.
```

README.md
myzhang1029: Please correct me if I'm wrong.

J0WI: IMHO migrating an existing instance to Docker is a different topic.

myzhang1029: I see your point. I created this as a whole just because my files were deleted, and I found those two places to take a note at. Please feel free to separate the changes if you feel that is better.

bsiddans:
I'd love to help get this one over the line - in addition to the issue mentioned above there's also #1528. This looks like a straight-forward solution compared to #1548.

With regard to the proposed addition, would a slight tweak to the wording help make clear that this issue is relevant for both installation, and upgrades? E.g.:

> Users should note that data inside the main folder (/var/www/html) may be removed during installation and upgrades. Users should consider:
> - Mounting storage volumes to locations outside of /var/www/html; and/or
> - Listing files and folders that should not be removed from the main folder in [upgrade.exclude](https://github.com/nextcloud/docker/blob/master/upgrade.exclude).

I'm not sure what the status of this PR is but I'm happy to submit a new one if required.

PR #1181, opened 2020-07-19
Fix triggering install and update script

stackcoder:
The default commands set in Dockerfile, entrypoint.sh and README.me were out of sync. Therefore, the install and update script was not trigger at initial startup unless `NEXTCLOUD_UPDATE=1` were set.

### DIFF
--- a/17.0/apache/entrypoint.sh
+++ b/17.0/apache/entrypoint.sh
@@ -43,13 +43,13 @@ file_env() {
     unset "$fileVar"
 }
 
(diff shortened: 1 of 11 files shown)


tilosp:
@J0WI @stackcoder i don't see how the regex `apache2-foreground` matches something that the regex `apache` couldn’t.

I personally like the more generally approach of using the regex to ignore the apache major version.
But even if we want to match on the exact command it makes no sense to use a regex.

So this should be either reverted or we replace the reqex matching with a simply string comparison like we already do for `php-fpm`.

PR #1183, opened 2020-07-20
Update stable to 18.0.7 and production to 17.0.8

ffried:
With https://github.com/nextcloud/updater_server/commit/0aeb993e31ede6beaa9a8dcf6539e098a3b8af5e version 19.0.1 was raised to stable and production in the updater-server, but that is currently not reflected in the docker tags.
Is this change enough to fix this? Or do I need to do anything else?

**Update (27.07.20)**:
As mentioned by @tilosp, the 17 and 18 tracks only get the next new patch version. This PR now reflects that change and updates 18 to 18.0.7 and 17 to 18.0.8.

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -2,8 +2,8 @@
 set -Eeuo pipefail
 
 declare -A release_channel=(
-	[stable]='18.0.6'
-	[production]='17.0.7'
+	[stable]='18.0.7'
+	[production]='17.0.8'
 )
 
 self="$(basename "$BASH_SOURCE")"


generate-stackbrew-library.sh
```diff
@@ -2,8 +2,8 @@
 set -Eeuo pipefail
 
 declare -A release_channel=(
-	[stable]='18.0.6'
-	[production]='17.0.7'
+	[stable]='19.0.1'
```
tilosp, line 5: the stable channel is still on 18, you can see here that 18 instances only get the `18.0.7` update and not `19.0.1`
https://github.com/nextcloud/updater_server/blob/aad2440bcec4af6d19c3cae5993868dea8d8da64/config/config.php#L403-L405
```suggestion
	[stable]='18.0.7'
```

generate-stackbrew-library.sh
```diff
@@ -2,8 +2,8 @@
...
+	[production]='19.0.1'
```
tilosp, line 6: the production channel is still on 17, you can see here that 17 instances only get the `17.0.8` update and not `19.0.1`
https://github.com/nextcloud/updater_server/blob/aad2440bcec4af6d19c3cae5993868dea8d8da64/config/config.php#L140-L142
```suggestion
	[production]='17.0.8'
```

generate-stackbrew-library.sh
ffried: Ah, my bad, sorry. Mis-interpreted the config file thinking that since 19.0.1 was *added* (as a whole) in the commit I've referenced above, that this now means that instances will be offered the upgrade to 19.

I've now adjusted the versions to 18.0.7 and 17.0.8, respectively.

generate-stackbrew-library.sh
ffried: Fixed (see comment above)

PR #1193, opened 2020-07-29
Adds ability to configure S3 object stores via environment variables

adamjenkins1:
Allows consumers of the Nextcloud docker image to configure S3 compatible object storage as Nextcloud's primary storage through environment variables. Supporting the configuration of this feature through the environment will make the image that much easier to leverage. Ref: https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/primary_storage.html#simple-storage-service-s3

### DIFF
--- a/.config/s3.config.php
+++ b/.config/s3.config.php
@@ -1,0 +1,21 @@
+<?php
+if (getenv('OBJECTSTORE_S3_BUCKET')) {
+  $use_ssl = getenv('OBJECTSTORE_S3_SSL');
+  $use_path = getenv('OBJECTSTORE_S3_USEPATH_STYLE');
+  $CONFIG = array(
+    'objectstore' => array(
+      'class' => '\OC\Files\ObjectStore\S3',
+      'arguments' => array(
+        'bucket' => getenv('OBJECTSTORE_S3_BUCKET'),
+        'key' => getenv('OBJECTSTORE_S3_KEY') ?: '',
+        'secret' => getenv('OBJECTSTORE_S3_SECRET') ?: '',
+        'region' => getenv('OBJECTSTORE_S3_REGION') ?: '',
+        'hostname' => getenv('OBJECTSTORE_S3_HOST') ?: '',
+        'port' => getenv('OBJECTSTORE_S3_PORT') ?: '',
+        'use_ssl' => (strtolower($use_ssl) === 'false' || $use_ssl == false) ? false : true,
+        // required for some non Amazon S3 implementations
+        'use_path_style' => $use_path == true && strtolower($use_path) !== 'false'
+      )
+    )
+  );
+}
--- a/17.0-rc/apache/Dockerfile
+++ b/17.0-rc/apache/Dockerfile
@@ -1,0 +1,151 @@
+# DO NOT EDIT: created by update.sh from Dockerfile-debian.template
+FROM php:7.3-apache-buster
+
+# entrypoint.sh and cron.sh dependencies
+RUN set -ex; \
+    \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+        rsync \
+        bzip2 \
+        busybox-static \
+    ; \
+    rm -rf /var/lib/apt/lists/*; \
+    \
+    mkdir -p /var/spool/cron/crontabs; \
+    echo '*/5 * * * * php -f /var/www/html/cron.php' > /var/spool/cron/crontabs/www-data
+
+# install the PHP extensions we need
+# see https://docs.nextcloud.com/server/stable/admin_manual/installation/source_installation.html
+RUN set -ex; \
+    \
+    savedAptMark="$(apt-mark showmanual)"; \
+    \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+        libcurl4-openssl-dev \
+        libevent-dev \
+        libfreetype6-dev \
+        libicu-dev \
+        libjpeg-dev \
+        libldap2-dev \
+        libmcrypt-dev \
+        libmemcached-dev \
+        libpng-dev \
+        libpq-dev \
+        libxml2-dev \
+        libmagickwand-dev \
+        libzip-dev \
+        libwebp-dev \
+        libgmp-dev \
+    ; \
+    \
+    debMultiarch="$(dpkg-architecture --query DEB_BUILD_MULTIARCH)"; \
+    if [ ! -e /usr/include/gmp.h ]; then ln -s /usr/include/$debMultiarch/gmp.h /usr/include/gmp.h; fi;\
+    docker-php-ext-configure gd --with-freetype-dir=/usr --with-png-dir=/usr --with-jpeg-dir=/usr --with-webp-dir=/usr; \
+    docker-php-ext-configure gmp --with-gmp="/usr/include/$debMultiarch"; \
+    docker-php-ext-configure ldap --with-libdir="lib/$debMultiarch"; \
+    docker-php-ext-install -j "$(nproc)" \
+        exif \
+        gd \
+        intl \
+        ldap \
+        opcache \
+        pcntl \
+        pdo_mysql \
+        pdo_pgsql \
+        zip \
+        gmp \
+    ; \
+    \
+# pecl will claim success even if one install fails, so we need to perform each install separately
+    pecl install APCu-5.1.18; \
+    pecl install memcached-3.1.5; \
+    pecl install redis-4.3.0; \
+    pecl install imagick-3.4.4; \
+    \
+    docker-php-ext-enable \
+        apcu \
+        memcached \
+        redis \
+        imagick \
+    ; \
+    \
+# reset apt-mark's "manual" list so that "purge --auto-remove" will remove all build dependencies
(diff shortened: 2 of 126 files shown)


J0WI:
Nextcloud with S3 has some serious issues:
https://github.com/nextcloud/server/issues/11826
https://github.com/nextcloud/server/issues/14027
https://github.com/nextcloud/server/issues/22077

pierreozoux (approved):

pierreozoux:
Looks good to me and I agree with @adamjenkins1 Encryption is still beta.

pierreozoux:
@J0WI what do we need to merge this?

.config/s3.config.php
```diff
@@ -0,0 +1,20 @@
+<?php
+if (getenv('OBJECTSTORE_S3_HOST')) {
+  $CONFIG = array (
+    'objectstore' => array(
+      'class' => '\\OC\\Files\\ObjectStore\\S3',
+      'arguments' => array(
+        'bucket' => getenv('OBJECTSTORE_S3_BUCKET'),
+        'autocreate' => getenv('OBJECTSTORE_S3_AUTOCREATE') ?: true,
```
pierreozoux, line 8:
```suggestion
        'autocreate' => getenv('OBJECTSTORE_S3_AUTOCREATE') ?: false,
```

pierreozoux: We had troubles with the true as it didn't create the bucket.

I don't recall that well, but I think it is better this way.

.config/s3.config.php
```diff
@@ -0,0 +1,20 @@
+<?php
+if (getenv('OBJECTSTORE_S3_HOST')) {
```
cuppett, line 2: S3_HOST isn't one of the required parameters. Might be better to key off of bucket name here to know we're in this.

.config/s3.config.php
```diff
@@ -0,0 +1,20 @@
...
+        'bucket' => getenv('OBJECTSTORE_S3_BUCKET'),
+        'autocreate' => getenv('OBJECTSTORE_S3_AUTOCREATE') ?: true,
+        'key'    => getenv('OBJECTSTORE_S3_KEY'),
+        'secret' => getenv('OBJECTSTORE_S3_SECRET'),
+        'hostname' => getenv('OBJECTSTORE_S3_HOST'),
+        'port' => getenv('OBJECTSTORE_S3_PORT'),
+        'use_ssl' => getenv('OBJECTSTORE_S3_SSL') ?: true,
+        'region' => getenv('OBJECTSTORE_S3_REGION'),
+        // required for some non Amazon S3 implementations
+        'use_path_style' => getenv('OBJECTSTORE_S3_USEPATH_STYLE') ?: true,
```
cuppett, line 16: I tried something like this and it didn't work with B2 S3 compatibility. Empty/blank doesn't respond well here. I had to have something like this:

`'use_path_style' => getenv('S3_USE_PATH_STYLE') === 'true' ?: false`

But I've also seen this:

`'use_path_style' => (bool) getenv('S3_USE_PATH_STYLE') ?: true`

I don't know if the latter works (haven't tried it).

pierreozoux, line 16:
```suggestion
        'use_path_style' => getenv('S3_USE_PATH_STYLE') === 'true' ?: true,
```

I think I prefer the non amazon deployment ;)

.config/s3.config.php
pierreozoux, line 2:
```suggestion
if (getenv('OBJECTSTORE_S3_BUCKET')) {
```

.config/s3.config.php
cuppett: How does somebody get false here? I think this needs to be:

`'use_path_style' => getenv('OBJECTSTORE_S3_USEPATH_STYLE') === 'true' ?: false`

.config/s3.config.php
adamjenkins1: Requiring `S3_HOST` makes sense to me. How would Nextcloud know how to access the bucket without knowing where it's hosted?

.config/s3.config.php
adamjenkins1: I agree with using the explicit comparison as anything except `false`, `0`, and `null` will be true. If we want `OBJECTSTORE_S3_USEPATH_STYLE` to default to true, we can do this
```php
'use_path_style' => getenv('OBJECTSTORE_S3_USEPATH_STYLE') === 'false' ?: true,
```

.config/s3.config.php
adamjenkins1: Sure, this makes sense. If `S3_HOST` won't be required, should we have a default value?

cuppett: It's automatic/computed for the standard S3 endpoints based on account key credentials and configured/default region of the SDK configuration. 

Ultimately, in the nextcloud docs, its not a required parameter, but this will promote it to required for the container to work. In 19 and earlier, bucket, key & secret are required. In 20 and later, only bucket name is required.

cuppett: https://github.com/nextcloud/server/blob/master/lib/private/Files/ObjectStore/S3ConnectionTrait.php#L63-L80

.config/s3.config.php
cuppett: The server code defaults to "false" when nothing is specified. Here, you will require the container environment to say "false" to get it, else, it defaults to true....

https://github.com/nextcloud/server/blob/master/lib/private/Files/ObjectStore/S3ConnectionTrait.php#L115

If the environment variable is unset, do we want the logic to be flipped from how the base config processing works in the main nextcloud server repository?

.config/s3.config.php
cuppett: Unset/empty appears to work based on the current version of https://github.com/nextcloud/server/blob/master/lib/private/Files/ObjectStore/S3ConnectionTrait.php

adamjenkins1: Ok, makes sense

.config/s3.config.php
adamjenkins1: I think it makes the most sense to remain consistent. I don't feel strongly about this, so I'm fine either way as long as we agree.

cuppett: Here's the one I was able to make work to be consistent with the main project:

https://github.com/cuppett/nextcloud-openshift-s2i/blob/master/autoconfig/objectstorage.config.php

Here's a snippet I used to try and run that down:

https://gist.github.com/cuppett/f25928cf91c7f0d229e515b96efdf4e6

pierreozoux:
@cuppett the ideal would be the php code to read env var.

Do you want to care of that upstream in the serveur code? I Think it is the bamest place.

.config/s3.config.php
```diff
@@ -0,0 +1,20 @@
...
+      'class' => '\\OC\\Files\\ObjectStore\\S3',
+      'arguments' => array(
+        'bucket' => getenv('OBJECTSTORE_S3_BUCKET'),
+        'autocreate' => getenv('OBJECTSTORE_S3_AUTOCREATE') ?: true,
+        'key'    => getenv('OBJECTSTORE_S3_KEY'),
+        'secret' => getenv('OBJECTSTORE_S3_SECRET'),
+        'hostname' => getenv('OBJECTSTORE_S3_HOST'),
+        'port' => getenv('OBJECTSTORE_S3_PORT'),
+        'use_ssl' => getenv('OBJECTSTORE_S3_SSL') ?: true,
+        'region' => getenv('OBJECTSTORE_S3_REGION'),
```
cuppett, line 14: I'm doing some testing of this. If we do this here like this it ends up in the final config.php like this:

```
      'hostname' => false,
      'port' => false,
      'region' => false,
```
The app seems to handle it okay (probably because it matches up okay in a weak null check).

cuppett:
I've done testing with various versions of this file. I was able to emulate the default behaviors of the param file wrt set/unset environment variables and trues/falses here:

https://github.com/cuppett/nextcloud-openshift-s2i/blob/master/autoconfig/objectstorage.config.php

A container with this available is here: quay.io/cuppett/nextcloud:latest  (uses Apache, Nextcloud 19 and /var/www/html as the storage location)

Can we catch the PR code and doc stanza up to this?

pierreozoux:
@cuppett I think you are the most knowlegble about this, feel free to update things :) Or if you prefer create a new PR if it is faster, I'll just trust you and merge it :)

cuppett:
I've opened a PR for @adamjenkins1 (https://github.com/adamjenkins1/docker/pull/1/commits).

If he merges it pulls in a rebase & has all these updates. Should then show here, we can merge and closes the issues!

cuppett:
@pierreozoux We have it here now.

pierreozoux:
Hum.. actually there are now 2 files objectstore and s3.
Plus I'd prefer to either have one commit, or 2 separate commit with the run separated.

As you want, if you don't mind, I'll redo the PR and put your names there. I'll take the objectstore file then.

pierreozoux:
@adamjenkins1 sorry, it is just there was too many files in your commit, difficult to review, and not in sync with master.

I created a new PR with the code, and merged it.

Thanks @adamjenkins1 and @cuppett for your help on getting this done :)

i put you as coauthor, I hope that's fine :)

PR #1213, opened 2020-08-25
Remove unecessary logic.

pierreozoux:
I did wrote this some years ago, when app update logic was not that great,
I believe that it is now resolved.


docker-entrypoint.sh
```diff
@@ -86,10 +86,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
 
     if version_greater "$image_version" "$installed_version"; then
         echo "Initializing nextcloud $image_version ..."
-        if [ "$installed_version" != "0.0.0.0" ]; then
-            echo "Upgrading nextcloud from $installed_version ..."
```
SnowMB, line 90: I would keep this line for the log.

PR #1219, opened 2020-08-27
Upgrade 19.0.2 (correct process?)

mbecca:
Hi, I don't know if this is the correct process to update Nextcloud version to 19.0.2. Sorry if not.

I modified latest.txt (19.0.2) and run update.sh locally and then push to my fork.

Please close this PR if it is not correct.

Greetings, Matías


tilosp:
You only need to run `update.sh` and push the changes, the script should update `latest.txt`.
But there is no need to do this manually since it automated.

The `19.0.2` image should be available now, it was a bit delayed because the tars were signed with the wrong key.

Issue #627, opened 2019-01-28
crontab question


hm400:
Im trying to use cron instead of ajax for background jobs. 
From docs.nextcloud.com  im trying to run 

#crontab -u www-data-l
*/5  *  *  *  * php -f /var/www/nextcloud/cron.php

but crontab is not found, how do i add this entry without crontab?
using docker image 15.0.2

SnowMB:
The crontab is already included in the image. It uses the default 15 minute interval from the nextcloud documentation.

https://github.com/nextcloud/docker/blob/8d133d39e602151973fbe5c52defc588b58cc369/Dockerfile-debian.template#L15

If you want to use another setting you should make a derived image and build it on your own.

If you are fine with the default you have two possibilities to use that:

1. Use a second container that runs the `cron` command. You can find an example here: https://github.com/nextcloud/docker/blob/master/.examples/docker-compose/insecure/mariadb-cron-redis/apache/docker-compose.yml

2. Build an derived image that uses `supervisord` or another process manager to run both processes (webserver / cron) in one container. We also have an example for that: https://github.com/nextcloud/docker/tree/8d133d39e602151973fbe5c52defc588b58cc369/.examples/dockerfiles/cron/apache

J0WI:
3. Use a local cronjob/timer and `docker exec -u www-data my-nextcloud php cron.php`

hm400:
> 1. Use a local cronjob/timer and `docker exec -u www-data my-nextcloud php cron.php`

This worked and changed Ajax to Cron in admin->basic settings

hm400:
ty for replies

Iceman0073:
Dear SnowMB,

sorry I'm also new here and I want to ask you something about cron and the external cron container. I selected the example with the second container "cron" but I don't know the right setting to choose within nextcloud. Is it the right way to select Webcron or Cron within the Nextcloud Settings?

Many Thanks for your help.

huuub:
> The crontab is already included in the image. It uses the default 15 minute interval from the nextcloud documentation.
> 
> https://github.com/nextcloud/docker/blob/8d133d39e602151973fbe5c52defc588b58cc369/Dockerfile-debian.template#L15
> 
> If you want to use another setting you should make a derived image and build it on your own.
> 
> If you are fine with the default you have two possibilities to use that:
> 
[...]

PR #1242, opened 2020-09-09
Starting cron without new docker-container

AndreKoepke:
Nextxloud should maintain the cron-service without multiple docker-containers which mounts the same directory.

See this issue as reference: https://github.com/nextcloud/docker/issues/627

### DIFF
--- a/19.0/apache/entrypoint.sh
+++ b/19.0/apache/entrypoint.sh
@@ -185,4 +185,10 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     fi
 fi
 
+if [ "${START_CRON}" = "true" ]; then
+  echo "Starting CRON."
+  sh /cron.sh &
+fi
+
+
 exec "$@"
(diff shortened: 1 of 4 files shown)


AndreKoepke:
I'm not sure why the checks failed. I assumed that I not changed anything for older versions and why their postgres got trouble.

J0WI:
You can simply use the `--entrypoint` flag: https://github.com/nextcloud/docker/blob/d1d4467e34df2b1d39c75f02c9b1f7e6f872d76a/.examples/docker-compose/insecure/mariadb-cron-redis/apache/docker-compose.yml#L35-L40

AndreKoepke:
@J0WI sadly not. I had to start two containers (both with the same mount point).
But one container is enough.

Iam using nextcloud in a kubernetes environment. And I don't want to start an additional container only for cron.

J0WI:
Then you should look for https://github.com/nextcloud/docker/tree/master/.examples/dockerfiles/cron

AndreKoepke:
Thanks, but there is the same problem: I have to start a second container with the same mount point. 

And the second container needs a whole PHP environment.

That's the reason why I want to run the cron-job parallel in the same container. ^^

AndreKoepke:
Ah, my bad.

Nice, that's exactly what I am looking for.

PR #1243, opened 2020-09-09
Adds ability to configure swift object store though environment variables

adamjenkins1:
Allows consumers of the Nextcloud docker image to configure OpenStack Swift object storage as Nextcloud's primary storage through environment variables. Supporting the configuration of this feature through the environment will make the image that much easier to leverage. Ref: https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/primary_storage.html#openstack-swift

### DIFF
--- a/.config/swift.config.php
+++ b/.config/swift.config.php
@@ -1,0 +1,31 @@
+<?php
+if (getenv('OBJECTSTORE_SWIFT_URL')) {
+    $autocreate = getenv('OBJECTSTORE_SWIFT_AUTOCREATE');
+  $CONFIG = array(
+    'objectstore' => [
+      'class' => 'OC\\Files\\ObjectStore\\Swift',
+      'arguments' => [
+        'autocreate' => $autocreate == true && strtolower($autocreate) !== 'false',
+        'user' => [
+          'name' => getenv('OBJECTSTORE_SWIFT_USER_NAME'),
+          'password' => getenv('OBJECTSTORE_SWIFT_USER_PASSWORD'),
+          'domain' => [
+            'name' => (getenv('OBJECTSTORE_SWIFT_USER_DOMAIN')) ?: 'Default',
+          ],
+        ],
+        'scope' => [
+          'project' => [
+            'name' => getenv('OBJECTSTORE_SWIFT_PROJECT_NAME'),
+            'domain' => [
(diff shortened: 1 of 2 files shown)


pierreozoux:
In other project, I have the feeling that you can configure switch with S3 vars. Why is it not the case with Nc?

Then Im fine with this PR thanks!

.config/swift.config.php
```diff
@@ -0,0 +1,30 @@
+<?php
+if (getenv('OBJECTSTORE_SWIFT_URL')) {
+  $CONFIG = array(
+    'objectstore' => [
+      'class' => 'OC\\Files\\ObjectStore\\Swift',
+      'arguments' => [
+        'autocreate' => getenv('OBJECTSTORE_SWIFT_AUTOCREATE') ? true : false,
```
cuppett, line 7: The default in the server project is false[0]. Can we make this work like what we tested out in s3.config.php?

$autocreate = getenv('OBJECTSTORE_SWIFT_AUTOCREATE')

'use_path_style' => $autocreate == true && strtolower($autocreate) !== 'false'

[0]: https://github.com/nextcloud/server/blob/7e6f04490e286450d009c156f08de6a22f0f653d/lib/private/Files/ObjectStore/SwiftFactory.php#L62-L67

cuppett (commented):
I'm not knowledgable on how Swift works.

Based on the doc, this seems to correspond to the v3 authentication mechanism only and may not work for v2 as well. How prevalent is the v2 mechanism still?

.config/swift.config.php
adamjenkins1: Sure, will do

.config/swift.config.php
```diff
@@ -0,0 +1,31 @@
+<?php
+if (getenv('OBJECTSTORE_SWIFT_URL')) {
+    $autocreate = getenv('OBJECTSTORE_SWIFT_AUTOCREATE');
```
cuppett, line 3: Nit. The indent here is a little different than the rest of the file.

cuppett (approved):
Thanks for the update on autocreate!

pierreozoux (approved):

pierreozoux:
Thanks for yiur contribution :)

PR #1254, opened 2020-10-01
Update docker-entrypoint.sh

artonge:
Config files do not seems to get updated during a nextcloud upgrade.

Adding this command should correctly update them.

Instead of adding this cmd we could replace `/config/` by `/config/config.php` in `upgrade.exclude`.

Or maybe I am doing something wrong ?


M4lik:
As far as I can see, with this extra line, seperate config files would be deleted. So if I created the file `config/my-custom-s3.config.php` the new script would delete that, and I would have lost my config and possibly broke my instance. 

I am no expert with NC, but I don't really see the necessity fo such a function. Non-existing config parameters are filled with default values by NC, so you don't need to have new config file versions. If you want to have them, you would need to get your hands dirty anyway, I think...

artonge:
I have updated the extra line so extra files are not deleted.

The function is useful as those config files bring new features for the docker images. For old installation wanting to use those features, currently they have to update them by hand.

Issue #1088, opened 2020-04-23
NEXTCLOUD_ADMIN_PASSWORD can't have spaces


jdorel:
When setting the env var NEXTCLOUD_ADMIN_PASSWORD with a password containing spaces, I get the following debug message : 
```
/entrypoint.sh: 30: local: straggler": bad variable name
```
This is annoying because I'd like to use a passphrase.

doc75:
Not sure what I propose will work, but looking at the entrypoint.sh file, you might be able to create a file containing your password containing spaces.
Reference the path to your file with the variable `NEXTCLOUD_ADMIN_PASSWORD_FILE`, the path being a path inside the docker container of course.

But having a variable value with space is always a pain in scripts language. Perhaps the easier way is to remove space in your passphrase. I am using passphrase too, but without space in it.

Hoe this helps.

invidian:
I also hit this, when trying to deploy using https://github.com/nextcloud/helm with the following values.yaml file:
```
nextcloud:
  password: "test test test"
```

nicobo:
I confirm the same with the latest stable image (19.0.3) ; here is the trace with `set -vx` (line 32 displayed here is actually 30 because I added 2 lines for debug) :

```
nextcloud_nextcloud.1.hl349n0m6tri@machine7    | + echo Initializing nextcloud 19.0.3.1 ...
nextcloud_nextcloud.1.hl349n0m6tri@machine7    | + [ 0.0.0.0 != 0.0.0.0 ]
nextcloud_nextcloud.1.hl349n0m6tri@machine7    | + echo Initializing finished
nextcloud_nextcloud.1.hl349n0m6tri@machine7    | + [ 0.0.0.0 = 0.0.0.0 ]
nextcloud_nextcloud.1.hl349n0m6tri@machine7    | + echo New nextcloud instance
nextcloud_nextcloud.1.hl349n0m6tri@machine7    | + file_env NEXTCLOUD_ADMIN_PASSWORD
nextcloud_nextcloud.1.hl349n0m6tri@machine7    | + set -vx
nextcloud_nextcloud.1.hl349n0m6tri@machine7    | + local var=NEXTCLOUD_ADMIN_PASSWORD
nextcloud_nextcloud.1.hl349n0m6tri@machine7    | + local fileVar=NEXTCLOUD_ADMIN_PASSWORD_FILE
nextcloud_nextcloud.1.hl349n0m6tri@machine7    | + local def=
[...]

PR #1257, opened 2020-10-04
Fixes "NEXTCLOUD_ADMIN_PASSWORD can't have spaces

nicobo:
Fixes "NEXTCLOUD_ADMIN_PASSWORD can't have spaces"

By double-quoting the buggy lines...

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -27,8 +27,8 @@ file_env() {
     local var="$1"
     local fileVar="${var}_FILE"
     local def="${2:-}"
-    local varValue=$(env | grep -E "^${var}=" | sed -E -e "s/^${var}=//")
-    local fileVarValue=$(env | grep -E "^${fileVar}=" | sed -E -e "s/^${fileVar}=//")
+    local varValue="`env | grep -E \"^${var}=\" | sed -E -e \"s/^${var}=//\"`"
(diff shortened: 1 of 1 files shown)


docker-entrypoint.sh
```diff
@@ -27,8 +27,8 @@ file_env() {
     local var="$1"
     local fileVar="${var}_FILE"
     local def="${2:-}"
-    local varValue=$(env | grep -E "^${var}=" | sed -E -e "s/^${var}=//")
-    local fileVarValue=$(env | grep -E "^${fileVar}=" | sed -E -e "s/^${fileVar}=//")
+    local varValue="`env | grep -E \"^${var}=\" | sed -E -e \"s/^${var}=//\"`"
+    local fileVarValue="`env | grep -E \"^${fileVar}=\" | sed -E -e \"s/^${fileVar}=//\"`"
```
NeroBurner, line 31: why replace the newer `$()` syntax with the (what I believe) older and less readable \` syntax?

nicobo: @NeroBurner I just tried to replace this at first to see if it was causing the strange behavior. You can probably put them back. 
BTW I find backquotes way more readable :wink:

joshtrichards:
Closing since #1088 isn't reproducible. If someone has a Compose/run command that reproduces this behavior today, please either open a new issue or respond on #1088.

PR #1259, opened 2020-10-05
Install apps via env var for initial setup

mgerhardy:


andrewgosselin-nt:
I am currently running this in my entrypoint, I hope this gets pushed soon as without this feature the auto deployment of apps is hard.

docker-entrypoint.sh
```diff
@@ -168,6 +168,12 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
                             NC_TRUSTED_DOMAIN_IDX=$(($NC_TRUSTED_DOMAIN_IDX+1))
                         done
                     fi
+                    if [ -n "${NEXTCLOUD_INSTALL_APPS+x}" ]; then
+                        echo "Install apps…"
+                        echo "$NEXTCLOUD_INSTALL_APPS" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//' | tr " " \\n | while read APP; do
```
J0WI, line 173: This breaks with multiple spaces. I guess a comma separated list would be much easier

AbcSxyZ: Will you consider PR https://github.com/nextcloud/docker/pull/1260 ? Would be great to have a more flexible way to configure Nextcloud for the automatic installation.

At Nextcloud (you're working there ?), did you think about enabling all options for `autoconfig.php` ? Would it be possible to enable more configuration variables there ?

It can be a great thing to automate the entire installation !

Loxeno: @AbcSxyZ I agree with you, custom script provided by Docker Config or Docker Volume will give more possibility like enabling encryption right after installation for example or even after upgrade process.

@J0WI any thoughts ?

J0WI: Yes, I think I'll prefer #1260

PR #1264, opened 2020-10-08
Update nginx.conf

cristi-bourceanu:
nginx times out when sign up is done, displaying the message "504 Gateway Timeout". This happens for the ssl one as well. I found that setting fastcgi timeout overcomes this issue.


kernitus:
Ah thank you for this, I thought I was going insane since I changed nothing in my config and it suddenly didn't work anymore!

J0WI:
This example is taken from the [upstream documentation](https://github.com/nextcloud/documentation/blob/master/admin_manual/installation/nginx.rst). Please file a PR there if the problem still occurs.

PR #1268, opened 2020-10-10
Trusted domains

julien235:
Add a call to file_env to export NEXTCLOUD_TRUSTED_DOMAINS variable environment.
Execute update.sh to update every versions.


J0WI:
`file_env` is for parsing docker secrets. The `NEXTCLOUD_TRUSTED_DOMAINS` env works as documented for new installations, for existing instances you may need to update the `config.php` files.

Issue #1087, opened 2020-04-23
Set correct nextcloud permissions for local storage access


33Fraise33:
I'm running Nextcloud in Docker through Ansible: 

```
- name: setup nextcloud docker container
  docker_container:
    name: nextcloud
    image: nextcloud:latest
    networks:
      - name: fraise_network
        ipv4_address: "{{ docker_network.nextcloud }}"
    purge_networks: yes
    volumes:
      - "/data/nextcloud/www:/var/www/html"
      - "/data:/data"
    user: 33:1003
    env:
      NEXTCLOUD_ADMIN_USER: gianni
      NEXTCLOUD_ADMIN_PASSWORD: "{{ nextcloud_admin_password }}"
      NEXTCLOUD_TRUSTED_DOMAINS: "files.frai.se"
      MYSQL_DATABASE: nextcloud
      MYSQL_USER: nextcloud
      MYSQL_PASSWORD: "{{ db_nextcloud_password }}"
      MYSQL_HOST: "{{ docker_network.nextcloud_db }}"
    restart: yes
    restart_policy: always
    state: started
    pull: true
```

I would like to be able to write into my /data directory through the external storage access. 

A group was created on the Docker host with the name 'data' where I added in http and changed the folder permissions of all folders to `775` and owner: `root:data`. 

When adding the external storage I see "You don't have permission to upload or create files here". 
After giving a certain folder temporary 777 permissios I see that nextcloud still writes as http:http. 

Is it possible to change this when running in docker?

flower1024:
i use this dockerfile since a few months

FROM nextcloud:fpm-alpine

ARG UID
ARG GID

RUN apk --no-cache add shadow && \
    rm -rf /var/cache/apk/* && \
    OUID=$(id -u www-data) && \
    OGID=$(id -g www-data) && \
    usermod -u $UID www-data && \
    groupmod -g $GID www-data && \
    find / -group $OGID -exec chgrp -h www-data {} \; && \
    find / -user $OUID -exec chown -h www-data {} \;

Oberfeldwedler:
THis is what i am running.
```
FROM nextcloud:20-fpm-alpine

ARG UID=1003
ARG GID=1003

RUN apk --no-cache add shadow && \
rm -rf /var/cache/apk/* && \
OUID=$(id -u www-data) && \
OGID=$(id -g www-data) && \
usermod -u $UID www-data && \
groupmod -g $GID www-data ;

[...]

Issue #1081, opened 2020-04-18
Make UserID and Group ID of www-data match with host


uchagani:
I'm running Ubuntu 19.10 as my host OS.

I've ran into an issue where the user id and the group id of the `www-data` user/group are different on my host and in the nextcloud containers.  This makes it difficult to move the data directory.

In order to get this to work I had to do the following:

1.  Get UID/GID for the www-data user in the nextcloud docker container
2.  Create data dir on the host
3.  Create a user/group on the host with the UID/GID of the www-data user from the nextcloud docker container (82:82 in my case)
4.  Change ownership of the data directory to the user/group created in step 3.
5.  recreate the docker-containers

J0WI:
There's no need to create a user on the host. You can `chown 82:82` on the host, `chown www-data:www-data` in the container (using `docker exec`) or `docker run --user` to run the container as custom user.

rgarrigue:
Why don't you add a UID & GID environment variable defaulting to 33 (default for www-data)  to be able to specify the container's data UID & GID ? 

It would be way more convenient to integrate with other containers downloading files (like boypt/cloud-torrent for example)

BlauerPulli:
@J0WI depending on the host setup running the whole container as non-root can lead to other problems (e.g docker socket is not able to bind port 80 / other permission problems). I think it would be a nice feature (besides it seems to be a good practice in other docker projects) to especially set the uid/gid for the user who writes data.

@rgarrigue but this was discussed in two other threads:
https://github.com/nextcloud/docker/issues/359
https://github.com/nextcloud/docker/issues/659

and if you need this feature urgently i think you have to implement it.. :smile:

PR #1278, opened 2020-10-17
Add env variables to change UID/GID of www-data

### LINKED ISSUE #359: How to change UID/GID in NextCloud container?

jan-di:
This PR adds the support for two environment variables that can change the UID/PID of the www-data user dynamically at the entrypoint script. This allows to use a specific uid/gid that is also used on the host to prevent permission issues on bind mounted volumes.

To stay backwards-compatible, the environment variables have the current UID/GID as default value, so they are 3 possible use cases:
- No UID/GID defined, no --user parameter given: Image starts as root, apache/php uses default UID/GID (as before)
- No UID/GID defined, --user parameter given: Image starts as $user, apache/php uses defaut UID/GID (as before)
- UID/GID defined, no --user parameter given: Image starts as root, apache/php uses defined UID/GID (new)

An additional advantage of changing the UID/GID of www-data instead of creating a new user is, that no config files have to be changed for this - every service can use www-data as before.

This should fix #359, #1081, #1087. See also https://github.com/nextcloud/docker/pull/772#issuecomment-710705819 for more info.


Ornias1993:
I think this is a rather clean solution to the issues at hand, nice work! :)

jan-di:
Hello @tilosp, i have rebased this PR to the latest changes. Some tests are failing due to database tests, but it seems rather random to me. Can I request a review or or is there something other I should do before?

uchagani:
@tilosp can we get this merged please?

20.0/fpm-alpine/entrypoint.sh
```diff
@@ -43,6 +43,22 @@ file_env() {
     unset "$fileVar"
 }
 
+# change uid/gid of www-data
+if [ "$(id -u)" = 0 ]; then
+    CURRENT_UID=$(id -u www-data)
+    CURRENT_GID=$(id -g www-data)
+    if [ -n "${UID+x}" ] && [ "$UID" -ne $CURRENT_UID ]; then
+        echo "Change UID of www-data from $CURRENT_UID to $UID"
+        usermod -u $UID www-data
```
Kami, line 52: Just a heads up - this won't work since alpine doesn't  ship with``usermod`` command by default.

So either ``shadow`` apk package needs to be installed or ``/etc/passwd`` edited using sed or similar (not really robust approach).

```bash
/entrypoint.sh: line 52: usermod: not found
/entrypoint.sh: line 52: usermod: not found
```

jan-di: Yeah, good point. I've tested it only with the debian variant. I would like to use the shadow package because it seems to be the most stable solution. I added a new commit with shadow and now it works on both variants. Thanks.

J0WI:
I'd prefer to use the same approach as in https://github.com/docker-library/wordpress/pull/249 (see also https://github.com/docker-library/wordpress/pull/351 and https://github.com/docker-library/wordpress/pull/371)

Ornias1993:
@J0WI The problem with the Wordpress (and LinuxServer) design, is the fact is isn't compatible with the docker "user" setting. For those two it's acceptable, because the settings is relatively new but for new implementations, like this one, not being able to use docker best-practices is a no-go.

J0WI:
@Ornias1993 do you refer to something else than `docker run --user`?

olivergg:
An alternative solution (or dirty hack ? :sweat_smile: ) would be to use bindfs (on Linux) to mirror the directory on your host and set the permission accordingly, then just use this directory instead of the original one in your docker volume

For example, if you have a MyFiles directory with the UID/GID 1000:1000 then you can create a mirror with the UID/GID changed to 33:33 (www-data by default).

```shell
# bindfs -u 33 -g 33 MyFiles MyFilesMirrored
```

See http://manpages.org/bindfs

eathtespagheti:
Hoping to see this implemented ASAP

Fel-o:
any updates?

djak250:
Is this still on people's radar?

tushev:
I ended up with building my own, custom Docker image, with
```
RUN addgroup --gid 1000 --system nextcloud
RUN adduser --uid 1000 --system --disabled-login --disabled-password --gid 1000 nextcloud
```

And my `docker-compose.yml` contains
```
  apache:
    build:
      context: ./images
      dockerfile: Dockerfile-nextcloud-apache
    user: '1000:1000'
```

bllngr:
@Kami @tilosp @J0WI, it seems like this is a clean solution to a frequently requested feature and base for other functionality. It also seems to align with Docker best-practices. But there hasn't been any feedback for one year. Is there anything that speaks against rebasing and merging it?

vincentDcmps:
> I'd prefer to use the same approach as in [docker-library/wordpress#249](https://github.com/docker-library/wordpress/pull/249) (see also [docker-library/wordpress#351](https://github.com/docker-library/wordpress/pull/351) and [docker-library/wordpress#371](https://github.com/docker-library/wordpress/pull/371))

in case of you use ldap for connection wordpress approach doesn't work because tou UID is not in /etc/passwd

skjnldsv:
Hey! There is something that bother me in this pr, it's the `chown` and `chgrp`.
This could lead to multiple issues and I don't think that is the job of a docker image.
The sysadmin should know and manually do those changes.

As a side note:
- [ ] Please fix tests
- [ ] Please only edit the root templates `docker-entrypoint.sh`. The changes are done automatically

skjnldsv (changes requested):
Hey! There is something that bother me in this pr, it's the `chown` and `chgrp`.
This could lead to multiple issues and I don't think that is the job of a docker image.
The sysadmin should know and manually do those changes.

As a side note:
- [ ] Please fix tests
- [ ] Please only edit the root templates `docker-entrypoint.sh`. The changes are done automatically
- [ ] Could you explain the requirements for the `shadow` dependency on alpine ? @jan-di

Ornias1993:
> Hey! There is something that bother me in this pr, it's the `chown` and `chgrp`. This could lead to multiple issues and I don't think that is the job of a docker image. The sysadmin should know and manually do those changes.
> 
> As a side note:
> 
> * [ ]  Please fix tests
> * [ ]  Please only edit the root templates `docker-entrypoint.sh`. The changes are done automatically
> * [ ]  Could you explain the requirements for the `shadow` dependency on alpine ? @jan-di


PUID/UID GID/PGID is quite globally used (and not a very good idea considering systems like kubernetes and docker already implement more secure user-setting, for example the docker `--user` flag or kubernetes `runAsUser`.). It's goal is 50% setting the user/group and 50% setting the file permisisons.

However, the design here is actually troublesome:
`No UID/GID defined, --user parameter given: Image starts as $user, apache/php uses default UID/GID (as before)`

When `--user` is set, everything should run as that user. Not run as an arbitrary other/default UID/GID. Regardless of what the UID/GID settings are.

Remember UID/GID env-vars are not a standard! Setting the `--user` or `runAsUser` is. So the image should focuss on those users, rather than users using UID/GID.

Issue #1275, opened 2020-10-15
Correct examples - fix known problems


Lejo1:
The first time to install nextcloud on docker using docker-compose and ran into multiple problems.

So let's have a look at the [maxi-install](https://github.com/nextcloud/docker/tree/master/.examples/docker-compose/with-nginx-proxy/mariadb-cron-redis/fpm) example with nginx + letsencrypt, fpm, mariadb, redis, cron

So let's look at the problems a new noob like me had:

1. #927 as this is a letsencrypt setup the :$server_port is not needed
2. #819 just add OVERWRITEPROTOCOL=https same here it's a letsencrpyt setup so it's always https
3. [redis needs a password](https://help.nextcloud.com/t/solved-latest-docker-image-broke-the-installation-redis-password-auth/87598/2) marked/mentioned
4.  #1041 should be at least mentioned if not fixed
5. I didn't knew what installation to chose, there should be some more explanation on the setups.

A docker-compose file should be a working easy install solution. We don't want everybody searching again for already known problems.

etique57:
Thanks, I just ran into 5, 3 and 4 respectively. For 4, I'd still prefer to use the OnlyOffice separate docker but cannot seem to make it work.

Lejo1:
> I'd still prefer to use the OnlyOffice separate docker but cannot seem to make it work.

Same here, there should also be an example for that as well.

cacack:
I'll chime in here that I've been running NextCloud in a docker-compose setup based on the [with-nginx-proxy/mariadb-cron-redis/fpm](/nextcloud/docker/tree/master/.examples/docker-compose/with-nginx-proxy/mariadb-cron-redis/fpm) arrangement for a while and have managed to keep it running over the years.  However, some recent changes with the mobile app authentication (nextcloud/android#4786) have led me to revisit my compose in an attempt to fix -- and now I'm completely hosed with HSTS warning.

It would be _very_ helpful if the examples were solid so that I could use them as a basis for my setup.  As they currently stand, I'm going to need to invest some serious time to dig out of this mess.

Polyterative:
I tried for hours and was unable to create a usable environment...

joshtrichards:
Most of the above Issues in the overview have been addressed various ways. The examples have also been streamlined. There is still work to be done, but there are various Issues tracking existing known bugs as well as enhancement plans/ideas for the remaining examples.

[...]

Issue #1270, opened 2020-10-12
Login Screen Loop 


MartinEnders:
Hi,

when I was setting the redis connection parameters in docker-compose file it resulted in a login-loop.
When I logged in I have been redirected to the login-page immediatly.
Setup: everything in docker, nginx reverse proxy in front of apache-20 version of nextcloud

When I removed the redis-config from the environment it worked fine (s. link below).
I assume that redis is working because it is inserted in the config.php file.

https://help.nextcloud.com/t/login-loop-after-updating-dockerized-nextcloud-to-version-19/88705/3

Now my nextcloud setup works fine.

Best regards
Martin

J0WI:
fixed in #1232

briancsimonsen:
This issue is not fixed. After pulling the latest stable version I was still able to recreate the issue.
The issue #1232 is not the same problem because that issue was caused by leaving the password blank. This problem here is caused by having the password filled in and by doing so, causing a loop.

Shurov:
I faced the same problem, however I have a workaround and a hint to possible problem.
- So, after clean install via docker and creating admin user - login screen was looping. This was done in a web-browser on a PC, pointing to the local installation (http://192.168.0.100:8080)
- I configured a reverse proxy (nginx) with LetsEncrypt certificate for it, being accessible at https://nextcloud.example.com. And it works! (this is a workaround part)
- I tried to configure official Android app to contact my server via a domain name.... And the app also has a login loop, however, it also shows an error `Strict mode: no http connection allowed`. (this is a hint part). But I'm not sure how to fix it. I'd like to move all the installation onto the HTTPS - even local access, but I can't find any manual how to do it using (official) docker install

Rebanar:
I found this issue since Shurov's comment has the issue I was searching for regarding the android app. I have a similar setup using caddy to make everything go through https.

I was able to solve the issue by adding 'overwriteprotocol' => 'https' to my config.php file. This forces it to use https since it doesn't actually know it's being set to https via the reverse proxy.

I don't know if this has anything to do with the original issue, but since I found my way here through a web search I figured someone else might too.

cyberjacob:
That workaround isn't ideal, as it would allow any connections not being made through https for whatever reason. If you're using that I would suggest having your reverse proxy only accept HTTPS and redirect everything attempting to use HTTP.

[...]

Issue #1179, opened 2020-07-16
With redis env config for docker need to check for false


mnestor:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Use redis without a password
2. Errror loading web. Also in occ `RedisException: ERR AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct? in /var/www/html/lib/private/RedisFactory.php:94`

### Expected behaviour
Tell us what should happen
Site should load and occ should not error

### Actual behaviour
Tell us what happens instead
Errors!

### Server configuration

**Operating system:**

**Web server:**

**Database:**

**PHP version:**

**Nextcloud version:** (see Nextcloud admin page)
19.0.1

**Updated from an older Nextcloud/ownCloud or fresh install:**
Docker image update

**Where did you install Nextcloud from:**
Docker

**Signing status:**
<details>
<summary>Signing status</summary>

```
Login as admin user into your Nextcloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results here.
```
</details>

**List of activated apps:**
<details>
<summary>App list</summary>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```
</details>

**Nextcloud configuration:**
<details>
<summary>Config report</summary>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder

or 

Insert your config.php content here. 
Make sure to remove all sensitive content such as passwords. (e.g. database password, passwordsalt, secret, smtp password, …)
```
</details>

**Are you using external storage, if yes which one:** local/smb/sftp/...

**Are you using encryption:** yes/no

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

#### LDAP configuration (delete this part if not used)
<details>
<summary>LDAP config</summary>

```
With access to your command line run e.g.:
sudo -u www-data php occ ldap:show-config
from within your Nextcloud installation folder

Without access to your command line download the data/owncloud.db to your local
computer or access your SQL server remotely and run the select query:
[...]

PR #1232, opened 2020-09-01
fix: set REDIS_HOST_PASSWORD to empty string when not set

stefanprobst:
not setting `REDIS_HOST_PASSWORD` currently leads to a "RedisException: ERR AUTH <password> called without any password configured for the default user." error. this seems to happen because `getenv` returns `false`, but the nextcloud server only checks `isset` and for an empty string (see [here](https://github.com/nextcloud/server/blob/master/lib/private/RedisFactory.php#L93)). casting to a string should hopefully fix this???

fixes #1179

PR #1285, opened 2020-10-25
Add redis password to fix AUTH exception

RosesTheN00b:
In the current version (nextcloud 19.04 and redis 6.08), a redis password is required for authentication.
This pull request sets a standard password.

Should affect the following bugs:

* #1275
* #1270
* #1179 

It works again for me and on my colleague's computer after the fix.


J0WI:
fixed in #1232

PR #1286, opened 2020-10-25
Set dbindex to separate data in a single Redis instance by environment variable

f466162:
Hi,

I want to share a single Redis instance and use it with several independent applications. Thus, I need a way to specify the dbindex. As I'm using a Docker deployment, I'd like to add the dbindex using an environment variable. For my Nextcloud 19 (i.e. "production") installation, I tested the following way. As I cannot test for different Nextcloud versions, I made this change only in the Nextcloud 19 folder. Perhaps the if-statement can be reused for different Nextcloud versions and can be copy/pasted.

Regards
Fabian

### DIFF
--- a/19.0/apache/config/redis.config.php
+++ b/19.0/apache/config/redis.config.php
@@ -14,4 +14,8 @@ if (getenv('REDIS_HOST')) {
   } elseif (getenv('REDIS_HOST')[0] != '/') {
     $CONFIG['redis']['port'] = 6379;
   }
+  
+  if (getenv('REDIS_DB_INDEX') !== false) {
+    $CONFIG['redis']['dbindex'] = getenv('REDIS_DB_INDEX');
+  }
 }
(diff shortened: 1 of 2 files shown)


stefan04:
Hello,
I have the same requirement.
But setting the dbindex in the nextcloud config is not enough by itself. You also need to update the dbindex in the entrypoint.sh for the /usr/local/etc/php/conf.d/redis-session.ini file.

Thanks a lot!

Regards
Stefan

efectn:
Any updates? I think this would be pretty useful :/

nderambure:
It seems a better approach to this is to have many Redis instances.
Correct me if I'm wrong, but Redis in mono-threaded so having many databases in same instance will create a bottleneck in performances.
You can easily create many instances on different ports : 
https://medium.com/@starikovs/run-multiple-redis-instances-on-a-linux-server-systemd-approach-a7d1c75482bd
and then set correct config on each nextcloud instance.

PR #1300, opened 2020-11-06
Use iconv from alpine edge repo to fix plugins breaking

### LINKED ISSUE #1299: iconv does not work properly on fpm-alpine
There currently seems to be an issue with iconv on alpine linux which means that specific plugins and other NC functionalities break if iconv is needed. This happens for example with the mail app and is noted in this issue: https://github.com/nextcloud/mail/issues/3999

There seems to be apossible workaround as linked to the corresponding StackOverflow link: https://stackoverflow.com/questions/63746822/why-does-iconv-returns-empty-string-in-php7-4-fpm-alpine-docker/63747282#63747282
However I'm not sure if this is desirable to be implemented in this Dockerfile

TheTimeWalker:
With this PR, gnu-libiconv and php7-iconv will be installed from the Alpine Linux Edge Community Repository instead of using the official repos. The reason for this is that the current iconv from the official repo doesn't work properly with iconv.

While yes, this is essentially a band aid fix, there seems to be no incentive for this to be fixed in Alpine directly anytime soon. So I think it's justified that we have to work around that as there are apps like the Mail app (v1.6.0 and up) that fully break if this iconv version is not installed.

Closes #1299

### DIFF
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -34,6 +34,11 @@ RUN set -ex; \
         gmp-dev \
     ; \
     \
+    apk add --update-cache --repository http://dl-cdn.alpinelinux.org/alpine/edge/community/ --allow-untrusted \
+	    gnu-libiconv \
+	    php7-iconv \
+    ; \
+    \
     docker-php-ext-configure gd --with-freetype --with-jpeg --with-webp; \
     docker-php-ext-configure ldap; \
     docker-php-ext-install -j "$(nproc)" \
@@ -95,6 +100,7 @@ VOLUME /var/www/html
 %%VARIANT_EXTRAS%%
 
 ENV NEXTCLOUD_VERSION %%VERSION%%
+ENV LD_PRELOAD=/usr/lib/preloadable_libiconv.so
 
 RUN set -ex; \
     apk add --no-cache --virtual .fetch-deps \
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -32,10 +32,7 @@ RUN set -ex; \
         imagemagick-dev \
         libwebp-dev \
         gmp-dev \
-    ; \
-    \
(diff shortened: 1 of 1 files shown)


J0WI:
Pulling packages from `edge` is risky because they might not be binary compatible with 3.12 and can break at any time. Pulling php packages from Alpine doesn't work either because we use php from the Docker base image.

TheTimeWalker:
Agreed that it's risky. Right now with these lines I was able to fix broken iconv for the Mail plugin, which fails loading mails because of that issue.

The only other options are either to find a possibility to pinpoint an iconv version to fetch so that it wouldn't suddenly update at some point and be incompatible or somehow fetch the source at a specific version and let it compile. I don't know if both options are possible though, I haven't checked yet.

Worst case we'll have to discourage others to use the alpine image as any plugin that uses iconv has the possibility to break and be non-functional.

jkellerer:
I've applied the same changes without adding edge repo and it also fixes iconv. Adding edge repo is not required but the env var must be set (also in FPM) and the additional packages must be installed.

AlterDepp:
Perhaps this fixes https://github.com/nextcloud/mail/issues/4351 (nextcloud mail broken)

HannesJo0139:
> Perhaps this fixes [nextcloud/mail#4351](https://github.com/nextcloud/mail/issues/4351) (nextcloud mail broken)

https://github.com/nextcloud/mail/issues/4351 seems to be fixed with alpine 3.13 anyway

J0WI:
Closing in favor of #1378

AlterDepp:
Problem still exists with dockerhub:nextloud 20.0.7-fpm-alpine with php:7.4-fpm-alpine3.13

AlterDepp:
Fixed for me with nextcloud 20.0.8-fpm-alpine and mail app 1.9.1
Thanks

PR #1304, opened 2020-11-13
Bail early on illegal upgrade, fixes #1129

### LINKED ISSUE #1129: Upgrade attempt over 2 major versions leads to hardly recoverable state
Although the documentation states, that upgrading must happen to the immediate successor version, trying to do so inadvertently leads to a non usable system afterwards.

As I see, in the entrypoint you check, wether we downgrade, or not. If I then return to the correct version, everything works.

In the other case, however, it seems the new nextcloud files are already rsynced, when the _Exception: Updates between multiple major versions and downgrades are unsupported.
app_1    | Update failed_ log appears.

Then trying to restart with the correct old version container makes it think a downgrade is in progress. Then we have an unusable system.

It would be just consequent, to cut off an upgrade like this as early as possible.

jeppetrost:
Added a helper function to check whether the upgrade is doable: returns true if major version (installed) + 1 is -lt major version (image).
This lets you upgrade from eg. 10.X.Y -> 10.X+1.Z and 10.X.Y -> 11.A.B, but not 10.X.Y -> 12.A.B.
If it detects this, the script will notify and bail immediately.
Fixes #1129

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -6,6 +6,11 @@ version_greater() {
     [ "$(printf '%s\n' "$@" | sort -t '.' -n -k1,1 -k2,2 -k3,3 -k4,4 | head -n 1)" != "$1" ]
 }
 
+# illegal_upgrade will notify if the major versions are too far apart
+illegal_upgrade() {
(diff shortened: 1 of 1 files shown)


wetneb:
Sad that this was not merged! It would have saved me quite some trouble. I have opened a new version of this PR here: #2018.

PR #1316, opened 2020-12-09
Environment variables in docker-compose section

tetebueno:
Docker-compose file in the example was missing database configuration environment variables of the app container, making the stack to use SQLite by default instead of the DB container.

Didn't add `NEXTCLOUD_ADMIN_*` variables since this addresses the DB being configured with SQLite. This still works setting up admin credentials on first NC run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,9 +241,14 @@ services:
       - 8080:80
     links:
       - db
+    restart: always
     volumes:
       - nextcloud:/var/www/html
-    restart: always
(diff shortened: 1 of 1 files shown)


J0WI (commented):
Could you also add it to the FPM snipped below?
Please make sure that the DCO check passes.

J0WI (approved):

J0WI:
Thanks!

PR #1324, opened 2020-12-14
replace  [ -n "${MY_VAR+x}" ]  in IF with  [[ -v MY_VAR ]]

KnallbertLp:


kesselb:
CI don't like it.

J0WI:
`[[` is bashism and not POSIX compliant

PR #1351, opened 2021-01-01
Add PHP bz2 extension

javabean:
Some Nextcloud apps require the bz2 PHP extension. This pull request adds bz2 into the official Docker image.

### DIFF
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -32,6 +32,7 @@ RUN set -ex; \
         imagemagick-dev \
         libwebp-dev \
         gmp-dev \
+        bzip2-dev \
     ; \
     \
     docker-php-ext-configure gd --with-freetype --with-jpeg --with-webp; \
@@ -48,6 +49,7 @@ RUN set -ex; \
         pdo_pgsql \
         zip \
         gmp \
+        bz2 \
     ; \
     \
 # pecl will claim success even if one install fails, so we need to perform each install separately
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -37,6 +37,7 @@ RUN set -ex; \
         libzip-dev \
         libwebp-dev \
         libgmp-dev \
+        libbz2-dev \
     ; \
     \
     debMultiarch="$(dpkg-architecture --query DEB_BUILD_MULTIARCH)"; \
@@ -56,6 +57,7 @@ RUN set -ex; \
         pdo_pgsql \
         zip \
         gmp \
+        bz2 \
     ; \
     \
 # pecl will claim success even if one install fails, so we need to perform each install separately


J0WI:
We can't support every app: https://github.com/nextcloud/docker/tree/master/.examples#dockerfiles

javabean:
That is very true, we can’t support every app!
But this `bz2` module is a bit special, since it is recommended by the [installation manual](https://docs.nextcloud.com/server/20/admin_manual/installation/source_installation.html#prerequisites-for-manual-installation), we would expect it to be included in the official Docker image. Heck, even `bcmath`, `gmp` and `exif` are there, and they are only in the optional section! ;-)

TL;DR: `bz2` is part of the Nextcloud recommended PHP modules; omitting it in the official Docker image seems like an oversight.

I promise I won’t come back and ask for LibreOffice to be included! :-p

J0WI:
Do you have an example for an affected app?

javabean:
The [Facerecognition](https://apps.nextcloud.com/apps/facerecognition) is one.
And consistency between Nextcloud documentation and the official Docker image is also a must (see my previous comment)!

J0WI:
This app also requires PDlib, so you'd need to customize the Dockerfile anyway.

javabean:
From your comments, I wonder if I got it all wrong since the begining. From the [documentation](https://docs.nextcloud.com/server/20/admin_manual/installation/source_installation.html#prerequisites-for-manual-installation):
> Recommended packages:
> PHP module bz2 (recommended, required for extraction of apps)

If we don’t need bz2 to extract apps, then maybe I should contribute a patch on the documentation instead.

What is the official status on `bz2` PHP extension: required for apps extraction or not? If it is, I believe it should be included. If it is not, that I shall close this PR and submit a documentation patch instead.
Sorry if I sound confused (I am!), will you please enlighten me? :-)

J0WI:
> What is the official status on `bz2` PHP extension: required for apps extraction or not?

I don't know. Maybe someone from @nextcloud/appstore could comment on the current status of `bz2` apps.

BernhardPosselt:
Only gz is supported https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/api/v1/release/parser.py#L82

javabean:
> Only gz is supported

Thanks for your answer. Can you confirm then that the documentation needs to remove reference to bz2 PHP extension?
(Just double-checking here to avoid mistakes!)

BernhardPosselt:
No, can't confirm. You'd need to confirm that with people from the server repo. They might be using the extension for something else.

javabean:
@BernhardPosselt @J0WI do you know any contact @nextcloud/server that could enlighten us as to whether the `bz2` PHP extension is needed for them?

J0WI:
Maybe @nextcloud/documentation? I couldn't find any other usage of bz2 in the server repo.

markuman:
https://github.com/nextcloud/server/blob/f552f23e437371b1a34c7df0fc7a11004f4cc1c2/lib/private/Archive/TAR.php#L56

J0WI:
> Switch to bz2 (or better zstd) will reduce your egress traffic.

The Docker image [uses tar.bz2](https://github.com/nextcloud/docker/blob/59391c541b8c37b1af48f807edfa3961b84a200a/Dockerfile-alpine.template#L111-L112) but everything else needs to be done in `nextcloud/server`.

javabean:
@J0WI So shouldn’t we enable `bz2` PHP module, since Nextcloud server has a code path that expects it to be available?

drsassafras:
bz2 is used extensively and is becoming more and more common. It should not be a question of if one of the best and most common ways to compress data should be supported.

kale1d0code:
Can the error messages for bz2 and imagick be removed?
I am using the official image, if the official image doesn't deem then necessary, it shouldn't be an error message.

h0lley:
> Can the error messages for bz2 and imagick be removed? I am using the official image, if the official image doesn't deem then necessary, it shouldn't be an error message.

it already is. read [here](https://github.com/nextcloud/docker/issues/2118#issuecomment-1924205125).

joshtrichards:
Closing. No longer applicable.

PR #1363, opened 2021-01-09
Add smbclient for External Storages support in Docker containers

cybik:
Quick PR that adds smbclient to the alpine and debian containers so that admins using the Docker containers can, with the right container configuration, use smbclient to mount external storages into Nextcloud without having to reconfigure their containers automatically.


J0WI:
Duplicate of #22. Please see https://github.com/nextcloud/docker/tree/master/.examples#php-module-smbclient

PR #1364, opened 2021-01-11
Enable session locking for the redis session handler and use more sane locking configs.

mgriego:
By default, the Redis session handler does *not* lock sessions while manipulating them.  This can be extremely problematic for applications that might fire multiple requests at the server simultaneously (ie background XHR tasks, images or other assets that are served via PHP, etc) as the session data is written to the handler as a single serialized value, and lack of locking can lead to race conditions where old data might be written back to a session.  I have seen this firsthand with the Talk/Spreed app while using this Docker image along with Redis.  When switching back and forth between conversations, you can sometimes be redirected to the spreed not-found page.  This is due to the multiple XHR requests being run/cancelled/etc, sometimes leading to newer room data in the session being overwritten by older data.  This update fixes the Redis session integration to ensure that sessions are properly locked so that data corruption does not occur.  With these settings, I no longer run into issues with Talk.

I am working on a separate PR for the phpredis project to try to implement some better default locking logic and fix a bug when it can't acquire the lock.  As it stands right now, if you enable locking, it will only attempt to acquire the lock for 20ms, then it will continue - just without the lock.  The session data will be read and provided to the running application.  When the application then attempts to write/close the session, phpredis will fail spectacularly if it failed to acquire the lock during session initialization.  Setting the retries to -1 allows phpredis to try forever to obtain the session lock, which is exactly the same behavior as the default `files` session handler.

References:
https://github.com/nextcloud/spreed/issues/4670
https://github.com/nextcloud/server/pull/1936#issuecomment-282008712
https://github.com/phpredis/phpredis/issues/1411
https://github.com/phpredis/phpredis/issues/1422

### DIFF
--- a/19.0/apache/entrypoint.sh
+++ b/19.0/apache/entrypoint.sh
@@ -69,6 +69,11 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
             else
                 echo "session.save_path = \"tcp://${REDIS_HOST}:${REDIS_HOST_PORT:=6379}\""
             fi
+            echo "redis.session.locking_enabled = 1"
+            echo "redis.session.lock_retries = -1"
+            # redis.session.lock_wait_time is specified in microseconds.
+            # Wait 10ms before retrying the lock rather than the default 2ms.
+            echo "redis.session.lock_wait_time = 10000"
         } > /usr/local/etc/php/conf.d/redis-session.ini
     fi
 
--- a/19.0/fpm-alpine/entrypoint.sh
+++ b/19.0/fpm-alpine/entrypoint.sh
@@ -69,6 +69,11 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
             else
                 echo "session.save_path = \"tcp://${REDIS_HOST}:${REDIS_HOST_PORT:=6379}\""
             fi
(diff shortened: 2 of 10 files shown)


mgriego:
@J0WI Can you look at this?

mgriego:
Can someone take a look at the CI tests?  It looks likes the failures were unrelated to this PR.

J0WI:
This is not something that only affects Docker instances. Could you file an issue or PR (maye [here](https://github.com/nextcloud/documentation/blob/master/admin_manual/configuration_server/caching_configuration.rst#redis)) in the documentation repo? I'd like to hear some more feedback for this change.

raimund-schluessler:
I can confirm that adding the changes from this PR fixed this issue https://github.com/nextcloud/spreed/issues/5148 with Talk for me. I would like to see this fix merged. 👍

mgriego:
@J0WI I have submitted a PR as requested to the documentation repo.  This PR still needs to be considered, though.  There has been lots more discussion on how this issue is affecting Talk users in the linked issues.

mgriego:
Rebased to fix the merge conflicts from when 21.0-beta was moved to 21.0.

mgriego:
@J0WI The documentation PR was merged pretty quickly without any discussion - and is already live: https://docs.nextcloud.com/server/21/admin_manual/configuration_server/caching_configuration.html#id2

Is there anything else that needs to be done with this PR before it can be approved/merged?

J0WI (approved):
LGTM

PR #1378, opened 2021-01-21
Alpine 3.13

### LINKED ISSUE #1299: iconv does not work properly on fpm-alpine

J0WI:
Closes #1299 #1300

### DIFF
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -1,4 +1,4 @@
-FROM php:%%PHP_VERSION%%-%%VARIANT%%3.12
+FROM php:%%PHP_VERSION%%-%%VARIANT%%3.13
 
 # entrypoint.sh and cron.sh dependencies
 RUN set -ex; \


AlterDepp:
Perhaps this fixes https://github.com/nextcloud/mail/issues/4351 (nextcloud mail broken)

J0WI:
Yes it will once [this](https://github.com/docker-library/official-images/blob/master/library/nextcloud) is updated.

snevas:
> Perhaps this fixes [nextcloud/mail#4351](https://github.com/nextcloud/mail/issues/4351) (nextcloud mail broken)

I tested today, and sadly it doesn't.

AlterDepp:
> > Perhaps this fixes [nextcloud/mail#4351](https://github.com/nextcloud/mail/issues/4351) (nextcloud mail broken)
> 
> I tested today, and sadly it doesn't.

same for me. Tested with dockerhub:nextloud 20.0.7-fpm-alpine with php:7.4-fpm-alpine3.13

AlterDepp:
Fixed for me with nextcloud 20.0.8-fpm-alpine and mail app 1.9.1
Thanks

PR #1383, opened 2021-01-24
Clearer migration documentation

Schroedingers-Cat:
This PR improves the readability of the migration section by properly indenting the ordered list. It also adds the `--preserve` command when copying the nextcloud data to the named volume in order to preserve metadata like modification date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -515,7 +515,7 @@ You're already using Nextcloud and want to switch to docker? Great! Here are som
         ```
 4. Copy your data (nextcloud_app_1 is the name of your Nextcloud container):
     ```console
-    docker cp ./data/ nextcloud_app_1:/var/www/html/
+    docker cp --preserve ./data/ nextcloud_app_1:/var/www/html/
     docker-compose exec app chown -R www-data:www-data /var/www/html/data
     docker cp ./theming/ nextcloud_app_1:/var/www/html/
     docker-compose exec app chown -R www-data:www-data /var/www/html/theming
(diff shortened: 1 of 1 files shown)


Schroedingers-Cat:
@tilosp It looks like the failing tests aren't related to this PR and can be ignored.

README.md
```diff
@@ -462,73 +462,71 @@ You're already using Nextcloud and want to switch to docker? Great! Here are som
...
-```console
-docker cp ./data/ nextcloud_app_1:/var/www/html/
-docker-compose exec app chown -R www-data:www-data /var/www/html/data
-docker cp ./theming/ nextcloud_app_1:/var/www/html/
-docker-compose exec app chown -R www-data:www-data /var/www/html/theming
-docker cp ./config/config.php nextcloud_app_1:/var/www/html/config
-docker-compose exec app chown -R www-data:www-data /var/www/html/config
-```
+    ```console
+    docker cp --preserve ./data/ nextcloud_app_1:/var/www/html/
```
tilosp, line 518: I was unable to find docs for this flag . Shouldn't it be documented here? https://docs.docker.com/engine/reference/commandline/cp/

Schroedingers-Cat: My bad. Because docker cp doesn't support copying with preserving modification date, I copied the data with the system's cp into the named volume, which supports the flag. When I prepared this PR from my notes of the migration process, I overlooked that the original doc suggested to use docker cp.  

0d4148f9a490aa3177cb00d1037c4c868e00f64b rewrites this as an alternative suggestion, making it clearer that this doesn't involve docker's cp.

Schroedingers-Cat: Alternatively, it could be something like this, making source and destination clearer:
```bash
cp --preserve --recursive /pathToOldNextcloud/data/ /pathToNextcloudDockerVolume/data
```

Schroedingers-Cat: @tilosp do you prefer the version from 0d4148f9a490aa3177cb00d1037c4c868e00f64b or should I give it another spin with the suggestion from https://github.com/nextcloud/docker/pull/1383#discussion_r563952856?

tilosp (approved):

tilosp:
Thanks for your contribution!

PR #1393, opened 2021-02-03
Update stable to 20.0.7

Crocmagnon:
Do I need to do something else to generate 20.0.7 images and push them to docker hub ? If no, is there some place I can check the progress ? :)

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -Eeuo pipefail
 
-stable_channel='20.0.6'
+stable_channel='20.0.7'
 
 self="$(basename "$BASH_SOURCE")"
 cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"


LedzZm:
I guess it needs to pass the tests, also waiting for that myself.

Crocmagnon:
I fixed the DCO check, but for the other failing I don't know what to do 😕

LedzZm:
I tried to pr a signed commit as well before this started running and still got same (and additional) errors. 

Could it be that there is some more general issue with the version or that we should run some update script?

nshtg:
Why is nextcloud incapable of pushing their newest releases to official docker in a timely manner? Almost 3 days have passed ?!

Crocmagnon:
Yeah that’s why I searched for how to do this here 😕
It’s like they don’t care about delivering their versions to docker users 😕

b3nis:
Relax. I am sure they care - but it might take a few days.

Crocmagnon:
Haha sorry for the harsh tone ☺️ I'm relaxed! English is not my main language.

I'll try to rephrase: It makes me feel like docker is a second-class citizen in the landscape of Nextcloud deployments methods.

b3nis:
> Haha sorry for the harsh tone relaxed I'm relaxed! English is not my main language.
> 
> I'll try to rephrase: It makes me feel like docker is a second-class citizen in the landscape of Nextcloud deployments methods.

Maybe it is :-) I dunno what is the most common "way" of hosting NC in general, but maybe not Docker? I mean, I can understand if Nextcloud is focusing on the platform used by their enterprise customers.

nshtg:
> Maybe it is :-) I dunno what is the most common "way" of hosting NC in general, but maybe not Docker? I mean, I can understand if Nextcloud is focusing on the platform used by their enterprise customers.

Far more than 10M+ pulls imply a huge amount of instances using docker.

And "focusing" on their enterprise customers may be the case, but in my opinion that is no excuse to not keep their docker images up to date in a timely manner - updating it literally costs **less than 5 minutes** !!!

Crocmagnon:
> updating it literally costs less than 5 minutes

It could (should?) even be automated and cost nothing :)

RomainTT:
What I don’t get is that errors come from other versions, not 20.0.7. And still it blocks the PR.

LedzZm:
Is it possible to rerun the deployment? 

The errors that appeared on the two consecutive runs were very inconsistent and the builds had an 1 day gap, with no actual changes in the codebase i believe. 

So maybe i was thinking there might have been an issue with the pipeline.

ChristophWurst:
> I'll try to rephrase: It makes me feel like docker is a second-class citizen in the landscape of Nextcloud deployments methods.

FWIW it has shown to work most reliably, hence https://docs.nextcloud.com/server/latest/admin_manual/installation/source_installation.html describes a classic LAMP setup.

While I agree that a delay on this deployment method is very unfortunate I don't see in which why the server team is blocking this? What's different to the 20.0.6? The change set is minimal: https://nextcloud.com/changelog/#latest20.

ChristophWurst:
I've restarted the checks. Let's see :v:

LedzZm:
The nextcloud-cli-mysql service seems to be the issue. And I see it failing on tests performed for other, older versions, unless i misunderstand something.

lenke182:
Someone could add more debug statements to the test scripts.

The relevant test scripts contains `$RANDOM` and `$$` as code, which both behave explicitly or implicitly random. These influence the database name, username and password as well as the Nextcloud admin username and password. Could this be the reason for the inconsistent results?

[1] https://github.com/docker-library/official-images/blob/master/test/tests/nextcloud-cli-mysql/run.sh
[2] https://github.com/docker-library/official-images/blob/master/test/tests/nextcloud-cli-postgres/run.sh

J0WI:
Sorry for the delay, I'll take care in a minute.

The issue with the stable tag and why it's not automated is mentioned here: https://github.com/nextcloud/updater_server/issues/68. TL;DR we can easily monitor and automate releases and tags, but not channels like stable or production.
For the test failures https://github.com/nextcloud/server/pull/16708 has been opened, but there seems to be another race condition in `occ maintenance:install`. Any help to debug the failures is highly appreciated.

J0WI (approved):

nRaecheR:
While the 20.0.7-* tags are online now, the 20 tag is still missing on docker hub.

nijel:
It seems that amd64 images are actually missing...

marliyev:
20.0.7 still missing, tried just now

nRaecheR:
It's there but not for the AMD64 arch it seems.

marliyev:
yep, tag there but image missing :)

PR #1412, opened 2021-02-19
RFC: drop prereleases

J0WI:
I suggest to drop prereleases from the Docker library.

Reasons:
- they're not for production usage
- it's hard to keep up with the rapid release cycle of prereleases
- tags stopped working (#1107)

cc @nextcloud/docker

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -83,21 +83,6 @@ function version_greater_or_equal() {
 	[[ "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1" || "$1" == "$2" ]];
 }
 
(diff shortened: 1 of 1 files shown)


tilosp (approved):
Sure I don't think they are getting much usage anyway, unfortunately we don't have per tag download stats to verify.

If there was demand for them an alternative would be to publish them outside of the "Official Images" program (for example the github container registry), this would make it easier to keep up with the rapid release cycle.

So removing them seems fine. And if it turns out that there is demand they can always be brought back in some other form that works better.

J0WI:
Thanks for your feedback!

Issue #1414, opened 2021-02-20
SVG imagemagick Nextcloud 21.0 Docker 


sparklyballs:
With the recent update to the vanilla library Nextcloud docker image on x86_64, i.e docker pull nextcloud,   bumping the version of nextcloud to 21.0, some warnings appear,of which all but one were trivial to solve.

Whilst I know that the issue of svg compatibility and imagemagick has been raised several times, and at risk of this being marked as duplicate etc and closed , I feel now is the time to actually address it by either adding whatever is required to the image or removing this new warning.

![Screenshot from 2021-02-20 11-02-20](https://user-images.githubusercontent.com/4343808/108593374-2a1a6580-736b-11eb-8243-745560681bea.png)

And I understand that current advice is to modify a local Dockerfile to add the required packages etc, this isn't always an easy solution, and that (at least for now) there is a specific warning appearing about this, to suggest that people effectively create their own image locally seems a little silly.

cformal:
Agreed. Also, there's a warning regarding the default phone region code.

sparklyballs:
> Agreed. Also, there's a warning regarding the default phone region code.

https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#user-experience

cformal:
> > Agreed. Also, there's a warning regarding the default phone region code.
> 
> https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#user-experience

Yes, of course. But I thought it's a good idea to provide an ENV VAR for docker users. I really don't want to modify php files while using docker.

Sveeeeeen:
> > > Agreed. Also, there's a warning regarding the default phone region code.
> > 
> > 
> > https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#user-experience
> 
> Yes, of course. But I thought it's a good idea to provide an ENV VAR for docker users. I really don't want to modify php files while using docker.

Your config file is in your volume/mount.

cformal:
> > > > Agreed. Also, there's a warning regarding the default phone region code.
> > > 
> > > 
> > > https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#user-experience
> > 
> > 
> > Yes, of course. But I thought it's a good idea to provide an ENV VAR for docker users. I really don't want to modify php files while using docker.
> 
[...]

PR #1381, opened 2021-01-23
Adds imagemagick to alpine image.

pierreozoux:
Fixes  #1105

PR #1424, opened 2021-02-25
WIP: Adding missing dependency for SVG files

LouisVallat:
Possible fix for #1414 with the addition of a missing package.
This only fixes the issue for Debian based images.
Tested with NC21 apache variant.

TODO: alpine variants

### DIFF
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -85,6 +85,7 @@ RUN set -ex; \
         | xargs -rt apt-mark manual; \
     \
     apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false; \
+    apt-get install -y --no-install-recommends libmagickcore-6.q16-6-extra; \
     rm -rf /var/lib/apt/lists/*
 
 # set recommended PHP.ini settings


J0WI:
See also https://github.com/nextcloud/docker/pull/1381#issuecomment-785899020

LouisVallat:
> See also [#1381 (comment)](https://github.com/nextcloud/docker/pull/1381#issuecomment-785899020)

Okay so should I close this PR? What could we do about it?
If I understand correctly, including imagick isn't a good idea, but now it's showing a warning in the NC overview page.

J0WI:
https://github.com/nextcloud/docker/issues/594#issuecomment-785891835
> AFAIK the imagemagick binary is only used to convert SVG favicons (using only default configs). Do I miss something?

I don't really understand why Nextcloud displays a warning for this, but I'm glad fore more feedback.

LouisVallat:
image

That's the exact warning that it now shows in NC21. I've never had this warning show up before, and have always been using the `nextcloud:latest` image. This warning is annoying, and the only known way to fix it is to add the package mentioned in issue #1414. I'd love to have an alternative or a way to fix this warning.

J0WI:
Closing due #1414

PR #1425, opened 2021-02-25
Update nginx.conf

Markus-Beier:
Solving issue: Your web server is not properly set up to resolve "/.well-known/webfinger"
Referencing: https://docs.nextcloud.com/server/latest/admin_manual/installation/nginx.html
Tested, warning message disappeared


thomwiggers:
Are these lines not also handled by the new rules? 

```
        # The following 2 rules are only needed for the user_webfinger app.
        # Uncomment it if you're planning to use this app.
        #rewrite ^/.well-known/host-meta /public.php?service=host-meta last;
        #rewrite ^/.well-known/host-meta.json /public.php?service=host-meta-json last;


        # The following rule is only needed for the Social app.
        # Uncomment it if you're planning to use this app.
        #rewrite ^/.well-known/webfinger /public.php?service=webfinger last;
```

doc75:
@J0WI and @thomwiggers I just updated all the nginx.conf files through #1478

Issue #420, opened 2018-08-14
Installation Wizard not accepting username and password


umardraz:
Hello

The Installation wizard is not picking mysql information Here is the information what I had provided through environment variables.

```
-e MYSQL_HOST=10.0.201.120 \
-e MYSQL_DATABASE=ncdb \
-e MYSQL_USER=ncloud \
-e MYSQL_PASSWORD=ncloud12345 \

```
On the first run the installation appears and asked for crate an admin account.

But when I clicked on Finish setup button then an error appears.

Error while trying to create admin user: Failed to connect to the database: An exception occured in driver: SQLSTATE[HY000] [1044] Access denied for user ‘oc_admin15’@’%’ to database ‘client1’

```
‘dbtype’ =>; ‘mysql’,
‘version’ =>; ‘13.0.5.2’,
‘dbname’ =>; ‘ncdb’,
‘dbhost’ =>; ‘10.0.201.120’,
‘dbport’ =>; ‘’,
‘dbtableprefix’ =>; ‘’,
‘mysql.utf8mb4’ =>; true,
‘dbuser’ =>; ‘oc_admin15’,
‘dbpassword’ =>; ‘Q9hwc3uqbzqU2G9E+TkJt6XWGqmGjE’,
```

Don't know why it is not accepting username and password, the other options are ok, like dbname, dbhost.

Would you please any body help me.

r2evans:
I'm also seeing this behavior when trying to connect to a postgres instance. 

```
root@D2SP3:/home/r2/docker# docker-compose up nextcloud
docker_postgres_1_4b526486f7ce is up-to-date
Creating docker_nextcloud_1_21d381008fc1 ... done
Attaching to docker_nextcloud_1_a56357f7c235
nextcloud_1_a56357f7c235 | Initializing nextcloud 14.0.4.2 ...
nextcloud_1_a56357f7c235 | Initializing finished
nextcloud_1_a56357f7c235 | New nextcloud instance
nextcloud_1_a56357f7c235 | Installing with PostgreSQL database
nextcloud_1_a56357f7c235 | starting nextcloud installation
nextcloud_1_a56357f7c235 | Error while trying to create admin user: Failed to connect to the database: An exception occured in driver: SQLSTATE[08006] [7] FATAL:  permission denied for database "nc"
nextcloud_1_a56357f7c235 | DETAIL:  User does not have CONNECT privilege.
nextcloud_1_a56357f7c235 |  -> 
[...]

PR #1435, opened 2021-03-05
Fix for issue #420, entrypoint has a race condition where installation fails if DB is not ready in time.

dghodgson:
The NextCloud installer seems to set a default DB user and password in config.php if the first installation attempt fails (see issue #420 ) . This PR gets around that issue by waiting for the DB service to become available before proceeding with installation.

This adds [wait-for-it](https://github.com/vishnubob/wait-for-it) to /usr/bin in the dockerfile templates. The script is called in entrypoint.sh right before installation is attempted.

A helper function `get_host_string()` is also added to set default ports for MySQL/Postgres if no port is provided in the DB host string.

I attempted to implement the `wait_for()` function from wait-for-it into the entrypoint script directly, but ran into various issues (including the use of `/bin/sh` by entrypoint.sh instead of `/bin/bash`, which complicated things a bit). Calling the script just works without any issues, but that does mean adding another file to the docker image.

Not sure if running the update script to update the files in 19/20/21 is required as part of the pull request, so I opted not to.


dghodgson:
Seeing the alpine images failing due to a lack of bash.

Bash provides /dev/tcp which can be used to test network connections without an added dependency like `netcat`. Will have to look into another solution that works on alpine.

PR #1444, opened 2021-03-13
Add "legacy_auth" option for S3 object stores

nupplaphil:
Because of some MinIO issues with kubernetes, nextcloud and MinIO I did switch to the "legacy_auth" protocoll, see: 
https://github.com/nextcloud/server/blob/137636b651168712ee88beaa8ac0483e38bf42b4/lib/private/Files/ObjectStore/S3ConnectionTrait.php#L128

I noticed that this setting was missing at the s3 config as well.

### DIFF
--- a/.config/s3.config.php
+++ b/.config/s3.config.php
@@ -18,7 +18,7 @@ if (getenv('OBJECTSTORE_S3_BUCKET')) {
         'autocreate' => (strtolower($autocreate) === 'false' || $autocreate == false) ? false : true,
         'use_ssl' => (strtolower($use_ssl) === 'false' || $use_ssl == false) ? false : true,
         // required for some non Amazon S3 implementations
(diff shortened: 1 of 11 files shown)


.config/s3.config.php
```diff
@@ -17,7 +18,9 @@
         'autocreate' => (strtolower($autocreate) === 'false' || $autocreate == false) ? false : true,
         'use_ssl' => (strtolower($use_ssl) === 'false' || $use_ssl == false) ? false : true,
         // required for some non Amazon S3 implementations
-        'use_path_style' => $use_path == true && strtolower($use_path) !== 'false'
+		'use_path_style' => $use_path == true && strtolower($use_path) !== 'false',
```
J0WI, line 21:
```suggestion
        'use_path_style' => $use_path == true && strtolower($use_path) !== 'false',
```

PR #1459, opened 2021-03-25
Explicitly set locale settings

### LINKED ISSUE #721: Locale problem and support for UTF-8 file names
Used names of files and folders with accents (Brazil pt_BR). I had to set the chart and set the environment variables.
To work I had to add these environment variables:
```
        env:
        - name: LC_ALL
          value: C.UTF-8
        - name: TZ
          value: America/SaoPaulo
```

kurzdev:
As of right now, this only works on non-alpine images, as `locale` is not yet supported by Alpine Linux. Therefore, only changes to the debian template have been made. There are some workarounds for Alpine as well (as seen [here](https://github.com/gliderlabs/docker-alpine/issues/144#issuecomment-339906345)), but this seems a little too hacky for me and maybe even unnecessary. Would be nice if someone with more experience could comment on that.

Fixes #721

### DIFF
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -14,6 +14,10 @@ RUN set -ex; \
     mkdir -p /var/spool/cron/crontabs; \
     echo '*/%%CRONTAB_INT%% * * * * php -f /var/www/html/cron.php' > /var/spool/cron/crontabs/www-data
 
+# setting proper locale to prevent encoding issues
+ENV LANG C.UTF-8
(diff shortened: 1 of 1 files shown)


pierreozoux:
Looks good to me, But I'd prefer to double check with other maintainers.

pierreozoux (approved):

J0WI:
I can't reproduce the issue in #721 and therefore can hardly review this.

Issue #763, opened 2019-05-30
Introduced redis session handler breaks entrypoint of non-root images


smueller18:
In commit 83ea69d54bea72e1df9a2647ee101c73deb86d44, a redis session handler was intrdoduced. Unfortunately, all images after this commit do not work with a non-root user when environment `REDIS_HOST` is set. The file `/usr/local/etc/php/conf.d/redis-session.ini` can not be written, because the folder is owned by root and other users do not have write permission.
```
/var/www/html # ls -lha /usr/local/etc/php/conf.d
total 72
drwxr-xr-x    1 root     root        4.0K Feb 26 08:50 .
drwxr-xr-x    1 root     root        4.0K Feb 26 07:34 ..
-rw-r--r--    1 root     root          35 Feb 26 08:50 docker-php-ext-apcu.ini
-rw-r--r--    1 root     root          18 Feb 26 08:47 docker-php-ext-exif.ini
-rw-r--r--    1 root     root          16 Feb 26 08:47 docker-php-ext-gd.ini
-rw-r--r--    1 root     root          21 Feb 26 08:50 docker-php-ext-imagick.ini
-rw-r--r--    1 root     root          18 Feb 26 08:48 docker-php-ext-intl.ini
-rw-r--r--    1 root     root          18 Feb 26 08:48 docker-php-ext-ldap.ini
-rw-r--r--    1 root     root          23 Feb 26 08:50 docker-php-ext-memcached.ini
-rw-r--r--    1 root     root          82 Feb 26 08:48 docker-php-ext-opcache.ini
-rw-r--r--    1 root     root          19 Feb 26 08:48 docker-php-ext-pcntl.ini
-rw-r--r--    1 root     root          23 Feb 26 08:49 docker-php-ext-pdo_mysql.ini
-rw-r--r--    1 root     root          23 Feb 26 08:49 docker-php-ext-pdo_pgsql.ini
-rw-r--r--    1 root     root          19 Feb 26 08:50 docker-php-ext-redis.ini
-rw-r--r--    1 root     root          20 Feb 26 07:34 docker-php-ext-sodium.ini
-rw-r--r--    1 root     root          17 Feb 26 08:49 docker-php-ext-zip.ini
-rw-r--r--    1 root     root          18 Feb 26 08:50 memory-limit.ini
-rw-r--r--    1 root     root         189 Feb 26 08:50 opcache-recommended.ini
```

For running my nextcloud image, I use the default non-privileged user `www-data`.
```
[...]

PR #1469, opened 2021-04-06
cron and redis fixes for running with non-root UID

RlndVt:
When not running as 'www-data', the redis and cron functionality (as shown in examples such as ".examples/docker-compose/insecure/mariadb-cron-redis/fpm/docker-compose.yml") break. These commits fix these issues. 

Partial fix for #359

Possible supplement to #1278.

### DIFF
--- a/20/apache/Dockerfile
+++ b/20/apache/Dockerfile
@@ -11,7 +11,10 @@ RUN set -ex; \
         busybox-static \
         libldap-common \
     ; \
-    rm -rf /var/lib/apt/lists/*
+    rm -rf /var/lib/apt/lists/*; \
+    \
+    touch /usr/local/etc/php/conf.d/redis-session.ini; \
+    chmod o+rw /usr/local/etc/php/conf.d/redis-session.ini;
 
 # install the PHP extensions we need
 # see https://docs.nextcloud.com/server/stable/admin_manual/installation/source_installation.html
--- a/20/apache/Dockerfile
+++ b/20/apache/Dockerfile
@@ -11,10 +11,7 @@ RUN set -ex; \
         busybox-static \
         libldap-common \
     ; \
-    rm -rf /var/lib/apt/lists/*; \
-    \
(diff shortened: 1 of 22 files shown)


RlndVt:
> I don't think that making the PHP config world writeable is a good practice.

In general I agree. Any alternative suggestion is welcome.

danieljkemp:
If the goal was to make the container run as *a* non-root UID (www-data:33)
it wouldn't be too bad, just chown the required files. That likely wouldn't
meet the goal of some, but it could be an improvement.

I solved this in k8s by just mounting a templates redis config as a
configmap, and not using that particular redis_host variable.

RlndVt:
> If the goal was to make the container run as *a* non-root UID (www-data:33) it wouldn't be too bad, just chown the required files. That likely wouldn't meet the goal of some, but it could be an improvement.

Problem is that at the time docker runs through this part of the entrypoint:

https://github.com/nextcloud/docker/blob/master/docker-entrypoint.sh#L75-L100

we've dropped to a uid/gid, (L53-73) but the Dockerfile does not know what that uid/gid it will be. Making it writable by all is a generic fix, but not great security wise.

>  I solved this in k8s by just mounting a templates redis config as a configmap, and not using that particular redis_host variable.

I don't know how you would apply that here.

-----

PR fixes #763

danieljkemp:
Yeah, that is why I commented that if the UID was known ahead of time
(www-data) this would be easier. Kubernetes makes it really easy to store a
config file with the rest of your manifests and just patch in whatever you
want.

The problem is that most security policies that require containers run as
root actually require the container to start as non-root. There are often
admission hooks or runtime checks that will entirely prevent workloads from
running, starting as root and dropping permissions will not satisfy them.

Changing the perms to www-data (33) would allow the container to run fully
as non-root provided that that uid is used. It would still require root if
a different uid was needed but this would still be an improvement in some
cases.

PR #1478, opened 2021-04-18
update nginx.conf examples to match current NC20 official documentation related to /.well-known

doc75:
The official documentation is mentioning a different configuration related to the /.well-known section.
This PR aims at aligning the examples files with the documentation on this specific point.

It seems that there are other parts that could be updated, but I am afraid to break things as I am not a specialist of NGINX config.

If anything need to be reworked, just let me know.

### DIFF
--- a/.examples/docker-compose/insecure/mariadb/fpm/web/nginx.conf
+++ b/.examples/docker-compose/insecure/mariadb/fpm/web/nginx.conf
@@ -63,21 +63,22 @@ http {
             access_log off;
         }
 
-        # The following 2 rules are only needed for the user_webfinger app.
-        # Uncomment it if you're planning to use this app.
-        #rewrite ^/.well-known/host-meta /public.php?service=host-meta last;
-        #rewrite ^/.well-known/host-meta.json /public.php?service=host-meta-json last;
-
-        # The following rule is only needed for the Social app.
-        # Uncomment it if you're planning to use this app.
-        #rewrite ^/.well-known/webfinger /public.php?service=webfinger last;
-
-        location = /.well-known/carddav {
-            return 301 $scheme://$host:$server_port/remote.php/dav;
-        }
-
-        location = /.well-known/caldav {
-            return 301 $scheme://$host:$server_port/remote.php/dav;
+        # Make a regex exception for `/.well-known` so that clients can still
+        # access it despite the existence of the regex rule
+        # `location ~ /(\.|autotest|...)` which would otherwise handle requests
+        # for `/.well-known`.
+        location ^~ /.well-known {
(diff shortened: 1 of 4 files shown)


doc75:
I am surprised that some tests are failing as I guess that examples are not impacting tests... If anything need to be updated, just let me know.

.examples/docker-compose/with-nginx-proxy/mariadb-cron-redis/fpm/web/nginx.conf
```diff
@@ -68,17 +68,22 @@ http {
...
+            rewrite ^/\.well-known/webfinger        /public.php?service=webfinger       last;
+            rewrite ^/\.well-known/nodeinfo         /public.php?service=nodeinfo        last;
+    
+            location = /.well-known/carddav     { return 301 /remote.php/dav/; }
+            location = /.well-known/caldav      { return 301 /remote.php/dav/; }
+    
+            try_files $uri $uri/ =404;
         }
 
         location = /.well-known/caldav {
```
0x47, line 89: Section should be removed as well

doc75: @0x47 well spotted. My mistake. Update done

0x47:
I'm not sure, it could be an issue on my end, but I still receive these messages in Settings -> Overview:
```
Your web server is not properly set up to resolve "/.well-known/webfinger". Further information can be found in the documentation.
Your web server is not properly set up to resolve "/.well-known/nodeinfo". Further information can be found in the documentation.
Your web server is not properly set up to resolve "/.well-known/caldav". Further information can be found in the documentation.
Your web server is not properly set up to resolve "/.well-known/carddav". Further information can be found in the documentation.
```
Anything I need to do besides updating the `nginx.conf` file according to this PR? I double checked and confirmed that the lines are updated in my `web` container.

doc75:
@0x47  was it an issue that the nginx conf was sending back an HTTP address instead of an HTTPS ?

How did you solve this ? (I have the same issue as my nginx is behind a traefik proxy which is the SSL endpoint and the only by-pass on my side is to add a middleware on traefik directly).

0x47:
@doc75 I don't know what the technical reason was but most likely it was HTTP/HTTPS related. I set up my NC instance from a LAN connection and did use HTTP to access it. The warnings disappeared automatically after I was done setting up and switched to accessing it from the internet via HTTPS (Letsencrypt).

BTW, the suggested configs for NC 21 are slightly different to the ones from NC 20 as well ([here](https://docs.nextcloud.com/server/latest/admin_manual/installation/nginx.html#nextcloud-in-the-webroot-of-nginx)).

PR #1481, opened 2021-04-20
added ghostscript for pdf preview

### LINKED ISSUE #1460: ImageMagick requires ghostscript for pdf preview
Configuring the preview generator to generate pdf previews I found that the default docker images lacks the ghostscript binary.
It would be great if you could add a `apt install ghostscript` to the nextcloud docker images.

Janl1:
If you enable pdf preview in the preview generator ghostscript binary is required.
Closes #1460

Signed-off-by: Jan Lahmer <jan@lahmer.eu>

### DIFF
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -32,6 +32,7 @@ RUN set -ex; \
         imagemagick-dev \
         libwebp-dev \
         gmp-dev \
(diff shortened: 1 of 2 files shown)


J0WI:
Thanks for your patch, but we decided recently to not include this feature by default: #1414

PR #1482, opened 2021-04-20
Fix broken smtp.config.php file

0x47:
The current `smtp.config.php` file does not work as advertised in the documentation. Both `MAIL_FROM_ADDRESS` and `MAIL_DOMAIN` should be optional as the Nextcloud instance may use multiple _from addresses_ like no-reply@example.com or passwordreset-no-reply@example.com (or similar). It may also just use the domain name from the instance instead of a configured one.

I tested this PR on NC 21 as I don't have access to the previous versions at the moment. I propose to leave NC 19 annd 20 as they are and to continue with this `smtp.config.php` from version 21 forward.

### DIFF
--- a/21.0/fpm-alpine/config/smtp.config.php
+++ b/21.0/fpm-alpine/config/smtp.config.php
@@ -1,5 +1,5 @@
 <?php
-if (getenv('SMTP_HOST') && getenv('MAIL_FROM_ADDRESS') && getenv('MAIL_DOMAIN')) {
+if (getenv('SMTP_HOST')) {
   $CONFIG = array (
     'mail_smtpmode' => 'smtp',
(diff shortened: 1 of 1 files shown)


J0WI (approved):
LGTM, please check https://github.com/nextcloud/docker/pull/1482/checks?check_run_id=2390227909

0x47:
@J0WI Thanks, done!

J0WI:
Oh, this just changed the generated file and not [the template](https://github.com/nextcloud/docker/blob/master/.config/smtp.config.php).

0x47:
Hm, I never knew that existed... I could create a new PR after I upgraded my instance. Might take a few days though...

PR #1511, opened 2021-06-02
Imagemagick not installed by default on debian versions (cf SVG troub…

### LINKED ISSUE #1414: SVG imagemagick Nextcloud 21.0 Docker

vincowl:
…les)

Resolves #1414

### DIFF
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -8,6 +8,7 @@ RUN set -ex; \
         rsync \
         bzip2 \
         busybox-static \
+        imagemagick \
     ; \
     rm -rf /var/lib/apt/lists/*; \
     \


J0WI:
I'm closing this due the discussion in https://github.com/nextcloud/docker/issues/1414#issuecomment-811896282.

Issue #1148, opened 2020-06-10
Support for docker secrets on first initialization broken 


KnallbertLp:
Unfortunately the problem described in #385 reappeared again... 
The exact thing happens, that I describe in [my comment on #385](https://github.com/nextcloud/docker/issues/385#issuecomment-641840068)

> Unfortunately this seems to be broken again...
> 
> ![Unknown](https://user-images.githubusercontent.com/19236891/84245717-c6fee080-ab05-11ea-9fa2-9e981fafc90f.PNG)
> 
> When built with this docker-compose . yaml it still asks for database configuration. If built with an inline password on the other hand, it works like intended.
> 
> `---
> version: '3.3'
> 
> services:
> nextcloud-db:
> image: mariadb
> container_name: nextcloud-db
> command: --transaction-isolation=READ-COMMITTED --log-bin=ROW
> networks:
> - nextcloud-db
> restart: always
> environment:
> MYSQL_ROOT_PASSWORD_FILE: /run/secrets/mysql_root_password
> MYSQL_PASSWORD_FILE: /run/secrets/mysql_user_password
> # MYSQL_PASSWORD: 345Test
> MYSQL_DATABASE: nextcloud
> MYSQL_USER: nextcloud
> MYSQL_INITDB_SKIP_TZINFO: 1
> secrets:
> - mysql_root_password
> - mysql_user_password
> --truncated--
> 
> nextcloud-app:
> image: nextcloud
> container_name: nextcloud-app
> restart: always
> depends_on:
> - nextcloud-db
> environment:
> MYSQL_HOST: nextcloud-db
> MYSQL_USER: nextcloud
> MYSQL_DATABASE: nextcloud
> MYSQL_PASSWORD_FILE: /run/secrets/mysql_user_password
> # MYSQL_PASSWORD: 345Test
> secrets:
> - mysql_user_password
> --truncated--
> 
> secrets:
> mysql_root_password:
> file: /opt/docker/secrets/mysql_root_password
> mysql_user_password:
> file: /opt/docker/secrets/mysql_user_password`

I would be very glad if someone might help me or point me in the right direction!

0x326:
Can confirm. Using `POSTGRES_PASSWORD_FILE` results defaults to the SQLite database whereas using `POSTGRES_PASSWORD` auto-configures the database properly

isdnfan:
[...]

PR #1516, opened 2021-06-15
Added fetching autoconfig data from Docker secrets

tushev:
Signed-off-by: Simon Tushev <tushev@users.noreply.github.com>

This should solve #1148

This is my first PR. I'm not very familiar with Docker, however, I hope that this solves the issue.
I changed only `.config/autoconfig.php` , as far as I can understand these changes should propagate to, say, `21.0/apache/config/autoconfig.php` automatically.

Please correct me if I'm wrong.

Another question to consider is the security of `trim(file_get_contents(getenv('MYSQL_DATABASE_FILE')));`. Theoretically, it is possible to specify any file, however, the very same approach is used in `entrypoint.sh`. Both the installation script and Docker secrets should be controlled by a privileged user. I see no issues with this, however, I'm not a security expert and thus I request a second opinion.

### DIFF
--- a/.config/autoconfig.php
+++ b/.config/autoconfig.php
@@ -6,13 +6,6 @@ if (getenv('SQLITE_DATABASE')) {
     $AUTOCONFIG['dbtype'] = 'sqlite';
     $AUTOCONFIG['dbname'] = getenv('SQLITE_DATABASE');
     $autoconfig_enabled = true;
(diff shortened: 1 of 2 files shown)


J0WI (commented):
I'd check first for the `_FILE` and then fallback to the env.

tushev:
@J0WI Should it be a new pull request or this one? (a novice's question :)

J0WI:
You can update this one

tushev:
Done, also added some info to README.md

J0WI (approved):

Issue #1536, opened 2021-07-13
mariadb needs to be locked to 10.5


pbek:
Today the `mariadb:latest` [docker image](https://registry.hub.docker.com/_/mariadb) got bumped to 10.6.3. :fearful:
After an update you will run into https://help.nextcloud.com/t/update-to-next-cloud-21-0-2-has-get-an-error/117028/14 with Nextcloud 21!
Doing a `image: mariadb:10.5` helped!

I wanted to create a pull request for the documentation change in the readme (https://github.com/nextcloud/docker/commit/94a24c2385a7be70a23a1c02661ddc4c675f4c32), but I didn't realize I've write permission to the repository, so the commit went through directly into the master branch. I'm sorry.

pbek:
Now with proper PR: https://github.com/nextcloud/docker/pull/1537

pbek:
Maybe this also could be fixed with `--skip-innodb-read-only-compressed` in mariadb 10.6...

Vitlik:
Can confirm.
The error was gone with `--skip-innodb-read-only-compressed`

thx!

pbek:
Thank you for confirming!

swiesend:
Where do I have to put this option `--skip-innodb-read-only-compressed` if I do not want to downgrade to `10.5`? I have a `docker-compose` setup.

pbek:
I haven't tested it, but yes, I guess you need to add it to `command`. But I didn't dare to stay on 10.6, do you need anything from 10.6 (until this issue is resolved in Nextcloud itself)?

Vitlik:
Yes, my command line looked like this:
`command: --transaction-isolation=READ-COMMITTED --binlog-format=ROW --skip-innodb-read-only-compressed`

swiesend:
No not really, but I see myself running into annoying problems when staying on `10.5` and forget about updating it in my `docker-compose.yml` in the future.

pbek:
https://mariadb.com/kb/en/changes-improvements-in-mariadb-105/ says:

MariaDB 10.5 is the current major stable version The first stable release was in June 2020, and it **will be supported until June 2025**.

LukBukkit:
Hey, I just want to throw in a link to the issue #1492, which solves the problem for MariaDB 10.6 using the parameter `--innodb-read-only-compressed=OFF`. I guess it functions the same as `--skip-innodb-read-only-compressed`.

x86dev:
[...]

PR #1537, opened 2021-07-13
Lock mariadb to 10.5 (#1536)

pbek:
Helps with #1536.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Database:
 ```console
 $ docker run -d \
 -v db:/var/lib/mysql \
-mariadb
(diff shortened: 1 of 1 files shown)


pbek:
I'll accept my own PR now, so new users don't end up with a broken installation.

thomasmerz:
Why didn't you also update the `docker.yml` and `docker-compose.yml` files for users convenience? 🤔

pbek:
I still don't know if there is a "official statement" from Nextcloud itself that locking to mariadb 15 is the "official" way to handle that issue. 😅

PR #1548, opened 2021-07-22
Apache no rsync

ChristianKniep:
Some tweaks to the entrypoint and Dockerfile to not copy files that are tied to a version of nextcloud.
Instead use multiple volumes to hold r/w directories and make the rest of the container file-system r/o.

As discussed with @f7o

```
cd 21.0/apache
docker build -t nextcloud/apache:21.0.2 --build-arg=NEXTCLOUD_VERSION=21.0.2 .
docker build -t nextcloud/apache:21.0.3 --build-arg=NEXTCLOUD_VERSION=21.0.3 .
export NC_IMAGE_NAME=nextcloud/apache:21.0.2
```
### Start the Stack
First we start the database in the background

```
$ docker-compose up -d db
Creating volume "apache_nextcloud_data" with default driver
Creating volume "apache_nextcloud_apps" with default driver
Creating volume "apache_nextcloud_config" with default driver
Creating volume "apache_nextcloud_themes" with default driver
Creating apache_db_1 ... done
```

Afterwards starting the nextcloud:21.0.2 instance
```
$ docker-compose up nextcloud 
apache_db_1 is up-to-date
Creating apache_nextcloud_1 ...
Attaching to apache_nextcloud_1
nextcloud_1  | + expr apache2-foreground : apache
nextcloud_1  | + [ -n  ]
nextcloud_1  | + expr apache2-foreground : apache
nextcloud_1  | + [ -n  ]
nextcloud_1  | + installed_version=0.0.0.0
nextcloud_1  | + php -r require "/var/www/html/version.php"; echo implode(".", $OC_Version);
nextcloud_1  | + image_version=21.0.2.1
...
```

Now we are able to open [localhost:8000](http://localhost:8000) and login as `administrator`/`adminpass`

Make some changes: added a user and some files. 

We are going to stop the container (CTRL + C) and update to a new version.

```
$ export NC_IMAGE_NAME=nextcloud/apache:21.0.3
docker-compose up nextcloud
```

### DIFF
--- a/21.0/apache/Dockerfile
+++ b/21.0/apache/Dockerfile
@@ -1,7 +1,6 @@
 # DO NOT EDIT: created by update.sh from Dockerfile-debian.template
 FROM php:7.4-apache-buster
 
-ENV NEXTCLOUD_VERSION 21.0.3
 ENV PHP_MEMORY_LIMIT 512M
 ENV PHP_UPLOAD_LIMIT 512M
 
@@ -123,7 +122,8 @@ RUN a2enmod headers rewrite remoteip ;\
     } > /etc/apache2/conf-available/remoteip.conf;\
     a2enconf remoteip
 
-
+ARG NEXTCLOUD_VERSION=21.0.3
+ENV NEXTCLOUD_VERSION=${NEXTCLOUD_VERSION}
 RUN set -ex; \
     fetchDeps=" \
         gnupg \
--- a/21.0/apache/Dockerfile
+++ b/21.0/apache/Dockerfile
@@ -1,6 +1,10 @@
 # DO NOT EDIT: created by update.sh from Dockerfile-debian.template
 FROM php:7.4-apache-buster
 
+ENV NEXTCLOUD_VERSION 21.0.3
+ENV PHP_MEMORY_LIMIT 512M
+ENV PHP_UPLOAD_LIMIT 512M
+
 # entrypoint.sh and cron.sh dependencies
 RUN set -ex; \
     \
@@ -17,8 +21,6 @@ RUN set -ex; \
 
 # install the PHP extensions we need
 # see https://docs.nextcloud.com/server/stable/admin_manual/installation/source_installation.html
-ENV PHP_MEMORY_LIMIT 512M
-ENV PHP_UPLOAD_LIMIT 512M
 RUN set -ex; \
     \
     savedAptMark="$(apt-mark showmanual)"; \
@@ -121,7 +123,6 @@ RUN a2enmod headers rewrite remoteip ;\
     } > /etc/apache2/conf-available/remoteip.conf;\
     a2enconf remoteip
 
-ENV NEXTCLOUD_VERSION 21.0.3
 
 RUN set -ex; \
     fetchDeps=" \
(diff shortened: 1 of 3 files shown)


21.0/apache/Dockerfile
```diff
@@ -121,8 +122,8 @@ RUN a2enmod headers rewrite remoteip ;\
     } > /etc/apache2/conf-available/remoteip.conf;\
     a2enconf remoteip
 
-ENV NEXTCLOUD_VERSION 21.0.3
```
ChristianKniep, line 124: Using `ARG` to be able to dynamically update the version without changing the Dockerfile.

21.0/apache/Dockerfile
```diff
@@ -139,19 +140,44 @@ RUN set -ex; \
...
+    chmod +x /var/www/html/occ; \
     \
     apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false $fetchDeps; \
-    rm -rf /var/lib/apt/lists/*
+    rm -rf /var/lib/apt/lists/* \
+ && mkdir -p /usr/src/nextcloud \
+ && mv /var/www/html/themes /usr/src/nextcloud/ \
+ && mv /var/www/html/config /usr/src/nextcloud/
+
+### Volumes
```
ChristianKniep, line 155: I use a bunch of `VOLUMES` to make explicit what directories need to be read/write.
In case they are not externally defined (see docker-compose.yml) they are going to contain what the container file-system has in that path up until that point.

21.0/apache/Dockerfile
```diff
@@ -139,19 +140,44 @@ RUN set -ex; \
...
+VOLUME /var/www/html/themes
 
-COPY *.sh upgrade.exclude /
+COPY entrypoint.sh /usr/local/bin/
+COPY cron.sh /
 COPY config/* /usr/src/nextcloud/config/
 
-ENTRYPOINT ["/entrypoint.sh"]
-CMD ["apache2-foreground"]
+ENTRYPOINT []
```
ChristianKniep, line 182: To make `docker run` possible, I switch to `CMD`.

J0WI:
Thanks for your contribution. This contains some relatively invasive changes. Some of them violates the [best practice requirements](https://github.com/docker-library/official-images#consistency) for official images. If there is no strong reason for these changes it's unlikely that we'll merge them.

kniec:
@J0WI Can you point me to the violations so that I can mitigate them? 
IMHO this changes make the docker image more useable as it does not `rsync` all the files into a local volume.

21.0/apache/Dockerfile
J0WI: `ARG` cannot be used in official images

21.0/apache/Dockerfile
J0WI: Creating a bunch of `VOLUMES` is a bad practice. They are created on each run and there is no way to undefine them.  You can create as many (read only) mounts as you need with `--volume`.

21.0/apache/Dockerfile
J0WI: This is already possible due the `exec "$@"` in the entypoint script.

alexfornuto:
As someone who lost a lot of time and effort when the rsync nuked the mount point where I copied my data from an old host, I'd love to see some activity on this.

Issue #1421, opened 2021-02-22
Use PHP 8.0 for Nextcloud 22 image


siccovansas:
The current Nextcloud 21 image still uses PHP 7.4. [Nextcloud 21 supports PHP 8.0](https://nextcloud.com/blog/nextcloud-hub-21-out-with-up-to-10x-better-performance-whiteboard-and-more-collaboration-features/) and it gives some performance improvements, so it would be awesome if the Nextcloud 21 image uses PHP 8.0.

markuman:
The main issue is, that the latest version of imagick (3.4.4) that is shipped with `pecl` is not compatible with php 8

https://github.com/Imagick/imagick/issues/331#issuecomment-779190777

markuman:
works for me.

```diff
diff --git a/21.0/apache/Dockerfile b/21.0/apache/Dockerfile
index 52fab6e..ba6f3bf 100644
--- a/21.0/apache/Dockerfile
+++ b/21.0/apache/Dockerfile
@@ -1,5 +1,5 @@
 # DO NOT EDIT: created by update.sh from Dockerfile-debian.template
-FROM php:7.4-apache-buster
+FROM php:8.0-apache-buster
 
 # entrypoint.sh and cron.sh dependencies
 RUN set -ex; \
@@ -60,10 +60,12 @@ RUN set -ex; \
     ; \
     \
 # pecl will claim success even if one install fails, so we need to perform each install separately
-    pecl install APCu-5.1.19; \
-    pecl install memcached-3.1.5; \
-    pecl install redis-5.3.3; \
-    pecl install imagick-3.4.4; \
+    pecl install APCu; \
+    pecl install memcached; \
+    pecl install redis; \
+    mkdir -p /usr/src/php/ext/imagick; \
+    curl -fsSL https://github.com/Imagick/imagick/archive/06116aa24b76edaf6b1693198f79e6c295eda8a9.tar.gz | tar xvz -C "/usr/src/php/ext/imagick" --strip 1; \
+    docker-php-ext-install imagick; \
     \
     docker-php-ext-enable \
         apcu \
@@ -130,17 +132,12 @@ RUN set -ex; \
     apt-get update; \
     apt-get install -y --no-install-recommends $fetchDeps; \
     \
-    curl -fsSL -o nextcloud.tar.bz2 \
-        "https://download.nextcloud.com/server/releases/nextcloud-${NEXTCLOUD_VERSION}.tar.bz2"; \
[...]

PR #1563, opened 2021-08-23
PHP 8.0 for 22+

J0WI:
Fix https://github.com/nextcloud/docker/issues/1421


FloThinksPi:
I tested this change with my environment and it works as intended, updates all base images correctly and those seem to speed things up quite a bit.

PR #1571, opened 2021-08-30
Switch to new nginx companion registry.

rafgk:
As stated in https://hub.docker.com/r/jrcs/letsencrypt-nginx-proxy-companion/ the the project has moved to a new registry and therefore that should be used instead.


J0WI:
Are any config changes required?

PR #1588, opened 2021-09-17
Fix memory limit config issue (close #1413)

### LINKED ISSUE #1413: Allowed memory size of xxx bytes exhausted in current latest-container
After upgrading my docker-setup to the current latest-container from the 21-branch some hours ago, I got the following error messages:
```
docker exec nextcloud su - www-data -s /bin/bash -c 'php -f /var/www/html/cron.php'

Fatal error: Allowed memory size of 2097152 bytes exhausted (tried to allocate 438272 bytes) in /var/www/html/3rdparty/composer/autoload_real.php on line 37
```

This happens basically with every php-command, e.g. occ to add the missing indices. Something seems wrong here with PHP Memory Limits. The following works fine:

```
docker exec nextcloud su - www-data -s /bin/bash -c 'php -d memory_limit=-1 -f /var/www/html/cron.php'
```

The PHP_MEMORY_LIMIT variable within the container seems to be set to 512. I found an older support-ticket regarding issues with this limit:
https://help.nextcloud.com/t/allowed-memory-size-of-xxx-bytes-exhausted/39371

So there seems an issue with the container's internal environment configuration.

dashohoxha:
Replace single quotes by double quotes on Dockerfiles.

### DIFF
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -89,9 +89,9 @@ RUN { \
     echo 'apc.enable_cli=1' >> /usr/local/etc/php/conf.d/docker-php-ext-apcu.ini; \
     \
     { \
-        echo 'memory_limit=${PHP_MEMORY_LIMIT}'; \
-        echo 'upload_max_filesize=${PHP_UPLOAD_LIMIT}'; \
-        echo 'post_max_size=${PHP_UPLOAD_LIMIT}'; \
+        echo "memory_limit=${PHP_MEMORY_LIMIT}"; \
+        echo "upload_max_filesize=${PHP_UPLOAD_LIMIT}"; \
+        echo "post_max_size=${PHP_UPLOAD_LIMIT}"; \
     } > /usr/local/etc/php/conf.d/nextcloud.ini; \
     \
     mkdir /var/www/data; \
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -102,9 +102,9 @@ RUN { \
     echo 'apc.enable_cli=1' >> /usr/local/etc/php/conf.d/docker-php-ext-apcu.ini; \
     \
     { \
-        echo 'memory_limit=${PHP_MEMORY_LIMIT}'; \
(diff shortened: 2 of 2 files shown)


dashohoxha:
> This would basically revert #1379. The single quotes are intentional to read the variable on runtime.

That commit does not do what it claims on the subject/description, because you cannot read ENV variables on runtime when PHP runs as www-data.
It is creating [problems](https://www.google.com/search?q=nextcloud+Allowed+memory+size+of+xxx+bytes+exhausted+in+current+latest-container) to everyone (#1413) for a long time.

J0WI:
I just double checked this with a simple `phpinfo()` script and it shows me the correct values from the environment variables.

dashohoxha:
Would you describe how you did the test, so that we can repeat it? Or we just have to take your word for it?

J0WI:
Sure. I run `docker run --rm -e PHP_MEMORY_LIMIT=1024M -e PHP_UPLOAD_LIMIT=2048M --name nextcloud-env-test -p 8080:80 nextcloud`, `docker exec nextcloud-env-test sh -c 'echo "<?php phpinfo();" > index.php'` and verified the results at `http://localhost:8080/`.

dashohoxha:
You are right, it works. Thanks for sharing the instructions.

In the default docker image apache2 is called with `exec apache2 ...`:
https://github.com/docker-library/php/blob/master/8.0/bullseye/apache/apache2-foreground#L40
So it inherits the environment of root and this is why it can access the variables PHP_MEMORY_LIMIT and PHP_UPLOAD_LIMIT.

However, in my case, I am extending the standard image by installing `systemd`, so `apache2` is called by `systemd` and in this case it cannot read the variables PHP_MEMORY_LIMIT and PHP_UPLOAD_LIMIT, which are defined in the root's environment.
My workaround is to modify `/usr/local/etc/php/conf.d/nextcloud.ini` directly with the correct values.

dashohoxha:
I am closing this pull request.

PR #1599, opened 2021-10-03
update stable to 21.0.5

Bleuzen:
the official updater rolls out 21.0.5 to all 21 users now, so do the same with docker stable tag


Thomas131:
I think DCO needs to be fixed - https://github.com/nextcloud/docker/pull/1599/checks?check_run_id=3783827970

PR #1609, opened 2021-10-14
Add libldap-common as runtime dependency, not build dependency

### LINKED ISSUE #1572: image tag:22.1.1 no /etc/ldap/ldap.conf file
Causes ldaps to be unavailable unless you choose to ignore certificate verification

am97:
Fixes #1572


sleif:
good point - in the fix before I overlooked the savedAptMark mechanism

Issue #1063, opened 2020-04-08
Use nginx unit for nextcloud-fpm image


strannick-ru:
Hello.

How about use nginx unit in your nextcloud-fpm images?
For debian its really simple — just add nginx unit repo, install it, add config and make little change into entrypoint.sh.

What is for? You get best from two apps — php-fpm for php and nginx for static files.
It will be like nextcloud-apache image, but suitable for heavy loads.

dm3ch:
@J0WI Is there any chance that there would be added official docker image using nginx-unit? Not as a replacement of php-fpm, but as a new distribution option (like apache).
If maintainers are ready to merge such contribution, I'm ready to update existing ananace PR or recreate it from scratch.

P.S. Sorry for mentioning

tippexs:
Hello @J0WI sorry for picking up this "old" issue again. But a lot has changed since 2022 and we are now an official Docker Image and NGINX Unit is a true alternative to PHP-FPM.

I have just created a new configuration for NC 27.1.2. I would love to contribute this if we can find a way adding Unit to the NC update script in some sort. If this is not possible is there a way to contribute to your docs? To me a documentation about how to use NGINX Unit for NC is better placed somewhere on your end as on ours. I am happy to share everything we have to make this happen! Just let me know.

Cheers Timo from NGINX :)

gaigelama:
> Hello @J0WI sorry for picking up this "old" issue again. But a lot has changed since 2022 and we are now an official Docker Image and NGINX Unit is a true alternative to PHP-FPM.
> 
> I have just created a new configuration for NC 27.1.2. I would love to contribute this if we can find a way adding Unit to the NC update script in some sort. If this is not possible is there a way to contribute to your docs? To me a documentation about how to use NGINX Unit for NC is better placed somewhere on your end as on ours. I am happy to share everything we have to make this happen! Just let me know.
> 
> Cheers Timo from NGINX :)

I was just checking this thread out a couple days ago and was going to comment that there was now an official NGINX Unit image but didn't have time. I circled back today to see your reply and it couldn't get any better than that lol.

If you get something pushed to Docker Hub or another public registry I can test this out on Nextcloud on k8s and provide feedback.

Thanks for taking an interest and getting something started @tippexs .

tippexs:
[...]

PR #1610, opened 2021-10-14
Add a nginx-unit -based image

ananace:
This sets up a new variant for nginx-unit, based on their nginx-unit PHP image and the Nextcloud example in their documentation.

Probably related to #1063 - though ths solution is different than the one suggested there as nginx-unit already includes a FastCGI-based PHP solution.

Images for testing have been built at `ghcr.io/ananace/docker-nextcloud` - only x86_64 and aarch64 as that's what's available in the upstream unit image.

### DIFF
--- a/Dockerfile-unit.template
+++ b/Dockerfile-unit.template
@@ -1,0 +1,166 @@
+FROM unit:%%UNIT_VERSION%%-php%%PHP_VERSION%%
+
+# make nginx-unit co-operate with the nextcloud entrypoint
+RUN set -ex; \
+    \
+    mv /docker-entrypoint.d /unit-entrypoint.d; \
+    mv /usr/local/bin/docker-entrypoint.sh /usr/local/bin/unit-entrypoint.sh; \
+    sed -e 's/docker-entrypoint.d/unit-entrypoint.d/' -i /usr/local/bin/unit-entrypoint.sh
+
+# entrypoint.sh and cron.sh dependencies
+RUN set -ex; \
+    \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+        busybox-static \
+        bzip2 \
+        libldap-common \
+        libmagickcore-6.q16-6-extra \
+        rsync \
+    ; \
+    rm -rf /var/lib/apt/lists/*; \
+    \
+    mkdir -p /var/spool/cron/crontabs; \
+    echo '*/%%CRONTAB_INT%% * * * * php -f /var/www/html/cron.php' > /var/spool/cron/crontabs/www-data
+
+# install the PHP extensions we need
+# see https://docs.nextcloud.com/server/stable/admin_manual/installation/source_installation.html
+ENV PHP_MEMORY_LIMIT 512M
+ENV PHP_UPLOAD_LIMIT 512M
+RUN set -ex; \
+    \
+    savedAptMark="$(apt-mark showmanual)"; \
+    \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+        libcurl4-openssl-dev \
+        libevent-dev \
+        libfreetype6-dev \
+        libgmp-dev \
+        libicu-dev \
+        libjpeg-dev \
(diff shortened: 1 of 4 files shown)


ananace:
Pushed another version of this to do some more testing myself, also pushed a version as `ananace/nextcloud-unit:27` - only amd64/aarch64 as that's what's available for the upstream unit image.

ananace:
Continued work on things, it should now mostly match cache headers for nginx/apache as well. I wasn't able to get the `v` argument handling working - will be digging more into that later.

It's supposed to support setting cache control using an inline template like this, but any attempts to include inline templates like they describe cause unit to fail to load the configuration.
```json
"response_headers": {
  "Cache-Control": "public, max-age=15778463, `${ arg_v ? 'immutable' : '' }`"
}
```

tianon:
The main thing that makes me nervous here is the reimplementation of the `unit` entrypoint script inside the `nextcloud` entrypoint script -- what I'd suggest is copying `/usr/local/bin/docker-entrypoint.sh` inside the `Dockerfile` before overwriting it to something like `/usr/local/bin/unit-entrypoint.sh` which can be invoked directly instead so it can remain the source-of-truth for the fiddly bits of initializing unit instead.

I started fiddling with how I might implement this in the WordPress image, and I came up with something like this:

```dockerfile
RUN set -eux; \
	sed -e 's/docker-entrypoint/unit-entrypoint/g' /usr/local/bin/docker-entrypoint.sh > /usr/local/bin/unit-entrypoint.sh; \
	grep -F '/unit-entrypoint.d/' /usr/local/bin/unit-entrypoint.sh; \
	chmod +x /usr/local/bin/unit-entrypoint.sh
COPY wordpress-unit.json /unit-entrypoint.d/wordpress.json
```

This allows me to invoke `unit-entrypoint.sh` inside the new entrypoint to make sure Unit is initialized before we exec it.

I also don't know whether the `--control` bit from https://github.com/nginx/unit/blob/fb33ec86a3b6ca6a844dfa6980bb9e083094abec/pkg/docker/Dockerfile.php8.2#L89 is strictly necessary or whether that's the default value, but that's also something to consider.

ananace:
@tianon When I did the first version of this, it wasn't at all possible to use the original unit entrypoint script due to how it was set up, but I should be able to retool to let it do the configuration through there now.

The control argument is only necessary if you want to keep exposing the control socket, to be able to dynamically reconfigure unit while it's running, something which nextcloud lacks any functionality for.

tianon:
Ah, I see -- in case you're curious, I finished up my WordPress PoC and pushed it to https://github.com/docker-library/wordpress/pull/875 :sweat_smile:

ananace:
Retooled the nginx-unit image to use the upstream entrypoint, not sure about any `generate-stackbrew-library.sh` change though. I tried running it locally on my test setup and the generated output looked correct.  
Though I've only been pushing the changes to the configuration and scripts, not the generated files, to keep the PR manageable - and it's likely that those are required for it to generate the correct library file.

ananace:
Rebased the branch, added in an IP forward config block, and built new image tags for 27, 28, and 29 for testing.

ananace:
Pushed new 29 and 30 images to `ghcr.io/ananace/docker-nextcloud` - since the dockerhub rate limit is really annoying me at this point.

k4my4b:
> Continued work on things, it should now mostly match cache headers for nginx/apache as well. I wasn't able to get the `v` argument handling working - will be digging more into that later.
> 
> It's supposed to support setting cache control using an inline template like this, but any attempts to include inline templates like they describe cause unit to fail to load the configuration.
> 
> ```json
> "response_headers": {
>   "Cache-Control": "public, max-age=15778463, `${ arg_v ? 'immutable' : '' }`"
> }
> ```

what you're after is [this](https://unit.nginx.org/scripting/#examples):

```
"response_headers": {
    "Cache-Control": "`public, max-age=15778463${vars.arg_v ? ', immutable' : ''}`"
}
```

PR #1614, opened 2021-10-18
fix SMTP_PASSWORD_FILE not triggering auth activation

Murazaki:


gmcinalli:
@Murazaki I think you need to update the other `smtp.config.php` files as well.

```
21/apache/config/smtp.config.php
21/fpm-alpine/config/smtp.config.php
21/fpm/config/smtp.config.php
22/apache/config/smtp.config.php
22/fpm-alpine/config/smtp.config.php
22/fpm/config/smtp.config.php
23/apache/config/smtp.config.php
23/fpm-alpine/config/smtp.config.php
23/fpm/config/smtp.config.php
```

Murazaki:
This should do it :) let me know if there's more to do

Murazaki:
One liner patch that waited one year to be merged though. it might be time.

PR #1619, opened 2021-10-22
Disable security labels on the proxy and letsencrypt companion containers

tommytran732:
The security lables needs to be disabled, otherwise SELinux will deny access to the docker socket.
https://access.redhat.com/solutions/4344161


J0WI (changes requested):
Better add the [actual labels](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label).

PR #1637, opened 2021-11-16
run bash via env command

Stunkymonkey:
`#!/usr/bin/env` searches `PATH` for `bash`, and `bash` is not always in `/bin`

### DIFF
--- a/.travis/test-example-dockerfiles.sh
+++ b/.travis/test-example-dockerfiles.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 image="$1"
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -Eeuo pipefail
 
 stable_channel='22.2.3'
--- a/update.sh
+++ b/update.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -eo pipefail
 
 declare -A php_version=(


Stunkymonkey:
ping @tilosp ? any problem with this change?

Stunkymonkey:
@J0WI can you rerun the checks?

Stunkymonkey:
@J0WI the running tests are not consistent. Last time they were run different images failed. Not sure how to continue here. Any advice?

J0WI:
Yes, unfortunately the the database initialisation is very flaky. But I'm not sure if I'm gonna merge this anyway. Does ist fix any issue? I assume `/bin/bash` is more (POSIX?) compliant than `/usr/bin/env`.

Stunkymonkey:
you are assuming `bash` is always located in `/bin/bash`. For most Linux distributions this is true.
There are distributions around, that are not having this (NixOS, most BSDs, ...).
Using `env` has the advantage of running the script with a different version then `/bin/bash`. You might have a locally installed `~/.local/bin/bash`.
It is simply more portable.

Surely the scripts might work on your machine, but they might not work on others. Simply depends if you care about that.
If you do not want this change feel free to close it, but i would be happy if a decision is made.

J0WI:
Is `env` alsways in `/usr/bin`? The scripts are mainly run inside GitHub CI.

martadinata666:
So here is the thing about docker image, the image itself based on debian specifically the apache `/bin/bash` always there, doesn't matter your OS is, because thats how docker image, assuming you run on alpine that doesn't use bash, you can still use bash if the docker image include it.

Stunkymonkey:
> Is env alsways in /usr/bin?

as far as i know: yes

PR #1260, opened 2020-10-05
Allow custom script execution on installation

mgerhardy:
Should fix issues
- https://github.com/nextcloud/docker/issues/820
- https://github.com/nextcloud/docker/issues/1203

This allows a custom script to be mounted as docker or kubernetes volume to be executed. This can be used to execute all kind of occ commands.

Alternative and more customizable to the PR https://github.com/nextcloud/docker/pull/1259

PR #1656, opened 2021-12-08
Support setting NEXTCLOUD_AUTH_TOKEN

mrueg:
Allows to set up the nextcloud auth token during install.

Purpose is to make the install smoother by exposing it and have server
info and metrics available to verify the status of the install.

As this is only supported in v22 and later, for now a version check is
included.

See also:
https://github.com/nextcloud/serverinfo/pull/269


mrueg:
@J0WI can you take a look? I saw you were interested in this feature as well. :)

J0WI:
I'd prefer #1260

mrueg:
I agree that there might be additional need for custom scripts, I still feel like exposing a single environment variable like this would help ensuring that it is easy and standardized to work with in helm and other systems.

mrueg:
I've removed the version check since 21 is now gone.

mrueg:
@J0WI since #1260 seems to be stale, any chance to get this merged?

PR #1679, opened 2022-01-16
Simplify first character check in entrypoint.sh

mr700:
Simplify first character check in entrypoint.sh
use "${X:0:1}" instead of ""$(echo "$X" | cut -c1-1)"

Signed-off-by: Doncho N. Gunchev <dgunchev@gmail.com>

### DIFF
--- a/21/apache/entrypoint.sh
+++ b/21/apache/entrypoint.sh
@@ -57,7 +57,7 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
             file_env REDIS_HOST_PASSWORD
             echo 'session.save_handler = redis'
             # check if redis host is an unix socket path
-            if [ "$(echo "$REDIS_HOST" | cut -c1-1)" = "/" ]; then
+            if [ "${REDIS_HOST:0:1}" = "/" ]; then
               if [ -n "${REDIS_HOST_PASSWORD+x}" ]; then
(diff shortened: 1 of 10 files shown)


nupplaphil:
@J0WI @mr700 Is this change really working for the `/bin/sh` script?
I had troubles with it, because this syntax is only bash compatible

J0WI:
I've tested it on `busybox sh`.

mr700:
10x @J0WI, `busybox sh` and `busybox ash` support `${X:0:1}`. I was building my own images...

nupplaphil:
OK, I'm sorry that I stressed you, but I took over this part of code for my Docker image and ran into the regression 
`/entrypoint.sh: 52: Bad substitution`
(see https://github.com/friendica/docker/issues/196 )
and the test-case to reproduce it was:
```console
docker run --rm -it --entrypoint="/bin/sh" php:7.4-apache-bullseye
# export REDIS_HOST=friendica-redis.friendica.svc.cluster.local
# echo ${REDIS_HOST:0:1}
/bin/sh: 2: Bad substitution
```

kemoantemo:
@nupplaphil I built the apache version myself and when I try to use Redis it gives me error
image
and without redis it works normally
don't know if the problem is with me building wrong the only thing i did is clone the repo and cd to
`/docker/23/apache`
then
`docker build -t nextcloud:23.0.2 .`
and used this image
the problem not there with the official 23.0.1 image

PR #1681, opened 2022-01-25
Add support for Redis DB index

iamruinous:
I'd like to share a single redis instance with other services. This adds the ability to specify the redis db index to avoid conflicts.


J0WI:
In this case it's recommend to use separate redis containers

Issue #1699, opened 2022-02-18
entrypoint.sh issue in the new versions


kemoantemo:
@nupplaphil I built the apache version myself and when I try to use Redis it gives me error
![image](https://user-images.githubusercontent.com/19779701/154618173-7ce9150c-adc3-49f6-85a9-85eaab22f094.png)
and without redis it works normally
don't know if the problem is with me building wrong the only thing i did is clone the repo and cd to
`/docker/23/apache`
then
`docker build -t nextcloud:23.0.2 .`
and used this image
the problem not there with the official 23.0.1 image

edit...
Official 23.0.1 image uses
the old way I checked so I think that PR will break the apache image
![image](https://user-images.githubusercontent.com/19779701/154619815-399a13c2-b4a4-4a05-bf79-843986c4ce93.png)
the entrypoint.sh official 23.0.1

edit2...
Reverted the line to
`if [ "$(echo "$REDIS_HOST" | cut -c1-1)" = "/" ]; then`
and all is working
![Screenshot 2022-02-18 065921](https://user-images.githubusercontent.com/19779701/154620807-61676d8b-a7ed-49f9-8556-7567eac1c54a.png)
![Screenshot 2022-02-18 070052](https://user-images.githubusercontent.com/19779701/154620855-00f1d14e-9574-4c4c-9480-8e7141b97903.png)

So I think that pr should reverted

_Originally posted by @kemoantemo in https://github.com/nextcloud/docker/pull/1679#issuecomment-1043873137_

martadinata666:
So it seems we oversight the `/bin/sh` same everywhere
![2022-02-18-133440_490x95_scrot](https://user-images.githubusercontent.com/2433562/154630653-917c262a-6e2e-4418-8822-e6e3e1f7c057.png)
e

debian/ubuntu `sh` is `dash` like alpine `sh` was `ash`.

mr700:
Confirmed, dash `does` not support "${X:0:1}" in `php:8.0-fpm-bullseye` for the `fpm` and `apache` versions... will prepare revert for these.

mr700:
Reverted the whole commit, created pull request #1700 . No need to have different entrypoin.sh only for alpine.

PR #1700, opened 2022-02-18
Revert "Simplify first character check in entrypoint.sh (#1679)" #1699

mr700:
In php:8.0-fpm-bullseye /bin/sh is dash, which does not support "${X:0:1}"
(substring expansion).

This reverts commit 05365221755163f7f513bac939d4851e79b06a04.

For the alpine build /bin/ash supports this, but having different entrypoint.sh makes no sense, revert all.


mr700:
Fixing the commit message

Issue #1709, opened 2022-02-27
Examples Improvement about cron


Karem-sobhy:
after my issue https://github.com/nextcloud/server/issues/31373
and the fix was just to mount all volumes that are mounted in the app container to the cron container
thanks to https://github.com/nextcloud/server/issues/25852#issuecomment-997964401
so i wanted to suggest that all docker-compose examples
should be like that
```
cron:
    image: nextcloud:apache
    restart: always
    volumes_from:
      - app
    entrypoint: /cron.sh
    depends_on:
      - db
      - redis
```
instead of 
```
cron:
    image: nextcloud:apache
    restart: always
    volumes:
      - nextcloud:/var/www/html
    entrypoint: /cron.sh
    depends_on:
      - db
      - redis
```
I think that is more convenient to avoid such issues.

keunes:
Actually maybe `volumes_from` is not supported any-more in Docker 3?
https://github.com/nextcloud/docker/pull/1677#issue-1102312795

danieljkemp:
see also: https://docs.docker.com/compose/compose-file/compose-versioning/#upgrading

Karem-sobhy:
Closed because volumes from have compatibility issues.

PR #1710, opened 2022-02-27
Using volumes_from instead of volumes for cron in examples

kemoantemo:
Using volumes_from instead of volumes in the cron container to make sure it had access to all mounts that app container have and that will help new people that will be using docker cron to not have my problem. #1709 

Signed-off-by: Karem Sobhy <karemsobhy020@gmail.com>

### DIFF
--- a/.examples/docker-compose/insecure/mariadb/apache/docker-compose.yml
+++ b/.examples/docker-compose/insecure/mariadb/apache/docker-compose.yml
@@ -35,8 +35,8 @@ services:
   cron:
     image: nextcloud:apache
     restart: always
-    volumes:
-      - nextcloud:/var/www/html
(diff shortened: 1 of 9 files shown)


szaimen (approved):

martadinata666:
The only catch is `volumes_from` cant be use with `docker stack deploy`

kemoantemo:
> The only catch is `volumes_from` cant be use with `docker stack deploy`

Maybe it's not a problem because all new comers will likely use tutorials from the internet which all encourage the use of docker-compose with the examples from this repo. So I think that is more like advanced use case and in that case they are probably know what to do. that's only my opinion and I'm new to that I just did that PR because it would have solved my problem

PR #1711, opened 2022-03-01
Fix nextcloud mail Message-ID header

kamukvark:
See also: https://github.com/nextcloud/mail/issues/5764


J0WI:
The config is taken from here: https://github.com/nextcloud/documentation/blob/master/admin_manual/installation/nginx-root.conf.sample. Please open a PR there first.

kamukvark:
Done: https://github.com/nextcloud/documentation/pull/8111

PR #1718, opened 2022-03-23
Set recommend opcache.revalidate_freq

J0WI:


te-online:
Hey @J0WI 🙂 

Can you put a few words to how this change is beneficial? I have no expertise with opcache. The only thing I can say is that I see some issues in regular usage. I can't find a related issue?

Take updating an app as an example: It can take up to 60 seconds until all file changes have settled. In this timeframe we can get weird results from php files being cached. This happens to me on an Ubuntu 22.04 system.

It's especially more noticeable when doing app development with the container: I need to wait up to 60 seconds after I saved my php file until php picks up the change. This happens to me using macOS 12.

I've made a custom Dockerfile overriding this change back to `1` and I can't see any obvious performance penalty, just php files being updated immediately.

Would be happy to learn if I'm doing something wrong here or if I can do anything to improve things 😊

te-online:
If I look [at this comment](https://github.com/nextcloud/documentation/issues/1439#issuecomment-492173985), my understanding is that "the admin" in this case is you in extension of the docker-image users, pre-defining the config for many people using the docker image.

> If the admin knows what it does then we are fine with a different value. [different then 1, that is]

With my limited knowledge, using the `60` value is not a good or bad option, same as `1` is not inherently bad. For me with a pretty average Nextcloud instance it causes issues and does not increase performance. If there is an internal cache invalidation, it does not work. I have to wait 60 seconds after app updates for them to take effect. In the meantime the log might be flooded with errors, because frontend and backend won't necessarily match.

You might disagree with `1` as a value and I can live with that, but I just want to advertise for reverting this change as it seems unnecessary to me and seems only recommended if you like to have it set to this value and you – as an admin – are aware of the effects and can take them into account when administering your Nextcloud instance.

PR #1719, opened 2022-03-23
Remove Nextcloud 21 (EOL)

J0WI:
Signed-off-by: J0WI <J0WI@users.noreply.github.com>


mrueg:
Thanks for the follow-up PR. I think this commit should remove the 21/ folders as well?

PR #1728, opened 2022-04-11
Allow simultaneous initialization of common html volume

skjnldsv:
## Context
- Have multiple docker containers running behind a load balancer
- Share the html files directory (it helps for various reasons, one of it being the htaccess update needed in all containers otherwise)
- Start multiple containers at the same time

## Issue
- Rsync is complaining because files gets created/deleted from multiple sources simultaneously

## Solution
- Add a lock file to make sure only one container is allowed to init the shared html folder at the same time
- Lock file is shared for all containers within `/var/www/html`
- If html is not shared, then this will not be an issue and the lock file will be useless

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -101,92 +101,114 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
         else
             rsync_options="-rlD"
         fi
-        rsync $rsync_options --delete --exclude-from=/upgrade.exclude /usr/src/nextcloud/ /var/www/html/
 
-        for dir in config data custom_apps themes; do
-            if [ ! -d "/var/www/html/$dir" ] || directory_empty "/var/www/html/$dir"; then
-                rsync $rsync_options --include "/$dir/" --exclude '/*' /usr/src/nextcloud/ /var/www/html/
-            fi
-        done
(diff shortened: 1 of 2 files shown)


J0WI:
There's a syntax error, all tests are failing.

skjnldsv:
> There's a syntax error, all tests are failing.

Done :)

skjnldsv:
Ah damn, this is because of `/bin/sh` instead of `/bin/bash`

skjnldsv:
> Ah damn, this is because of `/bin/sh` instead of `/bin/bash`

Fixed!

skjnldsv:
@J0WI won't work on non-ubuntu based installs? Is that why you thumbs down? :)

J0WI:
`bash` is not installed on Alpine. Please use POSIX compatible syntax.

pierreozoux:
I'd like to review, but I changed my mind about putting the html in the container a long time ago.
I think it was a mistake.

On kubernetes we don't use it anymore and are really happy :)

I tried here without much success :upside_down_face:

skjnldsv:
> I'd like to review, but I changed my mind about putting the html in the container a long time ago.

I agree with you! This is also what I'm going forward with. There are a few issues related to stateless containers (like updating htaccess), but this is more or less a sanity check if people wanna do it.
There are multiple ways of addressing scalability :)

J0WI (approved):

PR #1730, opened 2022-04-13
Add mandatory summary step for branches protection

skjnldsv:
Will allow us to add a branch protection requiring all images build to be successful

EDIT: added protection, have a look at the output

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -14,7 +14,7 @@ defaults:
 
 jobs:
 
-  generate-jobs:
+  init:
     name: Generate Jobs
     runs-on: ubuntu-latest
     outputs:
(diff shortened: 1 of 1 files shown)


J0WI:
@skjnldsv I think the branch protection broke the [update.sh](https://github.com/nextcloud/docker/blob/master/.github/workflows/update-sh.yml) bot.

skjnldsv:
Possibly.
Honestly we should fix the bot then, not having protection in such a used repository is an issue imho :)

skjnldsv:
One solution is to open a pull request instead
https://github.com/nextcloud/.github/blob/d36c4842c33110848a92318d68caad89b9ec2c6d/.github/workflows/dispatch-workflow.yml#L61-L73

The other is to use a token that can push. But it also fall back to everyone having write permissions in that repository.

J0WI:
@skjnldsv right now no change in the template get applied to the relevant files. The Nextcloud update job is also failing and the new version was not pushed to this repo.

J0WI:
IMHO the permissions are now to restrictive. Neither the bot nor I can bump the Nextcloud version anymore. :confused:

skjnldsv:
disabled, this is far from ideal imho :)

PR #1757, opened 2022-05-21
use flock automatic lock for upgrade

### LINKED ISSUE #1756: use ephemeral instead of manual/persistent lock during upgrade
`docker-entrypoint.sh` manually creates a file and uses this to indicate an in-progress upgrade. Manual intervention is required if the upgrade fails halfway and `nextcloud-init-sync.lock` is never `rm`'d. How about if we use an automatic `flock` instead? There's example code in the `flock(1)` manual page:

```text
(
  flock -n 9 || exit 1
  # ... commands executed under lock ...
) 9>/var/lock/mylockfile
```

We should be able to just surround all upgrade code like that example, and get rid of `touch $lock`, `if [ -f "$lock" ]; then` and `rm $lock` in `docker-entrypoint.sh`.

Thoughts?

meonkeys:
fix #1756

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -102,14 +102,15 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
             rsync_options="-rlD"
         fi
 
+        # Prevent multiple images syncing simultaneously:
         # If another process is syncing the html folder, wait for
         # it to be done, then escape initalization
-        lock=/var/www/html/nextcloud-init-sync.lock
         count=0
         limit=10
 
-        if [ -f "$lock" ]; then
-            until [ ! -f "$lock" ] || [ "$count" -gt "$limit" ]
+        (
+        if ! flock -n 9; then
+            until flock -n 9 || [ "$count" -gt "$limit" ]
             do
                 count=$((count+1))
                 wait=$((count*10))
@@ -122,8 +123,6 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
             fi
             echo "The other process is done, assuming complete initialization"
         else
-            # Prevent multiple images syncing simultaneously
-            touch $lock
             rsync $rsync_options --delete --exclude-from=/upgrade.exclude /usr/src/nextcloud/ /var/www/html/
 
             for dir in config data custom_apps themes; do
@@ -211,10 +210,9 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
 
             fi
 
-            # Initialization done, reset lock
-            rm $lock
             echo "Initializing finished"
         fi
+        ) 9> /var/lock/nextcloud-init-sync.lock
     fi
 fi
 


skjnldsv (changes requested):
Doesn't seems to work for me.
If I run two simultaneous containers sharing the same html volume, both of them initialize at the same time. Instead of one waiting for the other to finish.

Also, I'm totally unfamiliar with flock. Is it available for alpine too?

martadinata666:
on alpine flock is a dependent package https://pkgs.alpinelinux.org/packages?name=flock&branch=v3.14
on debian based is from `util-linux` package, usually already included by default.

skjnldsv:
> on alpine flock is a dependent package [pkgs.alpinelinux.org/packages?name=flock&branch=v3.14](https://pkgs.alpinelinux.org/packages?name=flock&branch=v3.14)

[then it's an issue for the alpine container](https://github.com/nextcloud/docker/blob/master/Dockerfile-alpine.template)

meonkeys:
> Doesn't seems to work for me. If I run two simultaneous containers sharing the same html volume, both of them initialize at the same time. Instead of one waiting for the other to finish.

Is that the use case? I wasn't aware you could/would run more than one nextcloud container at a time.

chilli230:
i have the same issue with migrate nextcloud:22-2.7 to **22.2.8-apache**. I was delete the nextcloud-init-sync.lock file manualy, so it was work. On my Server running three Nextcloud instances parallel but with different volumes.

meonkeys:
@skjnldsv so I think this is the current state:

1. can we just add the flock package for the alpine image? Seems pretty low-risk
2. is running multiple Nextcloud app containers against the same db backend a use case we need to support? If so, I can look into it, but I'm guessing flock will not do the job. flock is also not usable with samba and NFS... I think there needs to be some OS-level interaction that is only supported on certain filesystems, and I'm guessing Docker filesystems (like overlay / union / whatever) are unsupported. Again, only an issue if multiple concurrent Nextcloud app containers (with a single database backend) must be supported.

skjnldsv:
> On my Server running three Nextcloud instances parallel but with different volumes.

Then this doesn't quite fit the use case :smile_cat: 

> Is that the use case? I wasn't aware you could/would run more than one nextcloud container at a time.

Yeah,sharing the same DB and data/config, you can fire as much containers as you want and load-balance in front, it's very effective :wink:

rugk:
So what again was the problem with this pr and why did you decide for that opt-in solution in the other PR instead?

meonkeys:
> So what again was the problem with this pr and why did you decide for that opt-in solution in the other PR instead?

@rugk I recall (based on the thread above):

1. the Alpine image didn't have `util-linux`
1. I was not sure if flock would work on NFS and samba, and
1. the solution by @skjnldsv fixed the problem for me because I only run one Nextcloud container (which might be an irrelevant detail anyway).

Ideas:

number 1: surely an easy fix, just add `util-linux` (if it hasn't already been added by now).

number 2: @remram44 says in #1903 that flock will work fine on NFS and samba.

number 3 is irrelevant

So that just leaves one problem I'm seeing now: I put `nextcloud-init-sync.lock` in `/var/lock/`. That file should be placed in `/var/html/` instead. That might be why it broke for @skjnldsv (see [their comment above](#pullrequestreview-984388967).

Heck, let's fix this thing! Seems like there is some interest to improve it.

PR #1784, opened 2022-07-08
php Memory limit issue while running as www-data in Docker Cron

mynksh:
https://github.com/nextcloud/docker/issues/1413


adripo:
Shouldn't `memory_limit` be the same as the one set in php config?

In my case by running `/bin/sh -c 'php -i' | grep memory_limit` it is 512MB

PR #1805, opened 2022-08-12
Add crond log level environment variable

MarcelCoding:
I couldn't find any documentation regarding the second cron entry point, I just found it in the stack.yaml.
I added an environment variable to configure the log level, the default case is logging every minute about 3 lines and I found that quite much - not everyone can configure it how he wants it. The default is `0` just as before.

Signed-off-by: Marcel <34819524+MarcelCoding@users.noreply.github.com>

### DIFF
--- a/docker-cron.sh
+++ b/docker-cron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -l 0 -L /dev/stdout
+exec busybox crond -f -l ${NEXTCLOUD_CRON_LOG_LEVEL:-0} -L /dev/stdout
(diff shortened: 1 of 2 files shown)


MarcelCoding:
Is this a review? I can't check out profile, if you are a maintainer.

joshtrichards:
@MarcelCoding Yes, @J0WI is a maintainer.

joshtrichards (commented):
> I'd rather change the default than introducing a new config. If someone needs customisation the script or command can always be overwritten.

I don't have a strong opinion either way. I lean towards changing the default outright too, for simplicity sake and because AFAIK the current verbose level was completely arbitrary (#220).

arge-oevv (commented):
Yes, this looks simple enough - 8 seems to be also the default for busybox crond.

joshtrichards (approved):
Works for me

PR #1812, opened 2022-09-01
Allow to run with custom uid

J0WI:
fix: #359, #772, #1081, #1087, #1278

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -13,7 +13,7 @@ directory_empty() {
 
 run_as() {
     if [ "$(id -u)" = 0 ]; then
-        su -p www-data -s /bin/sh -c "$1"
+        su -p "$user" -s /bin/sh -c "$1"
     else
         sh -c "$1"
     fi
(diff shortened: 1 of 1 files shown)


Ornias1993:
Afaik the redis section also needs patching @J0WI

namelivia:
Hello! I really appreciate this update, could you please explain briefly how to use it? I'm not super sure while reading the code. Should I just pass `uid` and `gid` as env variables to the container?

Chuckame:
Hello, this is not working with a fresh setup with apache, because of the redis stuff just before (permission denied during writing redis client params).
I it a problem if we `chown $user:$group` the folder `/usr/local/etc/php/conf.d` before ?

docker-entrypoint.sh
```diff
@@ -50,6 +50,29 @@ if expr "$1" : "apache" 1>/dev/null; then
...
 if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UPDATE:-0}" -eq 1 ]; then
+    uid="$(id -u)"
+    gid="$(id -g)"
+    if [ "$uid" = '0' ]; then
+        case "$1" in
+            apache2*)
+                user="${APACHE_RUN_USER:-www-data}"
+                group="${APACHE_RUN_GROUP:-www-data}"
+
+                # strip off any '#' symbol ('#1000' is valid syntax for Apache)
```
PhrozenByte, line 61: @J0WI Just skimming through this code and I'm wondering where this is coming from, because the base image apparently doesn't support this either (it might have in the past though)? See https://github.com/docker-library/php/blob/95a88177c3038064d1259b4fd2d4a0275201a6d7/Dockerfile-linux.template#L113. Or am I missing something?

g0r3:
Is this also supposed to work with Nexcloud AIO? I want to remap the `www-data` user to 1000 and tried using it with `docker run -d --user 33:1000 ...` but all I get is a container failed to start.

joshtrichards:
@g0r3 This repo/image set has nothing to do with AIO.

goooooouwa:
@joshtrichards is there a possibility to also port this change to AIO? AIO's nextcloud docker [entrypoint.sh](https://github.com/nextcloud/all-in-one/blob/main/Containers/nextcloud/entrypoint.sh) seems like sharing a lot of the same code.

PR #1814, opened 2022-09-02
enable mariadb auto upgrade

meonkeys:
This fixes after-upgrade issues, for example:

    [ERROR] Incorrect definition of table mysql.event: expected column 'definer' at position 3 to have type varchar(, found type char(141).

The upgrade only happens when necessary.

I also disabled the system database backup because I assume users create and test their own backups.


meonkeys:
Sure, will do.

Issue #1842, opened 2022-10-19
NEXTCLOUD_INIT_HTACCESS=true fails startup on a fresh install breaks startup


mabeett:
When starting nextcloud with `NEXTCLOUD_INIT_HTACCESS=true` on an fresh-empty volume the entrypoint exits whit this log

```
Nextcloud is not installed - only a limited number of commands are available

                                                                        
  There are no commands defined in the "maintenance:update" namespace.  
                                                                        
  Did you mean this?                                                    
      maintenance                                                       
                                                                        


```
checked with image tag `24.0.6-apache`.

PR #1843, opened 2022-10-19
prevents running `occ maintenance:update:htaccess` on a first run volume

mabeett:
Solves #1842


docker-entrypoint.sh
```diff
@@ -242,7 +242,7 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
     fi
 
     # Update htaccess after init if requested
-    if [ -n "${NEXTCLOUD_INIT_HTACCESS+x}" ]; then
+    if [ -n "${NEXTCLOUD_INIT_HTACCESS+x}" ] && [ "$installed_version" != "0.0.0.0" ]; then
```
mabeett, line 245: This strategy respect the feedback interaction in #1761

Issue #1233, opened 2020-09-01
All files are missing after pulling and updating nextcloud (version 19.0.1.1 to 19.0.2.2)


baschi29:
After running 

docker-compose stop
docker-compose pull
docker-compose up -d

all files seem to have gone missing and it's no longer possible to upload files to nextcloud. When checking the user directories (in my case nextdata/name of user) only the cache directory is still there, everything else seems to have been deleted.

Manually recreating the file directory and running occ files:scan "fixes" the problem as in that nextcloud is usable again and recreation from backup is possible.

Calendar and everything database related works as usual. Trying to open files through the nextcloud interface throws an error: - for example:

> 172.18.0.2 - - [01/Sep/2020:18:12:32 +0000] "GET /core/preview?fileId=25652&c=d6a2601d4a5e929f1a272c65ad31f249&x=250&y=250&forceIcon=0 HTTP/1.1" 404 729 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:80.0) Gecko/20100101 Firefox/80.0"

This happened the second time after pulling the latest nextcloud image. When not pulling I can stop and remove the container and the data is persistent. I hope somebody can help me out with this. I can reproduce this bug (or maybe misconfiguration on my site) as well to provide further logs.

My docker-compose file:

> version: '3.8'
> 
> volumes:
>   n_vhost:
>   n_html:
>   n_dhparam:
>   n_certs:
>   p_db:
>   nextcloud:
>   redis:
> 
> networks:
>   nextcloud_network:
> 
> services:
> 
>   nginx-proxy:
>     image: jwilder/nginx-proxy
>     container_name: nginx-proxy
>     restart: always
>     ports:
>       - "80:80"
>       - "443:443"
>     networks:
>       - nextcloud_network
>     volumes:
>       - /etc/nginx/conf.d:/etc/nginx/conf.d
>       - n_vhost:/etc/nginx/vhost.d
>       - n_html:/usr/share/nginx/html
>       - n_dhparam:/etc/nginx/dhparam
>       - n_certs:/etc/nginx/certs:ro
>       - /var/run/docker.sock:/tmp/docker.sock:ro
>     labels:
>       - "com.github.jrcs.letsencrypt_nginx_proxy_companion.nginx_proxy" 
> 
>   letsencrypt:
>     image: jrcs/letsencrypt-nginx-proxy-companion
>     container_name: nginx-proxy-le
>     restart: unless-stopped
>     networks:
[...]

PR #1862, opened 2022-11-20
Add words to README to draw attention to the use of upgrades.exclude, potential for files to be removed during installation/upgrades.

bsiddans:
During installation and upgrades files under the main folder may be removed if not listed in upgrade.exclude. Adding a note to the README, in the section discussing the use of persistent volumes, would help make this behaviour clearer and reduce the risk of accidental data loss.

Examples of situations where this change may provide benefit include:
- #1165
- #1528 
- #1233 

This PR addresses essentially the same issue as PR #1168 (and potentially replaces it). I've opened a new PR as I've proposed a rephrased explanation that I think is clearer about the options, and because the previous PR is outdated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,10 @@ $ docker run -d \
 -v theme:/var/www/html/themes/<YOUR_CUSTOM_THEME> \
 nextcloud
 ```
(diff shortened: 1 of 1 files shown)


bsiddans:
> I think this needs a rebase to fix the CI

Done.

PR #1872, opened 2022-12-01
Full ghostscript

J0WI:


Chartman123:
@J0WI will this also be added to the official Docker images or do I need to use an own image here?

PR #1881, opened 2022-12-09
Enable MariaDB binary log

meonkeys:
This resolves a warning in the database server log:

    [Warning] You need to use --log-bin to make --binlog-format work.

Pros:

* support for point-in-time recovery
* necessary for replication

Cons:

* slows down database operations ("slightly", per the manual)
* takes up disk space (mitigated by `--expire-logs-days=2`)

See also:

* <https://mariadb.com/kb/en/binary-log/>
* <https://mariadb.com/kb/en/full-list-of-mariadb-options-system-and-status-variables/>

Alternatives:

1. Do not add `--log-bin`. Remove `--binlog-format` instead. This causes the least amount of change for existing installations.

Signed-off-by: Adam Monsen <haircut@gmail.com>

### DIFF
--- a/.examples/docker-compose/insecure/mariadb/apache/docker-compose.yml
+++ b/.examples/docker-compose/insecure/mariadb/apache/docker-compose.yml
@@ -3,7 +3,7 @@ version: '3'
 services:
   db:
     image: mariadb:10.5
-    command: --transaction-isolation=READ-COMMITTED --binlog-format=ROW
+    command: --transaction-isolation=READ-COMMITTED --log-bin=binlog --expire-logs-days=2 --binlog-format=ROW
     restart: always
     volumes:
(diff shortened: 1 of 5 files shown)


meonkeys:
Thank you, I reviewed those sections of documentation. The first one recommends either using a row-level binary log or disabling the binary log. The second one (with the example `my.cnf`) indicates use of a row-level binary log.

My patch is mostly coherent with the documentation, and it fixes the MariaDB warning. I anticipate the impact of the change will be minimal and positive.

To better align with recommendations in the Nextcloud documentation, I'll remove `--expire-logs-days=2`. This isn't necessary. The MySQL and MariaDB documentation both state that binary logs will be purged on startup and flush/rotation anyway.

PR #1760, opened 2022-05-31
Add `NEXTCLOUD_INIT_LOCK` to enable shared html volume update

skjnldsv:
cc @meonkeys

Fix #1742 
Closes #1757 

## Easy reproducibility test

1. Clone this repo
2. Copy the `docker-entrypoint.sh` into `24/apache/entrypoint.sh`
3. Change current directory for `24/apache`
4. Edit the Dockerfile to
    ```Dockerfile
	FROM nextcloud:apache

	COPY *.sh upgrade.exclude /
	  
	ENTRYPOINT ["/entrypoint.sh"]
	CMD ["apache2-foreground"]
    ```
6. Build the image `docker build .` and save the built image hash (here `6e7615a878a6`)
    image
7. Create a dummy folder `mkdir /tmp/ncdocker`
8. Open two cosole and run this command at the same time in each shells (replace the hash by yours)
    ```bash
    docker run -e NEXTCLOUD_INIT_LOCK=true -e NEXTCLOUD_ADMIN_USER=admin \
      -e NEXTCLOUD_ADMIN_PASSWORD=admin -e SQLITE_DATABASE=true \
      -v /tmp/ncdocker:/var/www/html 6e7615a878a6
    ```
9. See one of the two processes waiting for the other to complete

| Without `NEXTCLOUD_INIT_LOCK` | With `NEXTCLOUD_INIT_LOCK=true` |
|:---------:|:------:|
|image|image|

PR #1905, opened 2023-01-27
Use flock for the upgrade lock

### LINKED ISSUE #1903: Use flock for the upgrade lock
In #1757, an attempt was made at replacing the locking mechanism for upgrade (when running multiple containers) from simply creating a file to using Linux advisory locks with flock. Admittedly the patch had a few flaws.

Eventually the MR was closed in favor of making the locking optional with an environment variable in #1760. This is a very different approach, but it resolves the issue you had at the time, #1742.

I am opening this issue to gauge whether the is still interest in using flock rather than the mere presence of the file, e.g. did you close #1757 because you don't want to do that, or simply because it was not ready and another patch was able to fix the pressing upgrade issue in #1742?

Pros of using flock:

* If a container crashes or is stopped during upgrade, another container can do the upgrade. Right now if locking is enabled, you will have to manually go into the volume and remove the file.
* A concurrent container waiting to do the upgrade will be able to get the lock immediately, rather than the next time it "polls" after sleeping for a while.

Cons of using flock:

* If a container crashes during upgrade, another container can do the upgrade. I don't know if it is safe to resume an aborted upgrade (but at the same time, what else can you do?)
* It requires the volume to support locks (but even NFS, Cephfs, ... support it, so you'd have to use a weirdo non-POSIX-compliant filesystem)
* A bit more complexity in the entrypoint script

remram44:
Fixes #1903

This keeps the `NEXTCLOUD_INIT_LOCK` environment variable, though I don't know if it is still required. Without it, I don't need to factor `do_install_or_upgrade()` (because it's not called twice).

Locking around the entire upgrade process allows the container to check the version at the right time (so it can tell whether the upgrade is needed, e.g. it might have failed).

I think there is still a problem that the file copy might have completed but `occ upgrade` might not have finished, but I'm not sure how to detect that. The simplest way is to just always run `occ upgrade`, which succeeds immediately if there's nothing to do.

### DIFF
--- a/23/apache/entrypoint.sh
+++ b/23/apache/entrypoint.sh
@@ -43,127 +43,6 @@ file_env() {
     unset "$fileVar"
 }
 
-do_install_or_upgrade() {
-    installed_version="0.0.0.0"
-    if [ -f /var/www/html/version.php ]; then
-        # shellcheck disable=SC2016
-        installed_version="$(php -r 'require "/var/www/html/version.php"; echo implode(".", $OC_Version);')"
-    fi
-    # shellcheck disable=SC2016
-    image_version="$(php -r 'require "/usr/src/nextcloud/version.php"; echo implode(".", $OC_Version);')"
-
-    if version_greater "$installed_version" "$image_version"; then
-        echo "Can't start Nextcloud because the version of the data ($installed_version) is higher than the docker image version ($image_version) and downgrading is not supported. Are you sure you have pulled the newest image version?"
-        exit 1
-    fi
-
-    if version_greater "$image_version" "$installed_version"; then
-        echo "Initializing nextcloud $image_version ..."
-        if [ "$installed_version" != "0.0.0.0" ]; then
-            echo "Upgrading nextcloud from $installed_version ..."
-            run_as 'php /var/www/html/occ app:list' | sed -n "/Enabled:/,/Disabled:/p" > /tmp/list_before
-        fi
-        if [ "$(id -u)" = 0 ]; then
-            rsync_options="-rlDog --chown $user:$group"
-        else
-            rsync_options="-rlD"
-        fi
-
-        rsync $rsync_options --delete --exclude-from=/upgrade.exclude /usr/src/nextcloud/ /var/www/html/
-        for dir in config data custom_apps themes; do
-            if [ ! -d "/var/www/html/$dir" ] || directory_empty "/var/www/html/$dir"; then
-                rsync $rsync_options --include "/$dir/" --exclude '/*' /usr/src/nextcloud/ /var/www/html/
-            fi
-        done
-        rsync $rsync_options --include '/version.php' --exclude '/*' /usr/src/nextcloud/ /var/www/html/
-
-        # Install
-        if [ "$installed_version" = "0.0.0.0" ]; then
-            echo "New nextcloud instance"
-
-            file_env NEXTCLOUD_ADMIN_PASSWORD
-            file_env NEXTCLOUD_ADMIN_USER
-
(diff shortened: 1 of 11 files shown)


meonkeys:
1. Ooh, good call to put the lock in `/var/www/html`, I should have done that. I bet that's why #1757 [failed when @skjnldsv  tested it](https://github.com/nextcloud/docker/pull/1757#pullrequestreview-984388967).
2. If we use `flock`, `NEXTCLOUD_INIT_LOCK` is no longer necessary.
3. Nitpick (apologies in advance): the patch is hard to read because so many lines have changes. I tried ignoring whitespace but that didn't help. I'll try viewing the patch locally with other tools, just curious if there's any way to reduce the number of lines changed so it is more readable.

remram44:
I know the patch is big, it's because I had to move some code into a function above. I broke that down into separate commits to help with review, 7a0aba93a2229933513bac23ca424942917219a2 is nothing else but the move so you can read the rest.

If we remove the `NEXTCLOUD_INIT_LOCK` variable the diff will be much shorter.

meonkeys:
What do you think of #1917 ? I updated that to only change from using a file to using `flock`, and incorporated your placement of the lockfile in `/var/www/html` instead of `/var/lock`. I think we should do the refactoring separately. Unless I'm missing something and it is required to use flock... it just looks functionally/logically separate at first glance.

remram44:
The only reason for the factoring is to support `NEXTCLOUD_INIT_LOCK`, ie this part:

https://github.com/nextcloud/docker/blob/295cdf7fb710f005050d9b894e4f814c96f2c068/docker-entrypoint.sh#L226-L237

If there is no condition there is no need to factor into a function at all.

remram44:
I removed the `NEXTCLOUD_INIT_LOCK` condition from here (and the now-useless function).

The difference at this point is that I grab the lock earlier, before we load the version information, and that I show a message if the lock can't be acquired immediately. I also updated the README about the removed environment variable.

meonkeys:
Makes sense. I just needed some help understanding it, thank you. Great work! Seems like all upside to widen the scope of the locked region to include the read (version check) and write (install/upgrade) operations.

Let's test this and save the maintainers some time. See acceptance test steps at https://github.com/nextcloud/docker/pull/1760#issue-1253460157

meonkeys:
LGTM. I ran the test mentioned above on 25/apache and 25/fpm-alpine, both worked.

:shipit:

remram44:
Is the CI failing something I have to deal with? I can't find a specific error from it.

meonkeys:
@skjnldsv sorry to be a pest -- would you mind reviewing this? LGTM, FWIW

skjnldsv:
@meonkeys unfortunately I am lacking available time.
Have you manually tested this ? If so can you both take care of the reviewing?
If confirmed everything is in order, I can approve and merge for you

meonkeys:
@skjnldsv I've reviewed and tested the patch following the method in <https://github.com/nextcloud/docker/pull/1760#issue-1253460157>.

@remram44 will you test this too?

remram44:
Locks should work on any filesystems that claims to be "POSIX" or similar. I have tested on CephFS and it works :heavy_check_mark:, NFS/GlusterFS/SeaweedFS/JuiceFS/Lustre should work too.

remram44:
Locks should work on any filesystems that claims to be "POSIX" or similar. I have tested on CephFS and it works :heavy_check_mark:, NFS/GlusterFS/SeaweedFS/JuiceFS/Lustre should work too.

meonkeys:
@skjnldsv sounds like we're good to go!

skjnldsv:
Restarted CI

PR #1917, opened 2023-02-08
try 2: use flock automatic lock for upgrade

meonkeys:
This is #1757, revisited.

Only this *or* #1905 should be merged (if either are merged). Both PRs intend a change from (a) treating `nextcloud-init-sync.lock`'s existence to indicate an upgrade in progress to (b) using the `flock` shell command (either GNU's or Busybox's). Switching to (b) provides an atomic and ephemeral (automatically cleaned-up) lock.

### DIFF
--- a/23/apache/entrypoint.sh
+++ b/23/apache/entrypoint.sh
@@ -124,15 +124,15 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
             rsync_options="-rlD"
         fi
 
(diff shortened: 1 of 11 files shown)


meonkeys:
Sorry for all the force-pushes. I'm done editing now.

meonkeys:
We might want to rename `nextcloud-init-sync.lock` so folks aren't confused by old comments suggesting they manually delete that file to fix a stalled upgrade.

meonkeys:
Argh! I lied, one more force push (forgot to use `--signoff`)

meonkeys:
Let's use #1905 instead.

PR #1933, opened 2023-03-03
Add s3 storage class

### LINKED ISSUE #1932: Add S3 object store storage class option
## User Story

When using S3-compatible object stores, choosing the storage class for objects is an important consideration. By tailoring the storage class to the application requirements there can be massive cost savings.

As a small-time user I want to be able to choose a cost-efficient storage class for light yet bursty usage (intelligent tiering). I do not want to mount additional files / add custom configurations to the base Docker image. I want to interact with this setting the same as with all other similar settings - by environment variable.

## Proposed solution

Recently, this feature appears to have been implemented into nextcloud/server via https://github.com/nextcloud/server/pull/36075. The goal of this issue is to add it to the swath of environment variable settings available for the Docker image as described in the README.

<blockquote>
To use an external S3 compatible object store as primary storage, set the following variables:

    OBJECTSTORE_S3_HOST: The hostname of the object storage server
    OBJECTSTORE_S3_BUCKET: The name of the bucket that Nextcloud should store the data in
    OBJECTSTORE_S3_KEY: AWS style access key
    OBJECTSTORE_S3_SECRET: AWS style secret access key
    OBJECTSTORE_S3_PORT: The port that the object storage server is being served over
    OBJECTSTORE_S3_SSL (default: true): Whether or not SSL/TLS should be used to communicate with object storage server
    OBJECTSTORE_S3_REGION: The region that the S3 bucket resides in.
    OBJECTSTORE_S3_USEPATH_STYLE (default: false): Not required for AWS S3
    OBJECTSTORE_S3_LEGACYAUTH (default: false): Not required for AWS S3
    OBJECTSTORE_S3_OBJECT_PREFIX (default: urn:oid:): Prefix to prepend to the fileid
    OBJECTSTORE_S3_AUTOCREATE (default: true): Create the container if it does not exist

Check the [Nextcloud documentation](https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/primary_storage.html#simple-storage-service-s3) for more information.
</blockquote>

Following the naming convention and [current default storage class of STANDARD](https://github.com/nextcloud/server/pull/36075/files#diff-32333c316cf51e92bd3b53808f0fe1eca1635fb1f62daba1469ce6694959f97cR87), let's go with

```
OBJECTSTORE_S3_STORAGE_CLASS (default: STANDARD): The storage class to apply to all stored objects
```

## Alternative Considerations

- How will this behave for S3 compatible stores not S3? Do they all support it? In any case, that is probably out of scope of this issue. We just need to interact with the setting that has already been implemented.

## Notes

- available from Nextcloud 26
- currently undocumented

aentwist:
- [ ] documentation
- [ ] tests

Closes #1932


.config/s3.config.php
```diff
@@ -14,6 +14,7 @@
         'region' => getenv('OBJECTSTORE_S3_REGION') ?: '',
         'hostname' => getenv('OBJECTSTORE_S3_HOST') ?: '',
         'port' => getenv('OBJECTSTORE_S3_PORT') ?: '',
+        'storageClass' => getenv('OBJECTSTORE_S3_STORAGE_CLASS') ?: 'STANDARD',
```
aentwist, line 17: Set a default here instead of relying on the upstream default to decouple

README.md
```diff
@@ -172,6 +172,7 @@ To use an external S3 compatible object store as primary storage, set the follow
 - `OBJECTSTORE_S3_PORT`: The port that the object storage server is being served over
 - `OBJECTSTORE_S3_SSL` (default: `true`): Whether or not SSL/TLS should be used to communicate with object storage server
 - `OBJECTSTORE_S3_REGION`: The region that the S3 bucket resides in.
+- `OBJECTSTORE_S3_STORAGE_CLASS` (default: `STANDARD`): The storage class to apply to all objects stored. See S3 [`--storage-class`](https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html#options).
```
aentwist, line 175: Assuming the options are the same as those of the AWS CLI `aws s3 cp --storage-class`...

.config/s3.config.php
aentwist, line 17: Also note the unfortunately inconsistent naming convention. Soft reviews in upstream. This looks correct, going off of https://github.com/nextcloud/server/pull/36075/files#diff-dcd9e132a7c9faaa92d88afa9b90b299cc367bae128909d6c4c00f1c20952d94R50

aentwist:
I wanted to rerun the pipeline since the checks were failing to see if they would pass. I did not realize that GitHub requires this to be approved.

The checks are not failing as a result of my work. It seemed from the output that they could not authenticate with the database? I am not very familiar with GitHub forks / CI. Maybe this is the cause?

> Although workflows from forks do not have access to sensitive data such as secrets

https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks#about-workflow-runs-from-public-forks

Please advise...

README.md
aentwist, line 175: needs note - from Nextcloud 26

README.md
```diff
@@ -172,6 +172,7 @@ To use an external S3 compatible object store as primary storage, set the follow
 - `OBJECTSTORE_S3_PORT`: The port that the object storage server is being served over
 - `OBJECTSTORE_S3_SSL` (default: `true`): Whether or not SSL/TLS should be used to communicate with object storage server
 - `OBJECTSTORE_S3_REGION`: The region that the S3 bucket resides in.
+- `OBJECTSTORE_S3_STORAGE_CLASS` (since Nextcloud 26, default: `STANDARD`): The storage class to apply to all objects stored. See S3 [`--storage-class`](https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html#options).
```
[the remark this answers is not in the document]
aentwist: True, although the docs should just have a link pointing to Amazon documentation (e.g., this link or a link to the storage classes page, which sadly does not enumerate the options).

aentwist:
> Does this break any non-AWS implementation?

This is a great question and one I was curious about as well. We could be cautious here and only set the config option if the env var exists (instead of always setting it and using a default value if it doesn't exist). However, from what I can tell, note that the nextcloud/server implementation does not do this -

setting a default value if does not exist

https://github.com/nextcloud/server/pull/36075/files#diff-32333c316cf51e92bd3b53808f0fe1eca1635fb1f62daba1469ce6694959f97cR87

always sending it no matter what

https://github.com/nextcloud/server/pull/36075/files#diff-030e38a9e621019981acf91654c1d8d3a12ab3150dbd9fcc4c280479b8c853dbR108

If you want to let the handling of does not exist / default values fall-through to nextcloud/server I will make the change here. This is definitely the main design choice.

If you are wondering whether it is compatible with non-S3.. well, that is probably a question for https://github.com/nextcloud/server/pull/36075. Hopefully since by definition these APIs should be 'S3-compatible' everything is ok. Probably they just ignore the param if they don't support it..

joshtrichards:
> If you want to let the handling of does not exist / default values fall-through to nextcloud/server I will make the change here. This is definitely the main design choice.

`server` defaults safely to `standard` if it's not specified. So we should let `server` handle... What it's already handling. The default is unlikely to change upstream, but... No reason to create a possible future clash. And there's no upside/downside I can see to adding another layer here of unnecessary defaults in the image.

> If you are wondering whether it is compatible with non-S3.. well, that is probably a question for nextcloud/server#36075. 

Agreed. Not an issue for the image to mess with since server is already setting a default. And particularly if we're not setting our own here. :-)

README.md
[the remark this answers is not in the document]
joshtrichards: Yes, link to the NC docs if anything. Though in this case I'd drop the link entirely. Those messing with storageClass already know why/what they're doing by the time they get to configuring their Docker image.

However if it's left in and the official docs need further updating that can be addressed in the `documentation` repo separately. Not a dependency for this PR.

I still say just remote the link entirely.

.config/s3.config.php
```diff
@@ -12,6 +12,7 @@
...
```
joshtrichards, line 15: Why do we need to decouple storageClass? Already defaults to standard [upstream](https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/primary_storage.html#simple-storage-service-s3):

> storageClass defaults to STANDARD

README.md
```diff
@@ -176,6 +176,7 @@ To use an external S3 compatible object store as primary storage, set the follow
...
```
joshtrichards, line 179:
```suggestion
- `OBJECTSTORE_S3_STORAGE_CLASS` (since Nextcloud 26, default: `STANDARD`): The storage class to apply to all objects stored.
```

There's already a link to the NC Server S3 documentation in that section. Let's leave it to the upstream documentation to further document (or maintain links to) all the possible S3 setting variations (if need be).

joshtrichards (changes requested):

pathob:
@aentwist Any chance to move this forward?

aentwist:
@pathob No, my respect for Nextcloud wore out a long time ago. By all means feel free to work on this; it is an incredibly small change and might take 5 minutes.

pathob:
@aentwist Since you mention it - respect was the reason to ask first whether you want to complete this 5 minute change before submitting it myself. :)

PR #1936, opened 2023-03-10
set display_errors to Off

individual-it:
This PR provents errors to be put in the HTTP response.

The errors should not be displayed on production systems. E.g. the [PHP docs](https://www.php.net/manual/en/errorfunc.configuration.php#ini.display-errors) says: "This is a feature to support your development and should never be used on production systems (e.g. systems connected to the internet)."

Besides of security implications, having the errors displayed makes it impossible to catch some problems e.g. exceeding max_post_size.

### DIFF
--- a/23/apache/Dockerfile
+++ b/23/apache/Dockerfile
@@ -108,6 +108,7 @@ RUN { \
         echo 'memory_limit=${PHP_MEMORY_LIMIT}'; \
         echo 'upload_max_filesize=${PHP_UPLOAD_LIMIT}'; \
         echo 'post_max_size=${PHP_UPLOAD_LIMIT}'; \
+        echo 'display_errors=Off'; \
     } > "${PHP_INI_DIR}/conf.d/nextcloud.ini"; \
     \
     mkdir /var/www/data; \
--- a/23/fpm-alpine/Dockerfile
+++ b/23/fpm-alpine/Dockerfile
@@ -96,6 +96,7 @@ RUN { \
         echo 'memory_limit=${PHP_MEMORY_LIMIT}'; \
         echo 'upload_max_filesize=${PHP_UPLOAD_LIMIT}'; \
         echo 'post_max_size=${PHP_UPLOAD_LIMIT}'; \
+        echo 'display_errors=Off'; \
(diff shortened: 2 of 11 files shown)


individual-it:
@J0WI this seems not to work on the docker container.

How to reproduce:

1. start the containers `docker compose up -d` with this docker compose file:
    [docker-compose.zip](https://github.com/nextcloud/docker/files/10989278/docker-compose.zip)
    the limit is set as: `PHP_UPLOAD_LIMIT=1024`
2. setup NC: `docker compose exec --user www-data nextcloud php occ maintenance:install -vvv  --database pgsql --database-name nextcloud --database-host db --database-user nextcloud  --database-pass nextcloud --admin-user admin --admin-pass admin --data-dir /var/www/html/data`
3. login as `admin` to http://localhost:8080
4. change the profile image to something bigger than 1024B

The response is: 
HTTP Status: 200
content:
```
<br />
<b>Warning</b>:  POST Content-Length of 5442962 bytes exceeds the limit of 1024 bytes in <b>Unknown</b> on line <b>0</b><br />
Config file has leading content, please remove everything before &quot;&lt;?php&quot; in config.php<br />
<b>Fatal error</b>:  Uncaught Error: Call to a member function getLogger() on null in /var/www/html/index.php:71
Stack trace:
#0 {main}
  thrown in <b>/var/www/html/index.php</b> on line <b>71</b><br />
```

so its clearly does display the errors

individual-it:
@J0WI I don't think the `config.php` is the problem, I haven't touched it and uploading a smaller file works totally fine.

I've chosen 1024 as the limit only to make it easy to reproduce. If I set the limit to e.g. 20MB and upload a big picture the result is the same: 
```
Warning: POST Content-Length of 24413437 bytes exceeds the limit of 20971520 bytes in Unknown on line 0
Config file has leading content, please remove everything before "<?php" in config.php
Fatal error: Uncaught Error: Call to a member function getLogger() on null in /var/www/html/index.php:71 Stack trace: #0 {main} thrown in /var/www/html/index.php on line 71
```

I'm happy to open an issue in the nextcloud server repo, but is there any harm to add that config value in the docker container?

Issue #1938, opened 2023-03-13
Add env vars for SMTP secrets


aentwist:
Usernames are part of credentials and are therefore also secrets. Add an env var for SMTP username file so that `SMTP_NAME` can be used with Docker secrets.

aentwist:
Yes, are you sure you aren't thinking of `SMTP_HOST`?

Here is a good article I found, which appears to be based on RFCs - https://www.samlogic.net/articles/smtp-commands-reference-auth.htm

There exists both a username and password.

As an example, when you generate a key for AWS / SES it looks something like [keyboard mashing]

key: AOREHNK3UN34OU3OORS
secret: o23un4ouenowfenjh3on3/33ehnjo3uk

Imagine trying to hack something where the username is admin vs. something where it is instead roiehduyowaouwdhowuyhd but you don't know that. This is some kind of credential, and even if it is less sensitive than the secret, in my opinion it should be able to be a secret if users want it to be. The goal of this issue is to allow users to make that choice conveniently.

aentwist:
> Nextcloud itself does not treat mail_smtpname as a sensitive

Well that is just unfortunate.. thanks for looking into it.

I am ok with that point as long as it has parity with other usernames. i.e. database username is not treated as sensitive.

aentwist:
workaround: get the updated smtp config file from the MR and mount it over the existing smtp config file at /usr/src/nextcloud/config/smtp.config.php to support `SMTP_NAME_FILE`. if you don't mount over it then it will be overridden by the existing smtp.config.php, so you have to provide a replacement for the entire file.

PR #1942, opened 2023-03-15
Refactor env vars from file

### LINKED ISSUE #1940: Improve environment variable from file support
The environment variable from file support is currently very fragmented on the PHP side.

https://github.com/nextcloud/docker/blob/31c59aea24f8516f862ca136766168016e840593/.config/autoconfig.php#L9-L37

This is extremely fragile, as it is all or nothing with the `_FILE` environment variables - and to complicate things, one of them is non-`_FILE`.

https://github.com/nextcloud/docker/blob/31c59aea24f8516f862ca136766168016e840593/.config/smtp.config.php#L15-L21

---

These should also be refactored to use a common function `f(string) -> string` that returns the environment variable value given its name and a default (no matter whether it is from file or not). There is already a parallel idea implemented in the entrypoint script -

https://github.com/nextcloud/docker/blob/31c59aea24f8516f862ca136766168016e840593/docker-entrypoint.sh#L22-L26

---

Related to https://github.com/nextcloud/docker/issues/1938

aentwist:
This PR depends on all 'environment variable from file' changes. What is included in it vs. what is included in other open pull requests depends on the order in which they are merged. I suggest merging the feature branches first, and then making changes to this refactor pull request as needed. I would be happy to bring this up-to-date after the other PRs are merged.

Note that as a result of this pull request, the database configuration environment variables will become independent. That is, for example, `POSTGRES_USER`, `POSTGRES_PASSWORD`, and `POSTGRES_DB` may be defined to be from file or not independently. No longer will there be confusion over defining some but not all from file.

- [x] documentation
- [ ] testing

Related to #1938, #1945

Closes #1940

### DIFF
--- a/.config/autoconfig.php
+++ b/.config/autoconfig.php
@@ -1,37 +1,25 @@
 <?php
 
+require_once "util.php";
+
 $autoconfig_enabled = false;
 
 if (getenv('SQLITE_DATABASE')) {
     $AUTOCONFIG['dbtype'] = 'sqlite';
     $AUTOCONFIG['dbname'] = getenv('SQLITE_DATABASE');
     $autoconfig_enabled = true;
-} elseif (getenv('MYSQL_DATABASE_FILE') && getenv('MYSQL_USER_FILE') && getenv('MYSQL_PASSWORD_FILE') && getenv('MYSQL_HOST')) {
+} elseif (getFileEnv('MYSQL_DATABASE') && getFileEnv('MYSQL_USER') && getFileEnv('MYSQL_PASSWORD') && getenv('MYSQL_HOST')) {
     $AUTOCONFIG['dbtype'] = 'mysql';
-    $AUTOCONFIG['dbname'] = trim(file_get_contents(getenv('MYSQL_DATABASE_FILE')));
-    $AUTOCONFIG['dbuser'] = trim(file_get_contents(getenv('MYSQL_USER_FILE')));
-    $AUTOCONFIG['dbpass'] = trim(file_get_contents(getenv('MYSQL_PASSWORD_FILE')));
+    $AUTOCONFIG['dbname'] = getFileEnv('MYSQL_DATABASE');
+    $AUTOCONFIG['dbuser'] = getFileEnv('MYSQL_USER');
+    $AUTOCONFIG['dbpass'] = getFileEnv('MYSQL_PASSWORD');
     $AUTOCONFIG['dbhost'] = getenv('MYSQL_HOST');
     $autoconfig_enabled = true;
-} elseif (getenv('MYSQL_DATABASE') && getenv('MYSQL_USER') && getenv('MYSQL_PASSWORD') && getenv('MYSQL_HOST')) {
-    $AUTOCONFIG['dbtype'] = 'mysql';
-    $AUTOCONFIG['dbname'] = getenv('MYSQL_DATABASE');
-    $AUTOCONFIG['dbuser'] = getenv('MYSQL_USER');
-    $AUTOCONFIG['dbpass'] = getenv('MYSQL_PASSWORD');
-    $AUTOCONFIG['dbhost'] = getenv('MYSQL_HOST');
-    $autoconfig_enabled = true;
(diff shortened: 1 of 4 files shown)


.config/autoconfig.php
```diff
@@ -1,38 +1,26 @@
 <?php
 
+require_once "util.php";
```
aentwist, line 3: Not familiar with the autoloading for this or how to test it. Point me somewhere to get started and I'll handle it.

.config/util.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+$FILE_ENV_VAR_SUFFIX = "_FILE";
```
aentwist, line 5: Could move this into function scope to avoid polluting the global scope, if that is a thing

aentwist, line 5: Could move this into function scope to avoid polluting the global scope, if that is a thing

.config/autoconfig.php
```diff
@@ -1,38 +1,26 @@
...
+
 $autoconfig_enabled = false;
 
-if (getenv('SQLITE_DATABASE')) {
+if (getConfigFromEnv('SQLITE_DATABASE')) {
```
aentwist, line 7: database config from env also needs supported in the entrypoint script too. sqlite db isn't in there yet.

.config/autoconfig.php
```diff
@@ -1,38 +1,26 @@
...
-} elseif (getenv('MYSQL_DATABASE_FILE') && getenv('MYSQL_USER_FILE') && getenv('MYSQL_PASSWORD_FILE') && getenv('MYSQL_HOST')) {
-    $AUTOCONFIG['dbtype'] = 'mysql';
-    $AUTOCONFIG['dbname'] = trim(file_get_contents(getenv('MYSQL_DATABASE_FILE')));
-    $AUTOCONFIG['dbuser'] = trim(file_get_contents(getenv('MYSQL_USER_FILE')));
-    $AUTOCONFIG['dbpass'] = trim(file_get_contents(getenv('MYSQL_PASSWORD_FILE')));
-    $AUTOCONFIG['dbhost'] = getenv('MYSQL_HOST');
+    $AUTOCONFIG['dbname'] = getConfigFromEnv('SQLITE_DATABASE');
     $autoconfig_enabled = true;
-} elseif (getenv('MYSQL_DATABASE') && getenv('MYSQL_USER') && getenv('MYSQL_PASSWORD') && getenv('MYSQL_HOST')) {
+} elseif (getConfigFromEnv('MYSQL_DATABASE') && getConfigFromEnv('MYSQL_USER') && getConfigFromEnv('MYSQL_PASSWORD') && getConfigFromEnv('MYSQL_HOST')) {
```
aentwist, line 11: database config from env also needs supported in the entrypoint script too. for mysql and postgres hosts, this should just be a matter of adding a line that runs them through the existing `file_env` function.

.config/util.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Gets the value of an environment variable or the contents of the file
+ * at the path specified by its value if it exists with a suffix of "_FILE"
+ */
+function getConfigFromEnv(string $envVarName, ?string $defaultValue): ?string {
```
aentwist, line 9: Alternative naming: `getFileEnv`

.config/autoconfig.php
aentwist: As for the first part, are you suggesting that I try to have `getFileEnv` included by renaming the file to x.config.php? Or simply forget the whole idea?

---

As for the database part -

In the entrypoint script those values can independently be from file or not.

https://github.com/nextcloud/docker/blob/405e81528528f7f5a40242ca181ebdfe835e12e7/docker-entrypoint.sh#L162-L164

The improvement here isn't that we allow some to go missing, it is that this PR just so happens to allow the same thing in the PHP. That is,

current behavior (as described in the README section removed by this PR)

- must have postgres_db, _user, _password, _host OR
- postgres_db_file, _user_file, _password_file, _host

updated behavior

- must have one of postgres_db or postgres_db_file AND
- _user or user_file AND
- _password or _password_file AND
- _host

Just noting that any potential confusion from this (e.g. documented by https://github.com/nextcloud/docker/issues/1148#issuecomment-991697378) is additionally avoided by this PR.

Issue #1785, opened 2022-07-09
Obscure way to use unix socket for Redis


raycheung:
The parameter passing causes problems from the Docker container environment all the way to the phpredis->pconnect when using _unix socket_. The phpredis is kind of requiring the **port** setting to -1 if the **host** is actually a unix socket path starting with '/'. However in this image, the entrypoint.sh defaults REDIS_HOST_PORT to 6379, which would eventually passing into phpredis.

This would work though:
```
environment:
      REDIS_HOST: /tmp/redis.sock
      REDIS_HOST_PORT: -1
```

joshtrichards:
Good catch!

I didn't initially encounter this issue because I *switched* to UNIX socket use in my instances after they were already live. And when I did so I only modified config.php on the nc-app servers (where port 0 is specified). That is, I never added a REDIS_HOST to my Docker compose when I transitioned.

I encountered this problem though when I cleaned up my compose files.

The issue heres are two-fold:

1. If REDIS_HOST is specified in compose, than the config.php is always overridden with whatever is specified in it (via redis.config.php). And that'd be fine, but the way that redis.config.php currently does things, if a UNIX socket is specified than $CONFIG['redis']['port'] is simply left as null (which doesn't match the documented NC configuration parameters for sockets which specify a port 0): https://docs.nextcloud.com/server/25/admin_manual/configuration_server/caching_configuration.html#id2

2. phpredis's handling of socket connections and internal code for settling upon the default port was a little whacky. This was fixed in https://github.com/phpredis/phpredis/pull/2187 but that's *not* in any released version of phpredis including v5.3.7 (and that's what the NC docker image uses).

These two situations combined mean that REDIS_HOST_PORT is *not* an optional parameter for UNIX socket connections. And, 
worse, it's a non-sensical parameter because there is _no_ proper port to specify for a UNIX socket connection.

Your workaround - specifying the optional REDIS_HOST_PORT fixes things (either 0 or -1 appear to function, but not specifying anything leaves things broken; and I suspect any value other than null would work since it'll ultimately be a noop for a UNIX socket connection).

Given that the documented NC configuration parameters say to specify port 0 in config.php for UNIX sockets (which works in a standard NC install, but not via the docker container - unless one explicitly sets REDIS_HOST_PORT in compose) I suggest a small change to redis.config.php to set the port to 0 if a UNIX socket is specified as the REDIS_HOST and no port is provided (and there generally won't be/shouldn't be).

[...]

PR #1968, opened 2023-04-16
Guarantee REDIS_HOST_PORT is never null for UNIX sockets

joshtrichards:
Addresses #1785 (which arises when using UNIX sockets)

Without this fix, UNIX socket connections won't work if REDIS_HOST is used to specify the socket to the container unless an additional bogus REDIS_HOST_PORT is also specified.

This change makes the auto-generated REDIS socket configuration in the Docker container (which overrides whatever is specified in config.php whenver REDIS_HOST is used) match NC's recommended default (which includes specifying port 0 for sockets). More importantly it's more user friendly and eliminates non-working setups with error messages such as the following (which results in an Internal Server Error):

```
Deprecated: Redis::pconnect(): Passing null to parameter #2 ($port) of type int is deprecated in /var/www/html/lib/private/RedisFactory.php on line 137
```

P.S. I believe this is the only file in the docker repository that needs to be updated manually - my reading is that update.sh is used as part of the release process and will update the ones for each version/config model appropriately. If I'm incorrect let me know and I'll update those too. Should all be the same.

### DIFF
--- a/.config/redis.config.php
+++ b/.config/redis.config.php
@@ -13,5 +13,7 @@ if (getenv('REDIS_HOST')) {
     $CONFIG['redis']['port'] = (int) getenv('REDIS_HOST_PORT');
   } elseif (getenv('REDIS_HOST')[0] != '/') {
     $CONFIG['redis']['port'] = 6379;
+  } elseif (getenv('REDIS_HOST')[0] == '/') {
+    $CONFIG['redis']['port'] = 0;
   }
 }


PhrozenByte:
We should probably rather fix this in `nextcloud/server` and drop `$CONFIG['redis']['port']` if the user specifies no port, simply because default values should be applied at a single place (and `nextcloud/server` already defaults to port 6379 if no port is configured).

So, can you please open a PR against `nextcloud/server` setting `$port = 0` as default in [`lib/private/RedisFactory.php`](https://github.com/nextcloud/server/blob/fa23698c23a7ef9380f13c4fe439ff3aa6a9b9b0/lib/private/RedisFactory.php#L117) and then change this PR to drop `$CONFIG['redis']['port']` if the user specifies no port? Please don't forget to sign-off your commits (`git commit -s`, see [here](https://github.com/nextcloud/server/blob/master/.github/CONTRIBUTING.md#sign-your-work)) and please also run `./update.sh`.

PR #1978, opened 2023-05-10
Fix auto config file

flortsch:
This commit fixes the auto config file which has the following problems leading to Nextcloud not picking up the set options:
* autoconfig.php as a file name is not picked up by Nextcloud (it expects files ending with .config.php)
* the file sets a variable named $AUTOCONFIG, whereas it should be named $CONFIG for setting the values
* the option dbpass must be named dbpassword
* the option directory must be named datadirectory


PhrozenByte:
What issue is this PR fixing? Nextcloud autoconfig purposely affects the install process only and isn't meant for runtime configuration. See https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/automatic_configuration.html

flortsch:
I am sorry, please ignore this PR. I didn't know about the autoconfig feature in Nextcloud and thought that this config file is very specific to the Docker setup and the used Docker env vars.

PR #1994, opened 2023-05-29
Prefer compose config over building containers

oliv3r:
There doesn't seem to be a need to prefer building custom containers,
just to add a config file, a need that is easily solved with compose
configs. Configs are still managed through git, just no new container
needs to be built, slightly simplifying things.

While here, fix a cron-race-condition with regards the app container.


oliv3r:
@J0WI fair nuff; but it's 5 years later ...

PR #1996, opened 2023-05-29
autoconfig: Allow mixing _FILE variables

oliv3r:
We do not always need to configure variables as either a file or not for all variables.

Instead, allow us to mix-match variable storage. This allows us to keep all non-critical variables in regular (env) variables, but keep the password in a (secret) file only location.

By offering this change, we make it easier for the end-user, that does not have to know that this was an either/or proposition.


joshtrichards:
Related: #1942

PR #2043, opened 2023-07-31
Apache `LimitRequestBody 0`

### LINKED ISSUE #1796: Apache images built after ~mid June 2022 introduce 1GB upload limit
After my last docker image update I suddenly saw upload failures for very large files, even though I hadn't changed any configuration, and the nextcloud docker build also has no related configuration changes. I finally tracked the problem down to a change in Apache default configuration.

Steps to reproduce:
 1. Start with nextcloud apache docker image built before June 9, 2022
 2. Check included Apache version (is 2.4.53)
 3. Create large testfile (`dd if=/dev/zero of=testfile bs=1M count=1025`)
 4. Create upload share in nextcloud
 5. `curl -X PUT -H "X-Requested-With: XMLHttpRequest" -T  testfile http://<shareid>:<sharepass>@<host>/public.php/webdav/testfile`
 6. Update docker image to newest version (Apache is now 2.4.54)
 7. `curl -X PUT -H "X-Requested-With: XMLHttpRequest" -T  testfile http://<shareid>:<sharepass>@<host>/public.php/webdav/testfile2`

Actual result:
 Step 5 works, Step 7 doesn't work. A smaller testfile (`dd if=/dev/zero of=testfile bs=1M count=1024`) will still work with the new image.

Root cause: https://httpd.apache.org/docs/current/en/mod/core.html#limitrequestbody has changed default value in Apache version 2.4.53 from 0 (unlimited) to 1073741824.  Note: This also means the the various documentation entries about increasing PHP_UPLOAD_LIMIT etc. don't apply.

IMHO this is a regression, and the apache nextcloud docker image should change the builtin Apache configuration back to the old value of 0 (unlimited).

thechubbypanda:
Removes the 1GB upload restriction, allowing the PHP setting to work again.

Thanks to @t-schuster for the patch suggestion

Fixes #1796

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -29,7 +29,7 @@ declare -A base=(
 )
 
 declare -A extras=(
-	[apache]='\nRUN a2enmod headers rewrite remoteip ;\\\n    {\\\n     echo RemoteIPHeader X-Real-IP ;\\\n     echo RemoteIPTrustedProxy 10.0.0.0/8 ;\\\n     echo RemoteIPTrustedProxy 172.16.0.0/12 ;\\\n     echo RemoteIPTrustedProxy 192.168.0.0/16 ;\\\n    } > /etc/apache2/conf-available/remoteip.conf;\\\n    a2enconf remoteip\\\n    { echo LimitRequestBody 0 ; } > /etc/apache2/conf-available/limitrequestbody.conf;\\\n    a2enconf limitrequestbody'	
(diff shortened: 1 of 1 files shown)


Etienne-Schmitt:
Can a maintaner validate this? 
This is a real problem and having to mount a special config for apache or edit the .htaccess everytime work BUT is not as practical than this fix

t-schuster:
I still apply the same patch in my deployment, it is actually as simple as that.

thechubbypanda:
We've all read the docs and come to the conclusion that this patch is literally the one thing stopping the PHP env variables from working fully. It is actually something that used to work until apache changed the default value for it.
Regardless, it is merely allowing for one less hurdle to jump through when trying to enable big file uploads.
I'd also point out that the docs are terrible. It took me 5 days to figure out what was going on and find this fix.

Etienne-Schmitt:
Can confirm this, I'm using this only fix on my container and just this fixed the problem.

Like mention earlier in older version of apache2 the default value for LimitRequestBody was 0 before now it's set as 1GiB, as you can see here:
https://httpd.apache.org/docs/2.4/mod/core.html#limitrequestbody

Maybe for some people there will be more problems, like badly configured reverse proxy or else, but on the images library/nextcloud:*-apache "flavor" it will fix this upload problem.

Etienne-Schmitt:
It even was discussed on the AOI images maintain by nextcloud:
https://github.com/nextcloud/documentation/pull/10989

So, can we have progress on this?

thechubbypanda:
Closing in favour of #2065

PR #2073, opened 2023-09-24
fix erroneous env variable name

### LINKED ISSUE #2072: example docker-compose files contain wrong environment variable
The docker-compose.yml files in `.examples/docker-compose/insecure/` (both apache and fpm) contain the environment variable `MYSQL_ROOT_PASSWORD` which will not work as the database (mariadb) expects the value in `MARIADB_ROOT_PASSWORD`.

To fix this:
Change the variable name from `MYSQL_ROOT_PASSWORD` to `MARIADB_ROOT_PASSWORD` in the following files:
- `.examples/docker-compose/insecure/apache/docker-compose.yml`
- `.examples/docker-compose/insecure/fpm/docker-compose.yml`

lucascherzer:
Fixes #2072 

Short rundown: The env variable `MYSQL_ROOT_PASSWORD` does not work with MariaDB and has to be changed to `MARIADB_ROOT_PASSWORD`

I only tested this with the `docker-compose/insecure/mariadb/apache` compose file but as they all rely on the same images, it should not make a difference and work on all now.

### DIFF
--- a/.examples/docker-compose/insecure/mariadb/apache/docker-compose.yml
+++ b/.examples/docker-compose/insecure/mariadb/apache/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     volumes:
       - db:/var/lib/mysql:Z
     environment:
-      - MYSQL_ROOT_PASSWORD=
+      - MARIADB_ROOT_PASSWORD=
       - MARIADB_AUTO_UPGRADE=1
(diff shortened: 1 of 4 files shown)


joshtrichards:
https://github.com/nextcloud/docker/issues/2072#issuecomment-1745107988

PR #2077, opened 2023-10-02
Remove legacy header X-Download-Options in nginx.conf examples

am97:
The header was relevant for Internet Explorer only and was removed from the nextcloud/server code: https://github.com/nextcloud/server/commit/ea0e45d81e6963dab8c89981538f9d5fe2d51472


kesselb (approved):
Thank you

PR #2080, opened 2023-10-04
add ffmpeg to docker build

zooper:


zooper:
@meonkeys yes it was my intention from the beginning but I ended up pushing my changes to the wrong repo. :man_facepalming:

PR #2084, opened 2023-10-13
making NEXTCLOUD_TRUSTED_DOMAINS after installation

melvin-suter:
ALL configurations should be made with environment variables instead of manually editing configurations.
The ENV variable is already there but ignored after installation. This change will allow trusted domains to be configured over ENV variables, while still considering the configured values inside config.php.


kesselb:
Hi, 

- The new config file should go to https://github.com/nextcloud/docker/tree/master/.config
- Run generate-stackbrew-library.sh to populate the change

melvin-suter:
> Hi,
> 
> * The new config file should go to https://github.com/nextcloud/docker/tree/master/.config
> * Run generate-stackbrew-library.sh to populate the change

Changed it. Thanks.

.config/trusteddomains.config.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+if (getenv('NEXTCLOUD_TRUSTED_DOMAINS')) {
+  // Get originaly configured values
+  include(__DIR__.'/config.php');
```
kesselb, line 5: That's evil, don't do that.

melvin-suter: The `$CONFIG` is overwritten so that shouldn't be an issue. But if anyone does interesting things in config.php it could break things, you are right. I did this to make it backwards compatible, so people can still use their values manually configured in config.php (or with occ). Any ideas how to make this happen?

melvin-suter: @kesselb any ideas?

kesselb: Sorry, no ideas. But it still looks like a bad idea to include the actual configuration here again.

.config/trusteddomains.config.php
```diff
@@ -0,0 +1,11 @@
...
+
+  // Add environment variables
+  $CONFIG = array(
+    'trusted_domains' => array_merge($CONFIG['trusted_domains'],array_map(function($domain){return trim($domain);},explode(",",getenv('NEXTCLOUD_TRUSTED_DOMAINS'))))
```
utshomax, line 9: @melvin-suter I think the 'NEXTCLOUD_TRUSTED_DOMAINS' is separated as mentioned in the docker doc.

.config/trusteddomains.config.php
markusnasholm: The environment variable should override trusted_domains in config.php. But since this will be a change in behavior from earlier, when it was only set during the installation, I can see the issue of breaking a lot of existing setups. My suggestion is to create a new variable for this. Something like just `$TRUSTED_DOMAINS`, which will also be more in line with other env variables like `$TRUSTED_PROXIES` which do not use the `NEXTCLOUD_` prefix. Then you could just do this:
```
$trustedDomains = getenv('TRUSTED_DOMAINS');
if ($trustedDomains) {
  $CONFIG['trusted_domains'] = array_filter(array_map('trim', explode(' ', $trustedDomains)));
}
```
and nothing should break unless the user explicitly sets the new variable. The old `$NEXTCLOUD_TRUSTED_DOMAINS` should then perhaps be deprecated.

Issue #1896, opened 2023-01-12
Collabora Online - Built-in CODE Server in Nextcloud docker. Documents do not load 


Japhys:
I am using Docker behind nginx proxy manager and everything works fine. I installed Collabora Online - Built-in CODE Server  and Nextcloud office and that also seems fine. But opening an document does not work, it keeps spinning.

```
version: '2'

volumes:
  nextcloud:
  db:

services:
  db:
    image: mariadb:10.5
    restart: always
    command: --transaction-isolation=READ-COMMITTED --binlog-format=ROW
    volumes:
      - /srv/nextcloud/db:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=X
      - MYSQL_PASSWORD=X
      - MYSQL_DATABASE=X
      - MYSQL_USER=X
  app:
    image: nextcloud
    restart: always
    ports:
      - 8080:80
    links:
      - db
    volumes:
      - /srv/nextcloud/nextcloud/:/var/www/html
    environment:
      - MYSQL_PASSWORD=X
      - MYSQL_DATABASE=X
      - MYSQL_USER=X
      - MYSQL_HOST=db
      - PHP_MEMORY_LIMIT=4096M
      - PHP_UPLOAD_LIMIT=4096M
```

![2023-01-12_20-21](https://user-images.githubusercontent.com/24695937/212161707-19c5cca8-5ac0-4dbf-84d4-aea1d4b1c699.png)

both on desktop and mobile

![2023-01-12_20-24](https://user-images.githubusercontent.com/24695937/212161890-728b6651-3f8a-4df6-bf0d-4fa5b4902472.png)

Any idea's?

Japhys:
The log says this, no sure what to do


```
Error | richdocuments | GuzzleHttp\Exception\ConnectException: cURL  error 28: Operation timed out after 45000 milliseconds with 0 bytes  received (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for  https://example.com/custom_apps/richdocumentscode/proxy.php?req=/hosting/capabilities
-- | -- | --
```

Japhys:
I also tried a different docker image (linuxserver.io) on a different local server without nginx proxy manager. The exact same thing happens. Driving me nuts. Is Collabora built in app with nextcloud office simply not working with docker? Does anyone successfully use this at all?

martadinata666:
[...]

PR #2089, opened 2023-10-24
(README) Add note about Alpine based image incompatibility

### LINKED ISSUE #1041: docker-compose examples: alpine versions of Nextcloud containers doesn't work with Onlyoffice document_server app
Hi,
I've been trying to use the __onlyoffice__ `document_server` app with my `docker-compose` deployment based on the examples from this repository, and it seems that the document_server app doesn't work with __alpine__ containres as some binaries (like `x2t`) require a `glibc` and the __alpine__  comes with the `musl` libc.
It would be good to mention this somewhere in the doc.

In order to fix onlyoffice on my deployment I had to switch back to non alpine versions (for the nextcloud containers).

The principal source of info I found with regards this is this comment by @siccovansas: https://github.com/nextcloud/documentserver_community/issues/10#issuecomment-599009799

I've some more details on this discourse post: https://help.nextcloud.com/t/strugging-integrating-onlyoffice-with-nextcloud-fpm-alpine-mariadb-redis-cron-lets-encrypt-docker-compose/66639/2

So happy to now have onlyoffice working, thanks for all this work!

joshtrichards:
Alpine based images are not supported by OnlyOffice or Nextcloud Office.

Closes #1041 + some mingled in reports elsewhere like:
- https://github.com/nextcloud/docker/issues/1896#issuecomment-1601117063

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,12 @@ $ docker run -d nextcloud:fpm
 
 As the fastCGI-Process is not capable of serving static files (style sheets, images, ...), the webserver needs access to these files. This can be achieved with the `volumes-from` option. You can find more information in the [docker-compose section](#running-this-image-with-docker-compose).
 
+### Alpine image variant
+
+Alpine variants of the fpm image are available by appending `-alpine` to the tag - e.g. `nextcloud:stable-fpm-alpine`.
+
+**Beware that Alpine variants can have incompatibilities with some applications**. For example, Alpine images do not meet the [system requirements for OnlyOffice](https://github.com/nextcloud/documentserver_community#requirements) nor [Nextcloud Office (CODE/Collabora)](https://github.com/CollaboraOnline/richdocumentscode#system-requirements) due to Alpine's use of *musl libc*.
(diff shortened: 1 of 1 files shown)


joshtrichards:
Correct. Which, unfortunately, makes this is relevant to usage of the *Collabora Online - Built-in CODE Server* (aka: `richdocumentscode`) with the alpine variants of our fpm image. And it's an officially Recommended app + in the Hub bundle pushed at setup/install time by NC...

https://github.com/nextcloud/server/blob/23e5fe9a516a2befdeb59eb553bcc5149312db04/core/src/components/setup/RecommendedApps.vue#L86-L115

https://github.com/nextcloud/server/blob/23e5fe9a516a2befdeb59eb553bcc5149312db04/lib/private/App/AppStore/Bundles/HubBundle.php#L42-L47

joshtrichards:
The `recognize` app would be another (though it at least has a fallback slow mode):

- https://github.com/nextcloud/recognize/#requirements

PR #2098, opened 2023-11-07
Add `OBJECTSTORE_S3_SSE_C_KEY` environment variable to use S3 SSE-C encryption (server side encryption)

jessebot:
Thanks to everyone who maintains this repo!

I recently learned I could pass in env vars for using S3 as the primary object store, but I was missing the encryption key.

This PR would add an environment variable called `OBJECTSTORE_S3_SSE_C_KEY` to the docs and if set, it will be used to provide the `sse_c_key` argument to the `$CONFIG` array in s3.config.php. If the user doesn't provide `OBJECTSTORE_S3_SSE_C_KEY`, we don't set `ss3_c_key`, because it has no default.

It's mentioned in the docs [here](https://docs.nextcloud.com/server/27/admin_manual/configuration_files/primary_storage.html?highlight=object#s3-sse-c-encryption-support):

> Nextcloud supports server side encryption, also known as [SSE-C](http://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html), with compatible S3 bucket provider. The encryption and decryption happens on the S3 bucket side with a key provided by the Nextcloud server.

> The key can be specified with the sse_c_key parameter which needs to be provided as a base64 encoded string with a maximum length of 32 bytes. A random key could be generated using the the following command:

```bash
openssl rand 32 | base64
```

### DIFF
--- a/.config/s3.config.php
+++ b/.config/s3.config.php
@@ -23,6 +23,10 @@ if (getenv('OBJECTSTORE_S3_BUCKET')) {
     )
   );
 
+  if (getenv('OBJECTSTORE_S3_SSE_C_KEY')) {
+    $CONFIG['objectstore']['arguments']['sse_c_key'] = getenv('OBJECTSTORE_S3_SSE_C_KEY');
+  }
+
   if (getenv('OBJECTSTORE_S3_KEY_FILE') && file_exists(getenv('OBJECTSTORE_S3_KEY_FILE'))) {
     $CONFIG['objectstore']['arguments']['key'] = trim(file_get_contents(getenv('OBJECTSTORE_S3_KEY_FILE')));
   } elseif (getenv('OBJECTSTORE_S3_KEY')) {
--- a/README.md
+++ b/README.md
@@ -180,6 +180,7 @@ To use an external S3 compatible object store as primary storage, set the follow
 - `OBJECTSTORE_S3_LEGACYAUTH` (default: `false`): Not required for AWS S3
 - `OBJECTSTORE_S3_OBJECT_PREFIX` (default: `urn:oid:`): Prefix to prepend to the fileid
 - `OBJECTSTORE_S3_AUTOCREATE` (default: `true`): Create the container if it does not exist
+- `OBJECTSTORE_S3_SSE_C_KEY`: Server side encryption key - must be a base64 encoded string with a maximum length of 32 bytes
 
 Check the [Nextcloud documentation](https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/primary_storage.html#simple-storage-service-s3) for more information.
 


joshtrichards:
Hi @jessebot - Thanks for including the matching README update. :-) 

This variable should get `_FILE` ([Docker secrets](https://github.com/nextcloud/docker#docker-secrets)) support too if we're going to do it. Feel up for adding it? Can be modeled after the others at the same spot in the code.

README.md
```diff
@@ -180,6 +180,7 @@ To use an external S3 compatible object store as primary storage, set the follow
 - `OBJECTSTORE_S3_LEGACYAUTH` (default: `false`): Not required for AWS S3
 - `OBJECTSTORE_S3_OBJECT_PREFIX` (default: `urn:oid:`): Prefix to prepend to the fileid
 - `OBJECTSTORE_S3_AUTOCREATE` (default: `true`): Create the container if it does not exist
+- `OBJECTSTORE_S3_SSE_C_KEY`: Server side encryption key - must be a base64 encoded string with a maximum length of 32 bytes
```
jessebot, line 183:
```suggestion
- `OBJECTSTORE_S3_SSE_C_KEY`: _Optional_ Server side encryption key - must be a base64 encoded string with a maximum length of 32 bytes
```

.config/s3.config.php
```diff
@@ -23,6 +23,12 @@
     )
   );
 
+  if (getenv('OBJECTSTORE_S3_SSE_C_KEY_FILE') && file_exists(getenv('OBJECTSTORE_S3_SSE_C_KEY_FILE'))) {
+    $CONFIG['objectstore']['arguments']['sse_c_key'] = trim(file_get_contents(getenv('OBJECTSTORE_S3_SSE_C_KEY_FILE'));
+  } elseif (getenv('OBJECTSTORE_S3_SSE_C_KEY')) {
+    $CONFIG['objectstore']['arguments']['sse_c_key'] = getenv('OBJECTSTORE_S3_SSE_C_KEY');
+  }
+
```
pathob, line 31: IMHO this block should go to the end of the file since `OBJECTSTORE_S3_KEY_FILE` and `OBJECTSTORE_S3_SECRET_FILE` have much more relevance than this.

README.md
pathob: The existing wording in the README is `(not set by default)` - not `_Optional_`. Could you please add it to your PR?

pathob (commented):
Hi @jessebot , I'm not a maintainer here, but highly interested in this change, so I would like to help moving it forward. I've added two suggestions that might help get this merged. It would be great if you could make these two small adjustments. Thank you!

.config/s3.config.php
pathob: This block is also missing the `else` case...

.config/s3.config.php
```diff
@@ -23,6 +23,12 @@
     )
   );
 
+  if (getenv('OBJECTSTORE_S3_SSE_C_KEY_FILE') && file_exists(getenv('OBJECTSTORE_S3_SSE_C_KEY_FILE'))) {
+    $CONFIG['objectstore']['arguments']['sse_c_key'] = trim(file_get_contents(getenv('OBJECTSTORE_S3_SSE_C_KEY_FILE'));
+  } elseif (getenv('OBJECTSTORE_S3_SSE_C_KEY')) {
```
pathob, line 28: I've tried to run it... there is a third closing bracket `)` missing at the end

pathob (changes requested):

.config/s3.config.php
```diff
@@ -23,6 +23,12 @@
     )
   );
 
+  if (getenv('OBJECTSTORE_S3_SSE_C_KEY_FILE') && file_exists(getenv('OBJECTSTORE_S3_SSE_C_KEY_FILE'))) {
+    $CONFIG['objectstore']['arguments']['sse_c_key'] = trim(file_get_contents(getenv('OBJECTSTORE_S3_SSE_C_KEY_FILE'));
```
pathob, line 27: I've tried to run it... there is a third closing bracket `)` missing at the end

pathob (changes requested):

PR #2120, opened 2023-12-13
Warn if config folder misses original image configs or files differ

flortsch:


docker-entrypoint.sh
```diff
@@ -273,6 +273,19 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
...
 
+    # check if config files of user installation differ from image
+    for cfgPath in /usr/src/nextcloud/config/*.php; do
+        cfgFile=$(basename "$cfgPath")
+
+        if [ "$cfgFile" != "config.sample.php" ]; then
+            if ! cmp -s "/usr/src/nextcloud/config/$cfgFile" "/var/www/html/config/$cfgFile"; then
+                echo "Warning: your config folder is missing the file $cfgFile that comes with this Docker image or it differs from the original file."
+                echo "This could cause certain Docker configuration environment variables to be ignored or not work as expected."
+                echo "Please copy /usr/src/nextcloud/config/$cfgFile into /var/www/html/config/ to ensure that all Docker configuration environment variables work as intended."
```
joshtrichards, line 284: Maybe: 

```suggestion
                echo "This can happen if you've previously upgraded since configuration files stored on persistent volumes are not overwritten automatically."
                echo "To fix: copy /usr/src/nextcloud/config/$cfgFile into /var/www/html/config/ to ensure that all Docker configuration environment variables work as intended."
```

flortsch:
Hey @joshtrichards! Sound good. I can extend the PR with a section in the readme and a link to it in the console output. Since you mentioned warnings in the Nextcloud admin overview: is it possible to generate such UI warnings in the entrypoint script? I think that warnings in the UI would be a nice addition to the console output which can often be overlooked.

flortsch:
@joshtrichards Please check if the text and link is ok. I updated the README and added a link in the docker-entrypoint echo statement. Also, I inlined your last text feedback.

docker-entrypoint.sh
```diff
@@ -273,6 +273,20 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
         fi
     ) 9> /var/www/html/nextcloud-init-sync.lock
 
+    # check if config files of user installation differ from image
+    for cfgPath in /usr/src/nextcloud/config/*.php; do
+        cfgFile=$(basename "$cfgPath")
+
+        if [ "$cfgFile" != "config.sample.php" ]; then
+            if ! cmp -s "/usr/src/nextcloud/config/$cfgFile" "/var/www/html/config/$cfgFile"; then
+                echo "Warning: your config folder is missing the file $cfgFile that comes with this Docker image or it differs from the original file which could cause certain Docker configuration environment variables to be ignored or not work as expected."
```
joshtrichards, line 282:
```suggestion
                echo "***WARNING:*** Your config folder is missing the file $cfgFile that comes with this Docker image or it differs from the original file which could cause certain Docker configuration environment variables to be ignored or not work as expected."
```

Came from seeing issues like [this]( https://github.com/linuxserver/docker-nextcloud/issues/400) in the ls.io image where they have some similar warnings, but it being overlooked even though it's likely the cause of the reported issue (outdated nginx configs in this case) :-)

joshtrichards:
> Since you mentioned warnings in the Nextcloud admin overview: is it possible to generate such UI warnings in the entrypoint script? I think that warnings in the UI would be a nice addition to the console output which can often be overlooked.

Not currently.

docker-entrypoint.sh
```diff
@@ -273,6 +273,20 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
...
+        if [ "$cfgFile" != "config.sample.php" ]; then
+            if ! cmp -s "/usr/src/nextcloud/config/$cfgFile" "/var/www/html/config/$cfgFile"; then
+                echo "***WARNING***: Your config folder is missing the file $cfgFile that comes with this Docker image or it differs from the original file which could cause certain Docker configuration environment variables to be ignored or not work as expected."
+                echo "This can happen if you've previously upgraded since configuration files stored on persistent volumes are not overwritten automatically."
+                echo "To fix: copy /usr/src/nextcloud/config/$cfgFile into /var/www/html/config/ to ensure that all Docker configuration environment variables work as intended."
+                echo "Also see: https://github.com/nextcloud/docker#auto-configuration-and-nextcloud-updates"
+            fi
+        fi
+    done
+
```
joshtrichards, line 289:
```suggestion
    # check if at least one config file on persistent storage differs from running image
    for cfgPath in /usr/src/nextcloud/config/*.php; do
        cfgFile=$(basename "$cfgPath")

        if [ "$cfgFile" != "config.sample.php" ]; then
            if ! cmp -s "/usr/src/nextcloud/config/$cfgFile" "/var/www/html/config/$cfgFile"; then
		    outOfDateCfgFiles=1
		    break # we stop at detecting just one out of date file here so that we can generate the full list later (because arrays are a PITA w/o bash)
            fi
        fi
    done
    if [ $outOfDateCfgFiles -eq 1 ]; then # i.e. if some out of date or missing config files were noted let's try to inform the operator
		echo "*** WARNING ***"
		echo "*** The image-specific config files on your persistent volume differ from those that ship with the running image."
		echo "*** This could lead to behavior that differs from that documented for the image (such as auto configuration variables not working as expected)."
                echo "*** Don't panic! This can happen if you've ever upgraded your running container since we do not overwrite configuration files stored on persistent volumes after initial installation for your safety."
		echo "*** The following config files differ from that expected by the running image: "
    		for cfgPath in /usr/src/nextcloud/config/*.php; do
        	    cfgFile=$(basename "$cfgPath")

        	    if [ "$cfgFile" != "config.sample.php" ]; then
            	        if ! cmp -s "/usr/src/nextcloud/config/$cfgFile" "/var/www/html/config/$cfgFile"; then
				echo "**** /var/www/html/config/$cfgFile (versus the latest version for this image found at /usr/src/nextcloud/config/$cfgFile)"
            	    	fi
        	    fi
    		done
                echo "*** Details and instructions for resolving can be found at https://github.com/nextcloud/docker#auto-configuration-and-nextcloud-updates"
		echo "*** In short, compare the contents of /usr/src/nextcloud/config/ with /var/www/html/config/."
    fi

```

joshtrichards (commented):
Finally got a chance to test this a bit. Made some proposed adjustments. Let me know what you think.

flortsch:
Hey @joshtrichards! I included your suggestions. Only fixed the formatting and the uninitialized variable access. Tested locally with docker-compose. Missing or changed image-specific files trigger the warning & file listing. Custom config files don't trigger anything. Commits squashed and rebased on latest master.

joshtrichards:
@J0WI What do you think about this? 

I think a warning when the configs are out-of-date is a good step forward without introducing a breaking change. It's also something easy to check (or request the reporter check) in the container logs whenever a new bug comes in is unknowingly bogus due to out-of-date config files.

The other option is we take the full leap: documenting all the image `config/` files (names) as being "special" and exclusive to the image. Then we update them automatically whenever needed. That'll cause it's own problems (at least in the short-term) IMO.

We can always take the bigger leap later on after we add the warning. So I think this is a good compromise.

flortsch:
I reverted the code to simply print a warning line for every found config file mismatch.

docker-entrypoint.sh
```diff
@@ -273,6 +273,41 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
...
+        echo "*** This could lead to behavior that differs from that documented for the image (such as auto configuration variables not working as expected)."
+        echo "*** Don't panic! This can happen if you've ever upgraded your running container since we do not overwrite configuration files stored on persistent volumes after initial installation for your safety."
+        echo "*** The following config files differ from that expected by the running image:"
+
+        for cfgPath in /usr/src/nextcloud/config/*.php; do
+            cfgFile=$(basename "$cfgPath")
+
+            if [ "$cfgFile" != "config.sample.php" ]; then
+                if ! cmp -s "/usr/src/nextcloud/config/$cfgFile" "/var/www/html/config/$cfgFile"; then
+                    echo "**** /var/www/html/config/$cfgFile (versus the latest version for this image found at /usr/src/nextcloud/config/$cfgFile)"
```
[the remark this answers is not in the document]
flortsch: I reverted the code to simply print a warning line for every found config file mismatch.

flortsch:
@joshtrichards I reverted the changes and simplified the code to just print a one-line warning for every config file that differs. What do you think?

joshtrichards:
Thanks, @flortsch for getting this started and hanging in there! :-)

joshtrichards:
First sighting in the wild: https://help.nextcloud.com/t/question-on-some-warnings-on-outdated-files-since-update-to-29-0-3/196297

Maybe the docs link should get added back in :-)

flortsch:
I myself got this warning today after updating to 28.0.7 :rofl: Will double-check my configs, but it seems the check fulfills its purpose :D If the confusion gets to big I can provide a PR with the link to the README back in place.

Issue #2132, opened 2023-12-28
NC upgrade erased entire data folder content


MarcS1975:
I have recently tried to upgrade from 27.0.1.2 to 28.0.1.1 and this action erased all content of my data folder (which was mapped as a docker volume). Luckily I had a backup but I am curious why the NC upgrade procedure did not retain my old data. 
You can see from my compose file below, that I have mapped an existing folder 'MEDIA' into the NC container. This MEDIA folder held 1TB of data has been filled prior to installing NC.

`- /srv/mypath/mydata/MEDIA:/var/www/html/media`

I add MEDIA as External Folder in Nextcloud with file type "local" and access to all data worked fine. 
When I decided to upgrade NC, I added **:latest** behind the NC container name and re-deployed the stack (in Portainer) with the option "pull containers". 

Then the NC log files started filing up with the following recurring errors:

```
28/12/2023
02:06:09
Initializing nextcloud 28.0.1.1 ...
28/12/2023
02:06:09
Upgrading nextcloud from 27.0.1.2 ...
28/12/2023
02:06:10
Nextcloud or one of the apps require upgrade - only a limited number of commands are available
28/12/2023
02:06:10
You may use your browser or the occ upgrade command to do the upgrade
28/12/2023
02:06:10
rsync: [generator] delete_file: rmdir(media) failed: Device or resource busy (16)
28/12/2023
02:06:32
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.2.7]
```

Any ideas why the upgrade resulted in all my data erased (but not the folder itself). Is NC not supposed to keep all persistent data untouched during an upgrade?

My docker compose file is as follows
```
version: '2'

volumes:
  nextcloud:
  db:

services:
  db:
    image: mariadb:10.6
    restart: always
    command: --transaction-isolation=READ-COMMITTED --log-bin=binlog --binlog-format=ROW
    volumes:
      - /srv/mypath/NC_db:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=xxx
      - MYSQL_PASSWORD=yyy
      - MYSQL_DATABASE=xxx
      - MYSQL_USER=yyy
      - REDIS_HOST=redis
      - REDIS_HOST_PASSWORD=xxx

  app:
    image: nextcloud**:latest**
    restart: always
    ports:
      - 8585:80

    links:
      - db
    volumes:
      - /srv/mypath/NC_app:/var/www/html
[...]

PR #2133, opened 2023-12-28
enh(README) Persistent volume upgrade.exclude clarification

joshtrichards:
- Make it a called out warning
- Clarify the language
- Clarity trade-offs
- Adds a couple headings in that section to allow for direct linking (as a FAQ/etc)

Fixes: #2132

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ If mounting additional volumes under `/var/www/html`, you should consider:
 - Mounting storage volumes to locations outside of `/var/www/html`.
 
 > [!WARNING]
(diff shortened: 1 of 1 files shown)


MarcS1975:
_I would make it even more explicit. People need to understand the severity of a potential mistake here._

**WARNING: DO NOT MOUNT ANY USER DATA TO /var/www/html. YOU WILL LOOSE ALL YOUR DATA WHEN UPGRADING THE NEXTCLOUD DOCKER IMAGE.**
Unless you are OK to loose all your mounted data during NC upgrades, we recommend mounting only to **/var/www/html/data** or **/var/www/html/config** and below since those folders will be excluded from deletion during upgrades. Further exclusions (in addition to /data/ and /config/) are defined in upgrade.exclude but require a custom image build.

PR #2138, opened 2024-01-07
Allow setting AWS S3 storage class

pathob:
The AWS S3 storage class can now be set using the environment variable `OBJECTSTORE_S3_STORAGE_CLASS`.

The order of the AWS S3 environment variables in the `README.md` has also been adjusted to be more "logical".


pathob:
Closes #1932 and makes #1933 obsolete as wished by the original author.

PR #2141, opened 2024-01-08
Improve handling of secret variables and files in autoconfig

pathob:
In the recent state, the autoconfig feature required either all database secrets to be passed as environment variables, or all to be passed as secret files, but didn't allow to use a mix of these variants, e.g. passing the database user and name as environment variables, and only the password as a file.

This is now fixed and allows working with database values the same way, as the Postgres and MySQL/MariaDB images do; with mixed variants also.

### DIFF
--- a/.config/autoconfig.php
+++ b/.config/autoconfig.php
@@ -6,34 +6,62 @@ if (getenv('SQLITE_DATABASE')) {
     $AUTOCONFIG['dbtype'] = 'sqlite';
     $AUTOCONFIG['dbname'] = getenv('SQLITE_DATABASE');
     $autoconfig_enabled = true;
-} elseif (getenv('MYSQL_DATABASE_FILE') && getenv('MYSQL_USER_FILE') && getenv('MYSQL_PASSWORD_FILE') && getenv('MYSQL_HOST')) {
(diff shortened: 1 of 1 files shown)


pathob:
@J0WI 👀

pathob:
Hello @J0WI @joshtrichards how can I help to move this forward?

pathob:
@J0WI I think the logic of my PR is a bit better, because it also checks the existence of the files. But tbh there seems something wrong with Nextcloud. In my deployment I manually needed to set the `NEXTCLOUD_DB_ENV_POSTGRES_HOST` variable, because it was missing, even though `POSTGRES_HOST` is correctly set. Unfortunately I couldn't find the location yet where these `NEXTCLOUD_DB_ENV` variables get "generated".

It would be nice to move the PR forward, please let me know how I can help...

pathob:
Unfortunately, the way this PR was (not) handled led me to the conclusion that this image is not suitable for production. Since the AIO image violates basic Docker principles, I will replace my Docker-based pilot setup with the service from NixOS.

joshtrichards:
> But tbh there seems something wrong with Nextcloud. In my deployment I manually needed to set the NEXTCLOUD_DB_ENV_POSTGRES_HOST variable, because it was missing, even though POSTGRES_HOST is correctly set. Unfortunately I couldn't find the location yet where these NEXTCLOUD_DB_ENV variables get "generated".

Where precisely are you seeing references to (or a dependency on) this variable (`NEXTCLOUD_DB_ENV_POSTGRES_HOST`)?

PR #2166, opened 2024-02-21
Fix #2165

### LINKED ISSUE #2165: mariaDB errors: “Aborted connection 5 to db: ‘unconnected’ user: ‘unauthenticated’ host: ‘172.21.0.3’ (This connection closed normally without authentication)”
Since the recent update or MariaDb (11.3.2) the connectivity broke down between nextcloud and mariadb.

This can be fixed by adding addition flag to the mariadb command `--character-set-server=utf8`

A solution would be to update the example docker compose file with these arguments
```
    image: mariadb:11.3.2
    command: --transaction-isolation=READ-COMMITTED --log-bin=binlog --binlog-format=ROW --character-set-server=utf8
```

_All credits for this solution go to [Gypertena's reply on the nextcloud discourse](https://help.nextcloud.com/t/nextcloud-broken-on-docker-zimaboard/182378/3?u=lhoupert)_ .

lhoupert:
To address https://github.com/nextcloud/docker/issues/2165

### DIFF
--- a/.examples/docker-compose/insecure/mariadb/apache/docker-compose.yml
+++ b/.examples/docker-compose/insecure/mariadb/apache/docker-compose.yml
@@ -2,8 +2,8 @@ version: '3'
 
 services:
   db:
-    image: mariadb:10.6
-    command: --transaction-isolation=READ-COMMITTED --log-bin=binlog --binlog-format=ROW
+    image: mariadb:11.3.2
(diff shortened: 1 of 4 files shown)


joshtrichards (commented):
No for me. This problem shouldn't (and doesn't AFAIK) occur if sticking with LTS MariaDB release versions that are actually supported officially by Nextcloud Server. :-)

https://mariadb.com/kb/en/mariadb-server-release-dates/
https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html#server

Issue #2194, opened 2024-04-04
WARN[0000] docker-compose.yml: `version` is obsolete


theTyster:
Docker has [deprecated][1] the use of `version` in the root level of the docker compose yaml beginning with Docker Compose 1.27+.

> # Version top-level element
> 
> The top-level version property is defined by the Compose Specification for backward compatibility. It is only informative.
> 
> Compose doesn't use version to select an exact schema to validate the Compose file, but prefers the most recent schema when it's implemented.
> 
> Compose validates whether it can fully parse the Compose file. If some fields are unknown, typically because the Compose file was written with fields defined by a newer version of the Specification, you'll receive a warning message. Compose offers options to ignore unknown fields (as defined by ["loose"](https://github.com/compose-spec/compose-spec/blob/master/01-status.md#requirements-and-optional-attributes) mode).

I'll make a PR for this.

## These lines can be removed:

https://github.com/nextcloud/docker/blob/f23d5d3ab4af4a65ffea30d05d1d1a7a37cd4bc1/.examples/docker-compose/with-nginx-proxy/mariadb/apache/docker-compose.yml#L1
https://github.com/nextcloud/docker/blob/f23d5d3ab4af4a65ffea30d05d1d1a7a37cd4bc1/.examples/docker-compose/with-nginx-proxy/mariadb/fpm/docker-compose.yml#L1
https://github.com/nextcloud/docker/blob/f23d5d3ab4af4a65ffea30d05d1d1a7a37cd4bc1/.examples/docker-compose/with-nginx-proxy/postgres/apache/docker-compose.yml#L1
https://github.com/nextcloud/docker/blob/f23d5d3ab4af4a65ffea30d05d1d1a7a37cd4bc1/.examples/docker-compose/with-nginx-proxy/postgres/fpm/docker-compose.yml#L1
https://github.com/nextcloud/docker/blob/f23d5d3ab4af4a65ffea30d05d1d1a7a37cd4bc1/.examples/docker-compose/insecure/mariadb/apache/docker-compose.yml#L1
https://github.com/nextcloud/docker/blob/f23d5d3ab4af4a65ffea30d05d1d1a7a37cd4bc1/.examples/docker-compose/insecure/mariadb/fpm/docker-compose.yml#L1
https://github.com/nextcloud/docker/blob/f23d5d3ab4af4a65ffea30d05d1d1a7a37cd4bc1/.examples/docker-compose/insecure/postgres/apache/docker-compose.yml#L1
[...]

PR #2195, opened 2024-04-04
Fix/remove versions from examples

theTyster:
https://github.com/nextcloud/docker/issues/2194

I think this PR is necessary because without this change users are shown a warning:`WARN[0000] docker-compose.yml: 'version' is obsolete`.

From my perspective, many self-hosters are not experts in Docker and often use it as a way to make setting up services easier. The provided example documents are valuable to these users for this reason.

Having warnings show in the terminal or console when running docker compose commands might affect how users think and feel about the software that they are running. From my perspective, these types of users seem to expect an out-of-the-box working set up with example files like these.

For that reason, instead of running in [loose mode][1], I recommend simply removing the version name from the example yamls.

[1]: https://github.com/compose-spec/compose-spec/blob/master/01-status.md#requirements-and-optional-attributes

### DIFF
--- a/.examples/docker-compose/insecure/mariadb/apache/docker-compose.yml
+++ b/.examples/docker-compose/insecure/mariadb/apache/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3'
-
 services:
   db:
     image: mariadb:10.6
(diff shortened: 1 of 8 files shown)


theTyster:
Mentioning #2201. Which provides a more robust solution than this simple fix. The real conversation is happening there.

PR #2201, opened 2024-04-18
Documentation and examples update for use with docker compose v2

### LINKED ISSUE #2200: Docs and examples update, again
Using this issue as placeholder. 
I am going to update the Readme / docs / examples for the docker compose syntax and file naming according to Docker and Docker Compose requirements. I see there's an open PR for the removal of the "version" parameter, but that's not enough. I had this idea when I updated the examples last time, but didn't had the time for it, now I do. I also plan on updating the example MariaDB from 10.6 to 10.8.2 based on few years production experience, and add redis to the default examples because just like last time - I see a lot of people ignoring the warning that this repository is for expert use, clone the examples section as-is and then try to figure out why it is not working properly. 
 
CC @J0WI , CC  #2196 #2194

tzerber:
Howdy,

I noticed #2194 and the PR but there are more problems with the examples than just the `version` parameter. Docker changed the file name requirements too, changed the command and some of the syntax, jwilder's proxy is now part of nginx-proxy and more. This PR addresses those changes, and few others as well.

What was changed: 
1) Docs:  ( README.MD ) have commands, file names and examples updated everywhere. Minor edits to match the examples section and links.

2) Files: All `docker-compose.yml` files are renamed to `compose.yaml` files as per docker requirements.

3) Nginx configuration: I've been following the nginx mailing list since the problems we had with `mjs` file extension last year, and they wont fix it despite browsers expecting it as `text/javascript`. I have modified the examples to include mime-types from Nginx to the web container, and removed the "workaround" from the `nginx.conf` file to keep it as consistent as possible with the official Nextcloud docs. Effectively there is no difference from end user point of view. 

4) MariaDB: - Over the years we have several instances running using MariaDB and experience shows that 10.8.2 works better than 10.6 with recent versions of Nextcloud. I fully understand that it's not LTS release but we had several random problems with 10.6 that caused larger downtimes and then we switched to 10.8.2 in production, no issues whatsoever since then. It is also intentional that I didn't go newer than 10.8.2 because I cannot "promote" something to people that I hadn't tested properly myself. Considering majority of users copy-paste the examples and expect them to be working without issues - I believe this is better. 

5) Redis : I've added it on all the places it was missing. It should be there, all it does is speeding up nextcloud in general. 

6) Logging: Added a limit of 50 megabytes of log files on all containers. Recently I noticed that some instances had their log files grown to some absurd sizes (uptime for more than year without updates/downtime) and I've added that to the examples as well, for the same reason i changed the mariadb version. On few places(proxy containers, to be precise) I have intentionally added the logging driver to be `syslog`, because it works better with fail2ban on the host machine, it can't do any harm and it can hint to semi-sleeping sysadmins that the machine is under bots attack trying to log-in.

closes #2200

### DIFF
--- a/.examples/docker-compose/insecure/mariadb/apache/compose.yaml
+++ b/.examples/docker-compose/insecure/mariadb/apache/compose.yaml
@@ -1,10 +1,12 @@
-version: '3'
-
 services:
   db:
-    image: mariadb:10.6
-    command: --transaction-isolation=READ-COMMITTED --log-bin=binlog --binlog-format=ROW
+    image: mariadb:10.8.2
+    command: --transaction-isolation=READ-COMMITTED --binlog-format=ROW --innodb-read-only-compressed=OFF
     restart: always
+    logging:
+      driver: "json-file"
+      options:
+        max-size: "50m"
     volumes:
       - db:/var/lib/mysql:Z
     environment:
@@ -21,6 +23,10 @@ services:
   app:
     image: nextcloud:apache
     restart: always
+    logging:
+      driver: "json-file"
+      options:
+        max-size: "50m"
     ports:
       - 127.0.0.1:8080:80
     volumes:
@@ -37,6 +43,10 @@ services:
   cron:
     image: nextcloud:apache
     restart: always
+    logging:
+      driver: "json-file"
+      options:
+        max-size: "50m"
     volumes:
       - nextcloud:/var/www/html:z
     entrypoint: /cron.sh
--- a/.examples/docker-compose/insecure/mariadb/fpm/compose.yaml
+++ b/.examples/docker-compose/insecure/mariadb/fpm/compose.yaml
@@ -1,10 +1,12 @@
-version: '3'
-
 services:
   db:
-    image: mariadb:10.6
-    command: --transaction-isolation=READ-COMMITTED --log-bin=binlog --binlog-format=ROW
+    image: mariadb:10.8.2
+    command: --transaction-isolation=READ-COMMITTED --binlog-format=ROW --innodb-read-only-compressed=OFF
     restart: always
+    logging:
+      driver: "json-file"
+      options:
+        max-size: "50m"
     volumes:
       - db:/var/lib/mysql:Z
     environment:
@@ -21,6 +23,10 @@ services:
   app:
     image: nextcloud:fpm-alpine
     restart: always
+    logging:
+      driver: "json-file"
+      options:
+        max-size: "50m"
     volumes:
       - nextcloud:/var/www/html:z
     environment:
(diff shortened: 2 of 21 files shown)


tzerber:
> I still have the mjs problem and spent 4 hours yesterday (just to say I'm not just lazily expecting help) to see how modifying nginx.conf might help with the well known error message about the mjs mime type. No success. I only mount the nginx.conf specifically in the container from outside, not the mime.types.
> 
> I'm working off of this one with the fpm alpine docker compose:
> 
> [.examples/docker-compose/with-nginx-proxy/mariadb/fpm/web/nginx.conf](https://github.com/nextcloud/docker/blob/master/.examples/docker-compose/with-nginx-proxy/mariadb/fpm/web/nginx.conf)
> 
> Some of the posts I read through suggest it has something to do with DNS / domain entries, but a curl to an mjs file from inside the container yields results, so it doesn't seem it's a URL issue.
Hi,

Use the `nginx.conf` from this commit, it works fine without the need of mounting the mime-types file. ->> 05353cf4e5b743028a6d24e7496033c2b6811ea3

tzerber:
> > I thought so, reverted back to 10.6 as per docs.
> 
> > Removed the logging sections too.
> 
> Needs to be reverted in the README too

Reverted the README me, my bad. Also removed the GH-specific markdown. Fixed the little ":" too.

tzerber:
Bump, i guess ?

tzerber:
I added them, but I've also noticed the PR for MariaDB 10.11. Should i just change to 10.11 here as well ? @J0WI

tzerber:
> The command should match what's documented in https://docs.nextcloud.com/server/latest/admin_manual/configuration_database/linux_database_configuration.html#configuring-a-mysql-or-mariadb-database. `--innodb-read-only-compressed=OFF` is not set there.


Updated it to match to the docs link you provided. In this repo version (front page) [there is this section ](https://github.com/nextcloud/docker?tab=readme-ov-file#base-version---apache)  where it says `--transaction-isolation=READ-COMMITTED --log-bin=binlog --binlog-format=ROW`. Updated it to become `--transaction-isolation=READ-COMMITTED --binlog-format=ROW` as well.  Also found a .md file not updated and fixed it's `docker compose` command.

Also sorry, didn't noticed i pushed the wrong change last time, totally my fault. Will be a bit more careful from now on.

tzerber:
> > > > I thought so, reverted back to 10.6 as per docs.
> > > 
> > > 
> > > > Removed the logging sections too.
> > > 
> > > 
> > > Needs to be reverted in the README too
> > 
> > 
> > Reverted the README me, my bad. Also removed the GH-specific markdown. Fixed the little ":" too.
> 
> They are re-added in the README

Found it https://github.com/nextcloud/docker/pull/1881

tzerber:
I think this is ready now :)

PR #2202, opened 2024-04-24
PHP 8.3 for Nextcloud 29

J0WI:
blocked by https://github.com/Imagick/imagick/issues/640


joshtrichards:
Looks like AIO did a temporary patch:

nextcloud/all-in-one#5575
nextcloud/all-in-one#5582

PR #2206, opened 2024-04-26
Bump to 28.0.5 and 27.1.9

mejo-:


mejo-:
Sorry, just saw that you already pushed a change in ef382014777bc7d747792ab8d70b8ca162b25619 @J0WI. I wonder though why the `update.sh` workflow didn't get triggered afterwards :thinking:

mejo-:
Well sorry for my ignorance, I'm just finding my way around in this repo. `update.sh` gets run periodically, but even though it ran and bumped Nextcloud 28 to `28.0.5`, there's no `28.0.5` images available at https://hub.docker.com/_/nextcloud/tags yet.

PR #2225, opened 2024-05-24
Update Dockerfile-debian.template

mickaelmonsieur:
Fix Dockerfile template to properly evaluate environment variables in nextcloud.ini

The previous version of the Dockerfile used single quotes around the echo commands, preventing the proper evaluation of environment variables. This commit fixes the issue by replacing the single quotes with double quotes, allowing for the interpolation of environment variables. Now, the nextcloud.ini file correctly reflects the values of the PHP_MEMORY_LIMIT and PHP_UPLOAD_LIMIT environment variables.

### DIFF
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -107,9 +107,9 @@ RUN { \
     echo 'apc.enable_cli=1' >> "${PHP_INI_DIR}/conf.d/docker-php-ext-apcu.ini"; \
     \
     { \
-        echo 'memory_limit=${PHP_MEMORY_LIMIT}'; \
-        echo 'upload_max_filesize=${PHP_UPLOAD_LIMIT}'; \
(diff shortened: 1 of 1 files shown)


mickaelmonsieur:
the code is not evaluated, here is the proof:

```
root@3fbd95508472:/usr/local/etc/php/conf.d# cat nextcloud.ini
memory_limit=${PHP_MEMORY_LIMIT}
upload_max_filesize=${PHP_UPLOAD_LIMIT}
post_max_size=${PHP_UPLOAD_LIMIT}
```

joshtrichards:
Thanks for the contribution, however, you're mistaken. That ini file is telling PHP to pull in from those environment variables at runtime. e.g. see [here](https://www.php.net/manual/en/configuration.file.php). We do not want interpolation to occur there when generating the config file in the `Dockerfile` itself. So it's correct as-is.

joshtrichards (commented):
See prior comment. Closing to avoid confusion.

Issue #1402, opened 2021-02-10
BUG: files:scan user -> Exception during scan. Fresh install, and userns-remap disabled


schklom:
Hi,

I recently reinstalled `nextcloud` from scratch (all volumes deleted) after deciding to use userns-remap on Docker.
I deactivate it for `nextcloud` and the whole thing works as well as before, with one exception.

BUG:
```
$ docker exec -it nextcloud bash
I have no name!@3e615422f49e:/var/www/html$ php occ files:scan schklom
Starting scan for user 1 out of 1 (schklom)
Exception during scan: NOAUTH Authentication required.
#0 /var/www/html/lib/private/Memcache/Redis.php(110): Redis->set('b7277d39df73116...', 0, Array)
#1 /var/www/html/lib/private/Lock/MemcacheLockingProvider.php(84): OC\Memcache\Redis->add('files/0182828f2...', 0)
#2 /var/www/html/lib/private/Files/Storage/Common.php(751): OC\Lock\MemcacheLockingProvider->acquireLock('files/0182828f2...', 2, 'home::schklom::sc...')
#3 /var/www/html/lib/private/Files/Storage/Wrapper/Wrapper.php(594): OC\Files\Storage\Common->acquireLock('scanner::', 2, Object(OC\Lock\MemcacheLockingProvider))
#4 /var/www/html/lib/private/Files/Cache/Scanner.php(333): OC\Files\Storage\Wrapper\Wrapper->acquireLock('scanner::', 2, Object(OC\Lock\MemcacheLockingProvider))
#5 /var/www/html/lib/private/Files/Utils/Scanner.php(260): OC\Files\Cache\Scanner->scan('', true, 3)
#6 /var/www/html/apps/files/lib/Command/Scan.php(152): OC\Files\Utils\Scanner->scan('/schklom', true, NULL)
#7 /var/www/html/apps/files/lib/Command/Scan.php(208): OCA\Files\Command\Scan->scanFiles('schklom', '/schklom', Object(Symfony\Component\Console\Output\ConsoleOutput), false, true, false)
#8 /var/www/html/3rdparty/symfony/console/Command/Command.php(255): OCA\Files\Command\Scan->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
[...]

PR #2232, opened 2024-06-02
Update redis.config.php

VesperiaArt:
Correctly set the redis password config from the REDIS_HOST_PASSWORD_FILE environment variable. Fix an issue similar to #1402 when using the REDIS_HOST_PASSWORD_FILE environment variable to provide the redis host secret.

### DIFF
--- a/.config/redis.config.php
+++ b/.config/redis.config.php
@@ -5,7 +5,7 @@ if (getenv('REDIS_HOST')) {
     'memcache.locking' => '\OC\Memcache\Redis',
     'redis' => array(
       'host' => getenv('REDIS_HOST'),
-      'password' => (string) getenv('REDIS_HOST_PASSWORD'),
+      'password' => getenv('REDIS_HOST_PASSWORD_FILE') && file_exists(getenv('REDIS_HOST_PASSWORD_FILE')) ? trim(file_get_contents(getenv('REDIS_HOST_PASSWORD_FILE'))) : (string) getenv('REDIS_HOST_PASSWORD'),
     ),
   );
 


joshtrichards (approved):
Thanks!

VesperiaArt:
> Please update the Readme as well.

Not sure what to add, since the README already mention that REDIS_HOST_PASSWORD_FILE is supported for docker secrets. This PR fixes the nextcloud's redis config reading the password from this environment variable.

Do you mean adding instruction for the user to apply the fix themselves ?

.config/redis.config.php
```diff
@@ -5,7 +5,7 @@
     'memcache.locking' => '\OC\Memcache\Redis',
     'redis' => array(
       'host' => getenv('REDIS_HOST'),
-      'password' => (string) getenv('REDIS_HOST_PASSWORD'),
+      'password' => getenv('REDIS_HOST_PASSWORD_FILE') && file_exists(getenv('REDIS_HOST_PASSWORD_FILE')) ? trim(file_get_contents(getenv('REDIS_HOST_PASSWORD_FILE'))) : (string) getenv('REDIS_HOST_PASSWORD'),
```
tianon, line 8: If the variable is specified, but the file doesn't exist, this will fall back to a likely non-existent `REDIS_HOST_PASSWORD` -- shouldn't it error out or something instead?

tianon: Sorry, what I'm suggesting is that perhaps `file_exists(getenv('REDIS_HOST_PASSWORD_FILE'))` should be dropped from the expression so that `file_get_contents` will fail and give the operator at least a warning somewhere so they have some clue why they specified the environment variable for pulling from a file but it isn't working.

tianon: > An [E_WARNING](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-warning) level error is generated if filename cannot be found, length is less than zero, or if seeking to the specified offset in the stream fails.
>
> When file_get_contents() is called on a directory, an [E_WARNING](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-warning) level error is generated on Windows, and as of PHP 7.4 on other operating systems as well.

(from https://www.php.net/file_get_contents)

PR #2236, opened 2024-06-14
Update redis.config.php

Hujgo:
add dbindex for redis as comment. referenced from a6f2a3c commit.

### DIFF
--- a/.config/redis.config.php
+++ b/.config/redis.config.php
@@ -6,6 +6,7 @@ if (getenv('REDIS_HOST')) {
     'redis' => array(
       'host' => getenv('REDIS_HOST'),
       'password' => getenv('REDIS_HOST_PASSWORD_FILE') && file_exists(getenv('REDIS_HOST_PASSWORD_FILE')) ? trim(file_get_contents(getenv('REDIS_HOST_PASSWORD_FILE'))) : (string) getenv('REDIS_HOST_PASSWORD'),
+      'dbindex' => (int) getenv('REDIS_DB_INDEX'),
     ),
   );
 


joshtrichards:
That's not a comment. :-)

Also, this would have to be handled in the PHP session handler too otherwise the result might not be what people expect...

https://github.com/nextcloud/docker/blob/064069b3060092b64f6fc1e418932fa215d57364/docker-entrypoint.sh#L108-L133

.config/redis.config.php
```diff
@@ -6,6 +6,7 @@
     'redis' => array(
       'host' => getenv('REDIS_HOST'),
       'password' => getenv('REDIS_HOST_PASSWORD_FILE') && file_exists(getenv('REDIS_HOST_PASSWORD_FILE')) ? trim(file_get_contents(getenv('REDIS_HOST_PASSWORD_FILE'))) : (string) getenv('REDIS_HOST_PASSWORD'),
+      'dbindex' => (int) getenv('REDIS_DB_INDEX'),
```
joshtrichards, line 9:
```suggestion
      // 'dbindex' => (int) getenv('REDIS_DB_INDEX'),
```

GroondCharge:
> Can you elaborate the use case?

Having an external redis master-master cluster and deploying nextcloud only on designated node. I am aware that prefix hash is calculated by concatenating instanceid, path and version, so in theory there should be no conflicting prefixes, however being able to separate it in separate dbindices in case of needing to cleanup after update etc. is a bonus, especially if you're managing mutliple instances connecting to single cluster. One of the solutions would be having option to use consistent prefix for redis entries.

GroondCharge:
If this will get pulled into main, has the session lock configuration been adjusted aswell? I've tried an example here https://github.com/nextcloud/docker/pull/2288/files

joshtrichards:
> Having an external redis master-master cluster and deploying nextcloud only on designated node

But Redis Cluster doesn't support dbindex / multiple databases.

https://redis.io/docs/latest/commands/select/

PR #2090, opened 2023-10-26
use var PHP_OPCACHE_MEMORY_CONSUMTION for configuration

tclavier:

PR #2237, opened 2024-06-15
Increase OPcache memory consumption

Strit:
Recently I have been seeing the OPcache warning more and more while only having a small amount of Apps installed.

128 MB might just be too little, so double it to 256 MB, which should still be reasonable for most installations.

Signed-off-by: Dan Johansen <strit@strits.dk>


markuman:
https://github.com/nextcloud/docker/pull/2090/

PR #2220, opened 2024-05-18
add imagemagick-svg to fpm-alpine Dockerfile

felix-egli:
this adds support for SVG images.

PR #2246, opened 2024-06-25
add imagemagick-svg *back* to all the alpine images

jessebot:
*(Context: Alpine changed their imagemagick packaging)*

This is an alternative to #2220 to move it forward. Followed the instructions per this comment: https://github.com/nextcloud/docker/pull/2220#pullrequestreview-2089466835

I can do another PR for imagemagick-pdf as well if you want, but to keep this short and simple, I've only done svg here, as that's the only one that throws a warning.


joshtrichards:
> I can do another PR for imagemagick-pdf as well if you want, but to keep this short and simple, I've only done svg here, as that's the only one that throws a warning.

👍 Thank! Let's get this in since it's the most pressing with 29.0.3 right around the immediate corner.

But let's also create an issue now (or, better yet, WIP PR) to propose & firm up what others should also be added back in, in a follow-up[1]. 

[1] *We probably want to keep parity. Is there anything that doesn't belong or should we just load all of them back in to mirror the old packaging?* 

Cc: @felix-egli

joshtrichards (approved):

PR #2248, opened 2024-06-27
Add pdf+other image preview support to alpine images after Alpine's packaging change

jessebot:
I'm also creating this PR to add PDF preview support, which doesn't work without the `imagemagick-pdf` package. This is in reference to https://github.com/nextcloud/docker/pull/2246#issuecomment-2190435512:

> But let's also create an issue now (or, better yet, WIP PR) to propose & firm up what others should also be added back in, in a follow-up[1].

> [1] We probably want to keep parity. Is there anything that doesn't belong or should we just load all of them back in to mirror the old packaging?

> Cc: @felix-egli

We could also combine this PR with https://github.com/nextcloud/docker/pull/2246 since we missed the release for `29.0.3` this time anyway. Open to either way, or if I misunderstood, please feel free to give me gentle feedback and I'll remedy what's needed! 🙏

Also open to adding additional packages. `svg` and `pdf` are just the file formats that stuck out to me from the beginning.

### DIFF
--- a/27/fpm-alpine/Dockerfile
+++ b/27/fpm-alpine/Dockerfile
@@ -6,6 +6,12 @@ RUN set -ex; \
     \
     apk add --no-cache \
         imagemagick \
+        imagemagick-pdf \
+        imagemagick-jpeg \
+        imagemagick-raw \
+        imagemagick-tiff \
+        imagemagick-heic \
+        imagemagick-webp \
         rsync \
     ; \
     \
--- a/28/fpm-alpine/Dockerfile
+++ b/28/fpm-alpine/Dockerfile
@@ -6,6 +6,12 @@ RUN set -ex; \
     \
     apk add --no-cache \
         imagemagick \
+        imagemagick-pdf \
+        imagemagick-jpeg \
+        imagemagick-raw \
+        imagemagick-tiff \
+        imagemagick-heic \
+        imagemagick-webp \
         rsync \
     ; \
     \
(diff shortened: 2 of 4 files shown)


joshtrichards:
Per https://git.alpinelinux.org/aports/commit/community/imagemagick/imagemagick.post-upgrade?h=3.19-stable&id=61dce5f6a84d9f96ab7aa6437f12d2845b572ed8 these are our options:

```
+	* imagemagick support for various modules was split into subpackages.
+	* they autoinstall with the requisite library already installed.
+	* if not already present, install the module you want to use manually:
+	* (prefixed with imagemagick- )
+	* -heic -jpeg -pdf -raw
+	* -svg -tiff -webp -jxl
+	* if you want to exclude the support regardless, use e.g. 'apk add !imagemagick-pdf'
```

Looking at the formats supported in our Debian images, looks like we do all of these except for `jxl`. I guess we should include them all save for that one? :thinking: 

> since we missed the release for 29.0.3 this time anyway

Yeah. :-/ Up to you whether to combine them or not. Doesn't matter either way on this side now.

joshtrichards:
Ah, crap. It thinks there's a conflict now due to the one line change from the svg PR since that just got merged. :stuck_out_tongue_winking_eye: Can you resolve however you prefer and then let's get this one in too.

joshtrichards (approved):

joshtrichards:
Wow, all tests passed without having to kick them a few times to re-run. That's rare, but I'll take it!

PR #2275, opened 2024-08-13
Added custom opcache config override

bronco0:
Added the ability to customize PHP OPcache configuration.
Default values ​​have been set to environment variables that can be overridden.


bronco0:
I think it's a good idea to be able to customize any OPcache setting.

pankrazv:
@J0WI As I'm interested in this as well and @bronco0 isn't responding I might take over this PR and fix the points you mention in your review.

bronco0:
Sorry, I haven't had much time lately.
I'll see about quickly modifying my PR.

@J0WI do we keep all these parameters or only opcache.save_comments and opcache.jit ?

PR #2278, opened 2024-08-18
replace build image with bind mount

SebastianRzk:
A separate image is currently being built for the nginx proxy (service name: web). According to my information, this is not necessary, but even counterproductive

Why?

* The image does not need to be built (why should it?) 
* Fewer files and less code = easier maintenance and fewer potential bugs
* the container can be easily updated (for example via a program like watchtower)


tzerber:
I agree here. I would also move the configuration file outside of the "web" directory or just rename the web directory to "conf" one. This has the good side effect that will allow using that same "conf" directory for other stuff, like mounting mime-types; Nextcloud's own configuration there in the case of bind mounts; proxy related configurations; even in my use case - Nexcloud's container configs like `opcache-recommed.ini` and similar.

Juoelenis:
Agreed 🙂

Issue #2226, opened 2024-05-25
Document how to use `NC_*` environment variables with Nextcloud Server via the image


joshtrichards:
* via `docker run` / command line
* via `docker compose` / Compose File

Should significantly cut down on many requests to add additional configuration variables to the image.

joshtrichards:
Though I'm thinking maybe we should nudge people towards simply using the *Auto configuration via* hooks support since all `config.php` values can be set via `occ config:system:set` whereas `NC_*` doesn't support any parameters with `.` in them nor any array values (which are both big limitations). It's also not documented and thus not well supported upstream.

joshtrichards:
Upstream handling of `NC_*` variables should have better visibility now that nextcloud/server#46140 is merged (it'll be in next maintenance release for each supported major).

PR #2287, opened 2024-08-30
Added docker env debug mode

archmagece:
There doesn't seem to be an easy way to handle debug mode when running with docker, so it would be nice to add it.

I just added log.config.php to all config/ directory.

```php
<?php
if (getenv('NEXTCLOUD_DEBUG_MODE') !== false) {
    $CONFIG = array (
        'debug' => true,
      );
}
```


joshtrichards:
Hi @archmagece! And thanks for contributing!

Specifying `NC_debug=true` as a Docker environment variable should work, though documentation for this is lacking upstream and here. See #2226.

archmagece:
I got it. Thank you for your resposne.

PR #2288, opened 2024-09-02
Adding support for DBINDEX redis config value in docker compose

GroondCharge:
Currently, there is no support for adding dbindex config value ($CONFIG['redis']['dbindex']) within compose file. Even in manually updating or setting the value for a container, the changes aren't applied. This fixes the missing option.

### DIFF
--- a/29/apache/Dockerfile
+++ b/29/apache/Dockerfile
@@ -21,6 +21,7 @@ RUN set -ex; \
 # see https://docs.nextcloud.com/server/stable/admin_manual/installation/source_installation.html
 ENV PHP_MEMORY_LIMIT 512M
 ENV PHP_UPLOAD_LIMIT 512M
+ENV REDIS_DB_INDEX=0
 RUN set -ex; \
     \
     savedAptMark="$(apt-mark showmanual)"; \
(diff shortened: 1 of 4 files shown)


GroondCharge:
> Please see:
> 
> #2236 #1286
> 
> I'm not 100% against this, but the session handler needs to be considered too. It's not clear what the "right thing to do here is".
> 
> Personally I'd just avoid dbindex and use dedicated Redis. This is a container environment after all. :-)

https://github.com/nextcloud/docker/pull/1286#issuecomment-1797020464 this becomes a hassle especially, when one'd try to automate deployment. It feels like this would be a nice feature to have available if needed irregardless if it's not optimal for 90% of the cases. However! There is a middle ground imo - prefixes. This would work if you had a redis instance that's running for multiple services or even, if there are multiple NCs connecting to a redis cluster

GroondCharge:
According to this: https://github.com/nextcloud/server/issues/32203 the prefix for nextcloud is calculated, but then changes when either path, version and/or instanceid would change. If there was a way to make prefix optionally persistent, then these issues would be solved.

GroondCharge:
> Hi, maybe this helps as reference? [nextcloud/all-in-one#5126](https://github.com/nextcloud/all-in-one/pull/5126)

Thank you for this, i changed it to configure session locking appropriately, aswell as name for env variable, to match the ones you used

Hujgo (approved):

nioc:
Hello, any news from this PR?

Tireur2cables:
Hi, having the same need to set dbindex, it would be great if this could be merged.  
@GroondCharge do you have time to rebase and resolve conflicts ? If you want I can also start a new PR with these changes.

PR #2289, opened 2024-09-02
Added information about permissions errors (migrating to alpine)

alexandra-branco:


alexandra-branco:
@joshtrichards @J0WI  Thanks for the suggestions!  
I have already made the suggested changes.

joshtrichards:
Thanks for following up on this and wrapping it up, @alexandra-branco.

PR #2290, opened 2024-09-03
Exclude autoconfig.php from user config check

flortsch:
See: https://github.com/nextcloud/docker/pull/2120#issuecomment-2313378862


joshtrichards:
Thanks, @flortsch!

PR #2298, opened 2024-09-18
MariaDB 10.11

J0WI:


joshtrichards (approved):

tzerber:
Hi, i've been working on the examples and I saw this. While looking at "should or shouldn't use `--log-bin=binlog`", i stumbled across a lot of mysql compose files on legitimate repos like Wise's and they all point to [this issue](https://github.com/MariaDB/mariadb-docker/issues/434#issuecomment-1136151239) when using 10.11 MariaDB. Should we add this
```yaml 
    security_opt: # see https://github.com/MariaDB/mariadb-docker/issues/434#issuecomment-1136151239
      - seccomp:unconfined
      - apparmor:unconfined
```
to the compose files in the examples + the readme once 10.11 is in use ? I will give it a test tomorrow to see if it actually works, because that issue appears closed.

joshtrichards:
Interesting - and thanks for watching out for stuff l like this.

I did a quick look and I think it's limited to old Docker Engine / runtimes w/ outdated seccomp policies. I didn't do too deep of a dive though admittedly.

I'd rather not include a workaround for something that is probably not common / should be fixed in other ways

If it was (or still is) a big enough deal, MariaDB presumably has a doc link somewhere we can just link to and say "do whatever they say" if impacted by this.

joshtrichards:
I say let's merge this. I'll do a similar one for the postgresql examples.

PR #2307, opened 2024-10-08
Update examples nginx configuration

tzerber:
I noticed there's some changes in the official docs about the nginx configuration, and updated the examples files here too. 
Tested the configuration, it works fine.


joshtrichards:
Thanks for keeping up on these, @tzerber.

Issue #2308, opened 2024-10-08
`OBJECTSTORE_S3_SSL` does not work as documented


vbrandl:
The documentation for `OBJECTSTORE_S3_SSL` states that it defaults to `true`:

> (default: true): Whether or not SSL/TLS should be used to communicate with object storage server

I would assume, not setting `OBJECTSTORE_S3_SSL` at all would cause Nextcloud to connect to my S3 bucket via HTTPS.

I use the following `compose.yml` for testing.

```yaml
services:
  nextcloud:
    image: nextcloud
    volumes:
      - ./data/nextcloud:/var/www/html
    environment:
      - SQLITE_DATABASE=/tmp/nc.db
      - NEXTCLOUD_ADMIN_USER=admin
      - NEXTCLOUD_ADMIN_PASSWORD=this-is-not-a-password
      - NEXTCLOUD_TRUSTED_DOMAINS=example.com

      - OBJECTSTORE_S3_BUCKET=${OBJECTSTORE_S3_BUCKET}
      - OBJECTSTORE_S3_HOST=${OBJECTSTORE_S3_HOST}
      - OBJECTSTORE_S3_PORT=${OBJECTSTORE_S3_PORT}
      - OBJECTSTORE_S3_KEY=${OBJECTSTORE_S3_KEY}
      - OBJECTSTORE_S3_SECRET=${OBJECTSTORE_S3_SECRET}
        #- OBJECTSTORE_S3_SSL=true
```

If i start a new instance like this, I get the following `config.php`:

```php
objectstore' =>
  array (
    'class' => '\\OC\\Files\\ObjectStore\\S3',
    'arguments' =>
    array (
      'bucket' => '<snip>',
      'region' => '',
      'hostname' => '<snip>',
      'port' => '443',
      'storageClass' => '',
      'objectPrefix' => 'urn:oid:',
      'autocreate' => false,
      'use_ssl' => false,
      'use_path_style' => false,
      'legacy_auth' => false,
      'key' => '<snip>',
      'secret' => '<snip>',
    ),
  ),
```

`use_ssl` is set to `false` but I would expect it to be `true`.

Now deleting `./data/nextcloud`, uncommenting `- OBJECTSTORE_S3_SSL=true` in `compose.yml`, `use_ssl` in `config.php` will be set to `true`.

So the documentation does not match how the image actually behaves.

joshtrichards:
[...]

Issue #1948, opened 2023-03-21
Fix S3 `use_ssl` and `autocreate` default values


aentwist:
## Description

From the [PHP Manual for `getenv`](https://www.php.net/manual/en/function.getenv.php), we have

>  `getenv(string $varname, bool $local_only = false): string|false`

> Returns the value of the environment variable varname, or false if the environment variable varname does not exist.

Now check this out

https://github.com/nextcloud/docker/blob/a15c755399e299914acebe0a5005dd5ddff0db56/.config/s3.config.php#L3

https://github.com/nextcloud/docker/blob/a15c755399e299914acebe0a5005dd5ddff0db56/.config/s3.config.php#L19

Value not set -> `getenv` returns `false` -> `use_ssl` gets `false` -> ternary goes to `false`

This is a great example of why to not allow poorly written code into a codebase. It obfuscates things.

Should probably be

```php
// if defined and value is "false" then set to false
!(getenv('OBJECTSTORE_S3_SSL') && strtolower(getenv('OBJECTSTORE_S3_SSL')) === 'false')
// demorgans law
// if not defined or doesn't have value "false" then set to true
!getenv('OBJECTSTORE_S3_SSL') || strtolower(getenv('OBJECTSTORE_S3_SSL')) !== 'false'
```

---

There are several other suspicious variables. Using a ternary operator for a boolean expression is redundant. Not relying on truthy falsy values in favor of explicit checks is also somewhat suspicious here.

https://github.com/nextcloud/docker/blob/a15c755399e299914acebe0a5005dd5ddff0db56/.config/s3.config.php#L18-L23

https://github.com/nextcloud/docker/blob/a15c755399e299914acebe0a5005dd5ddff0db56/.config/swift.config.php#L8

---

## Requirements

- build docker image
- [ ] set *only* `OBJECTSTORE_S3_BUCKET` (use defaults)
- exec in
- examine config/config.php and check default values against README
  - [ ] `autocreate` is `true`
  - [ ] `use_ssl` is `true`
  - [x] `use_path_style` is `false`
  - [x] `legacy_auth` is `false`
- [x] set only `OBJECTSTORE_SWIFT_URL`
  - [x] `autocreate` is `false`

aentwist:
will pick this up if my other PRs are resolved

PR #2309, opened 2024-10-08
Fix initialization of `autocreate` and `use_ssl`

vbrandl:
According to the documentation, both `OBJECTSTORE_S3_SSL` and `OBJECTSTORE_S3_AUTOCREATE` should default to `true`. Currently, when these environment variables are not set, they default to `false`. (See https://github.com/nextcloud/docker/issues/2308).

With this fix, both values will be `true` if they are

* not set
* the empty string
* any string that is not equal to `false` when converted to lowercase

This should now match the documented behavior.


Juoelenis (commented):
I mightnotunderstand this gibberish, but i know its no good.

vbrandl:
What exactly do you mean? What "gibberish" don't you understand and which part of my proposed change is "no good"?

Juoelenis:
Ah yes sorry for da thang, it seems it is a legit bug fix, but submit it to mail, in PR they dont merge, i dunno why but you have to send the fix via mail

vbrandl:
How do I submit my PR via mail? Or do you mean that my commit message has to contain `Signed-off-by: Author Name <authoremail@example.com>` as stated in the failed CI action?

I will update the commit message and push again soon

Juoelenis:
They dont accept pr's thru the PR, you must send them the changes via mail, understand?

joshtrichards:
Related: #1948

vbrandl:
I just pushed again with proper `Signed-off-by` in my commit message.

Also sorry for falling for the troll...

PR #2311, opened 2024-10-10
feat(entrypoint): add debug mode for entrypoint

joshtrichards:


joshtrichards:
Might also think about adding `-v` to `occ maintennace:install`

joshtrichards:
Cc: @tzerber You might find this of interest too

tzerber:
> Cc: @tzerber You might find this of interest too

I do like it, neat. Would also help debugging fresh installs like the one we had few weeks(months? can't remember) ago where upon fresh install on slower hardware it was reporting weird errors and the only solution was to down/up the container.  Good job!

joshtrichards:
> It has almost the same effect as 

@J0WI Well, it also turns on verbose mode for rsync and the installer (added in v28: see nextcloud/server#41214).

Perhaps verbose mode for the installer should on by default in the entrypoint (with or without this PR). After all, updating is already fairly verbose. The installer just isn't by default. It's a one-off event so maybe that's the approach there.

Verbose mode for rsync could be helpful in troubleshooting situations that arise with folks with whacky NFS storage/etc. (seems to arise with k8s/helm users most often).

PR #2314, opened 2024-10-13
Disable absolute redirect

TommyTran732:
Absolute redirect should be disabled. Currently, it only works by chance because the `web` container is listening on port 80.

If you try to change the port to something like 8080, the 301 redirect for things like caldav and carddav will be broken.


TommyTran732:
It doesn't matter though, when you are using an app behind a reverse proxy (as is the case of nginx in this stack) it is expected that relative redirect is used instead of absolute redirect. Absolute redirect should be disabled for correctness.

I wasted quite a lot of time debugging weird error because of this and I see no reason to not do it to save other people the trouble.

PR #2334, opened 2024-11-16
Enable TLS connection for Redis

Ziris85:
Hello!

This is a hopefully simple PR to allow users to configure Redis to connect to a TLS-enabled Redis endpoint. It proposes a new environment variable `REDIS_PROTOCOL` that can be passed to the container, and replaces all areas (that I found) where `tcp://` was either hard-coded or the assumed (omitted) default. All adjustments will continue to use `tcp` as the default if the env var isn't passed, so this shouldn't impact existing installations.

I don't do PR's very frequently so apologies if I missed something here, but I'm interested in thoughts/feedback!

Thank you!

### DIFF
--- a/28/apache/config/redis.config.php
+++ b/28/apache/config/redis.config.php
@@ -4,7 +4,7 @@ if (getenv('REDIS_HOST')) {
     'memcache.distributed' => '\OC\Memcache\Redis',
     'memcache.locking' => '\OC\Memcache\Redis',
     'redis' => array(
-      'host' => getenv('REDIS_HOST'),
+      'host' => ($_ENV['REDIS_PROTOCOL'] ?? 'tcp') . "://" . getenv('REDIS_HOST'),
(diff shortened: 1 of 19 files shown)


LordRobinCbz:
Hello,

I was here to add this feature, but you've done it.
Do you think tat you can add suport for mTLS to ?

Seems not complicated if cert,key and ca are given with base64 string.
With something like REDIS_CLIENT_CERT_BASE64 and so on ?

LordRobinCbz:
I will probably add support for mTLS with Mariadb to ...

Ziris85:
Good point @J0WI , thanks! Just updated README for that.

geidsonc:
This merge is of public interest. Why hasn't it been accepted yet?

Ziris85:
Is this waiting on me for anything? I'll be happy to fix/update whatever else is necessary to get this finished :slightly_smiling_face:

PR #2335, opened 2024-11-18
Full blue angel

ribalba:
This is a full blue angel run. 

Make sure to remove the volumes before run
```
docker volume rm nextcloud-data-mariadb nextcloud-db-mariadb
```
Everything is written so that this should not be needed. But you don't need to run the install and user creation steps.

You can run everything with
```
./runner.py --uri /home/didi/nextcloud-docker --filename energy-tests/usage_scenario-mariadb-all.yml --allow-unsafe --skip-system-checks --dev-cache-build --dev-no-sleep
```

You will need to adapt the SLEEP_TIME to something longer. This is for debugging. This will depend on the machine you are running on. On mine everything is so slow you don't really need a sleep. 

Everything is made for firefox. I left the chrome stuff but it will not work!!!


ribalba:
Sorry this was meant to be against our own fork

PR #2343, opened 2024-11-26
fix(hooks): make the output more uniform an delineated

joshtrichards:
Previously:

- Some essentially similar hook folder states would result in different output (e.g. nonexistent versus empty folders)
- The completion of a given stage wasn't 100% delineated

With this PR, the broader hook handling output is more neutral/consistent while retaining detailed hints where appropriate about the state of things along the way.

Specifically:

- *Always* indicate when we're searching for scripts for a given stage
- Skip early if a given stage's folder is empty too 
- Always indicate when we're skipping (executing nothing for) a given stage + why
- Always delineate when we've completely finished executing all found scripts for given stage
- Add the term we use for this feature in our docs to the output ("hook")
- Minor language tweaks 

Tested with:

- Empty hook folders
- Nonexistent hook folders
- Scripts lacking exec permission in hook folder
- Non-`*.sh` files in hook folders
- Valid executable scripts in hook folder

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -24,30 +24,30 @@ run_path() {
     local hook_folder_path="/docker-entrypoint-hooks.d/$1"
     local return_code=0
 
-    if ! [ -d "${hook_folder_path}" ]; then
-        echo "=> Skipping the folder \"${hook_folder_path}\", because it doesn't exist"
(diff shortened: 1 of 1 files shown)


dvaerum:
Generally it looks fine to me, but you will probably be told to fix your git history because your 2nd commit makes corrects things in your 1st commit.

I would, for readability, make the 1 commit (the 1st) focus on the logic and have a 2nd commit which fixed the `echo` messages there you want them improved.

PR #2352, opened 2024-12-17
Raising value of php.memory_consumption from 128 to 256 at Dockerfile level for NC > versions 30

neokyx:
Hello,

According to the issue reported [here](https://github.com/mother-of-all-self-hosting/mash-playbook/issues/243)
I'm humbly suggesting this change to solve the (too many) recurrent log warning and as requested on :

*PHP OPcache module is not configured correctly The OPcache buffer is almost full. To make sure that all scripts can be retained in cache, it is recommended to add "opcache.memory_consumption" to your PHP configuration with a value higher than "128"... For more details check the documentation.*

It indeed appears frequently, even on modest Nextcloud Installation (4/5 apps on a 2vcpus and 8Go Ram for server configuration)

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -Eeuo pipefail
 
-stable_channel='30.0.4'
+stable_channel='30.0.5'
 
 self="$(basename "$BASH_SOURCE")"
 cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"


joshtrichards:
Hi @neokyx! You may wish to join one of the existing PRs/discussions on this topic. #2090 comes to mind.

neokyx:
Hi @joshtrichards ! Sorry, I should have looked for an open topic before submitting my PR...😅 Thank you for the  answer !

PR #2391, opened 2025-02-18
feature: add ffmpeg

realies:
It would be nice if I don't have to carry [my own](https://github.com/realies/nextcloud-docker) CI for this.

I understand the security considerations around preview generation mentioned in the [examples documentation](https://github.com/nextcloud/docker/tree/master/.examples#full).
However, I believe FFmpeg belongs in the base image alongside ImageMagick because:

1. It's already a recommended component in the official "full" example
2. Like ImageMagick, it's a core media processing dependency
3. Users can still disable preview generation for high-security deployments
4. Having it in the base image reduces duplicate CI efforts while maintaining the ability to control its usage

The change is minimal and follows the pattern of including essential media processing tools while letting administrators control their usage through Nextcloud's configuration.

### DIFF
--- a/29/apache/Dockerfile
+++ b/29/apache/Dockerfile
@@ -6,6 +6,7 @@ RUN set -ex; \
     \
     apt-get update; \
     apt-get install -y --no-install-recommends \
+        ffmpeg \
         busybox-static \
         bzip2 \
         libldap-common \
--- a/29/fpm-alpine/Dockerfile
+++ b/29/fpm-alpine/Dockerfile
@@ -5,6 +5,7 @@ FROM php:8.2-fpm-alpine3.21
 RUN set -ex; \
     \
     apk add --no-cache \
+        ffmpeg \
         imagemagick \
         imagemagick-pdf \
         imagemagick-jpeg \
--- a/29/fpm/Dockerfile
+++ b/29/fpm/Dockerfile
@@ -6,6 +6,7 @@ RUN set -ex; \
     \
     apt-get update; \
     apt-get install -y --no-install-recommends \
+        ffmpeg \
         busybox-static \
         bzip2 \
         libldap-common \
--- a/30/apache/Dockerfile
+++ b/30/apache/Dockerfile
@@ -6,6 +6,7 @@ RUN set -ex; \
     \
     apt-get update; \
     apt-get install -y --no-install-recommends \
+        ffmpeg \
         busybox-static \
         bzip2 \
         libldap-common \
--- a/30/fpm-alpine/Dockerfile
+++ b/30/fpm-alpine/Dockerfile
@@ -5,6 +5,7 @@ FROM php:8.2-fpm-alpine3.21
 RUN set -ex; \
     \
     apk add --no-cache \
+        ffmpeg \
         imagemagick \
         imagemagick-pdf \
         imagemagick-jpeg \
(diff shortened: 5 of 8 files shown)


joshtrichards:
It also adds something like ~500MB to the image size last I checked. ;-)

Between that and:

- it being easy enough to do a `build` directly within a Compose file (no need to build+upload an image via CI)
- more regular use of Imaginary for offloading preview generation these days with Nextcloud (which embraces the micro-services model)

I'm not sure if it's even as important as it once was. So I'm currently against adding it to the base image.

realies:
It adds 133.56 MB to the amd64 and 120.19 MB to the arm64 latest images right now.
- https://hub.docker.com/_/nextcloud/tags?name=latest
- https://hub.docker.com/r/realies/nextcloud/tags?name=latest

It's not particularly difficult to make custom solutions for this, but:
- doing a `build` directly within a Compose file does not work great with a lot of auto-update systems that just check upstream for a new image
- does not depend on other services
- is not a workaround

The idea of this PR is to bring a dependency for a Nextcloud functionality that requires it.
Similarly to how ImageMagick is included.

mateuszdrab:
It would be nice if ffmpeg was built into the image or the full build was made available.

I've been building my own image from the `.examples/dockerfiles/full/apache/` file for years due to the need for video previewers and other functionality.

Compose approach won't work for me because I'm running Kubernetes

Regarding Imaginary, I don't think it can provide thumbnails for video files, can it?

prairietree:
I am also building my own image to include ffmpeg. And I just came here to look at the Dockerfile to see if that was still needed. So I would like to have it in the base image but I can see the argument for keeping the size down. I also had a bit of configs to get previews to generate correctly. Also I got preview:pre-generate running on the host. The solution to all these might be to have another docker image that is focused a little more on media. 

I have not used Imaginary. Looking into it just now the only thing I could find that seemed to be being kept updated was https://hub.docker.com/r/nextcloud/aio-imaginary/tags but I could not find much documentation on it. The only thing I found was this old post https://okxo.de/speed-up-nextcloud-preview-generation-with-imaginary/. 

Looking at the AIO Imaginary Docker file (https://github.com/nextcloud/all-in-one/blob/main/Containers/imaginary/Dockerfile) it does not include ffmpeg. I also could not find any documentation on what file formats it supports. 

If anyone has more information on Imaginary let me know. Given the lack of documentation and up-to-date examples I am not to inclined to spend much time on trying to get to work and testing out what it can do.

BrainStone:
Having to build you own images defeats one major benefit of using images/containers in the first place.

It's software required for full functionality. Why do I even have to think to worry about providing it myself?
Even if it's 10GB, why would it matter?

And if you're really that worried about having larger images, provide a new set of tags. Either tagging the images without ffmpeg "minimal" or the ones with "full", "ffmpeg" or similar.

Right now the only feasible option to have ffmpeg available if your infrastructure offers no reasonable way to detect upstream changes and build up to date images itself, is by relying on thrird party images, like https://hub.docker.com/r/kilrah/nextcloud-ffmpeg, https://hub.docker.com/r/dmgarland/nextcloud-with-video-preview or https://hub.docker.com/r/aquarat/nextcloud-ffmpeg. Which brings a whole host of issues with it. The images not being up to date just being the least important.

PR #2403, opened 2025-03-02
Add ability to force upgrades even when Nextcloud version did not change

boomer41:
We'd like to bundle custom apps into the image by extending it.
However, we ran into issues when we want to push an update to an app in the image when the Nextcloud version did not change.

We solved that by
- adding a new hook called `pre-initialization` that runs even before the need for an initialization/upgrade is determined, and
- by forcing an initialization run when a marker file `/tmp/nextcloud-force-initialization` exists.

The later can be dynamically created by a script triggered the new hook mentioned above.
With this setup, the script can then **additionally** start an initialization run.
When the initialization is finished, the marker file is automatically removed.
The script **can not** stop an upgrade from happening, it can only additionally start one.


boomer41:
Any updates on this PR from the maintainers? :)

boomer41:
Can someone from the maintainers look at these PRs, please? :) Thank you!

@joshtrichards

PR #2445, opened 2025-07-09
Replace use of paulczar/omgwtfssl image in examples

Abhicodeitout:
@joshtrichards @tzerber  

### Summary

I have replaced the outdated `paulczar/omgwtfssl` image used for self-signed certificate generation with a custom `cert-generator` service built on Alpine and OpenSSL. The new solution supports environment-based configuration, improving flexibility and maintainability.

### What was changed

- Removed `paulczar/omgwtfssl` service from `docker-compose.yml`
- Added a new `cert-generator` service with:
  - Dockerfile using `alpine` and `openssl`
  - `entrypoint.sh` script to read cert parameters from environment variables
- Mounted cert output to existing shared `certs` volume for use by other services (like `nginx-proxy`)

### Why it matters

- `paulczar/omgwtfssl` is no longer maintained
- The new solution is minimal, secure, and customizable
- Aligns better with modern Docker Compose practices

### How to test

1. Build and run the `cert-generator`:
   ```bash
   docker compose build cert-generator
   docker compose up cert-generator


Abhicodeitout:
@joshtrichards  can you review this one !!

tzerber:
While I haven't tested it, it looks okay. A note : Add some documentation around it. I know about several ways of "oneline ssl via random alpine image" long commands, and these tend to confuse people. Use some sort of placeholder like <PUT.DOMAIN.HERE> and maybe add few lines to the docs about it.

Overall, it looks good. Will test it as soon as I can.

Abhicodeitout:
@tzerber  i have just added a section which includes the replacement of the paulczar/omgwtfssl i don't think i have used any domain information .
and document updation can you clarify little on that .

tzerber:
> @tzerber i have just added a section which includes the replacement of the paulczar/omgwtfssl i don't think i have used any domain information . and document updation can you clarify little on that .

Sorry for the late reply, but I was away from the civilization for a while.

Recently we started adding additional information in the readme.md / inline on the compose file for any non-nextcloud third party programs providing both links to their documents/manuals and some information on how to use that software in combination with Nextcloud.

Abhicodeitout:
@tzerber so what shall i do next

PR #2485, opened 2025-10-20
fix: call to deprecated command in entrypoint

### LINKED ISSUE #2484: Call to deprecated `maintenance:install` in entrypoint.sh
## Setup
I am using [NixOS](https://github.com/matthiasemde/Pandora/blob/main/services/nextcloud/flake.nix) to run nextcloud, so it is probably hard to reproduce, but I already have a fix for the bug as it is pretty obvious (at least I think so 😅 ). 

## Bug Description
After updating to version 32.0.0 from 31.0.9, I got an error:

`Retrying install...
                                                 
  Command "maintenance:install" is not defined.
                                                 
  Did you mean one of these?
      app:install
      maintenance:data-fingerprint
      maintenance:mimetype:update-db
      maintenance:mimetype:update-js
      maintenance:mode
      maintenance:repair
      maintenance:repair-share-owner
      maintenance:theme:update
      maintenance:update:htaccess` 

## Workaround
I added the following line to my [Dockerfile](https://github.com/matthiasemde/Pandora/blob/main/services/nextcloud/Dockerfile), which I use to build a derivation of this Image:
`RUN sed -i 's/maintenance:install/app:install/g' /entrypoint.sh`
Afterwards the new version ran smoothly 😊

matthiasemde:
This PR fixes #2484.

Honestly, I didn't test the fix for the _fpm_ images (only apache), but I am pretty confident, that it should work for all three images.

### DIFF
--- a/32/fpm/entrypoint.sh
+++ b/32/fpm/entrypoint.sh
@@ -240,7 +240,7 @@ if expr "$1" : "apache" 1>/dev/null || [ "$1" = "php-fpm" ] || [ "${NEXTCLOUD_UP
                         echo "Starting nextcloud installation"
                         max_retries=10
                         try=0
-                        until  [ "$try" -gt "$max_retries" ] || run_as "php /var/www/html/occ maintenance:install $install_options" 
+                        until  [ "$try" -gt "$max_retries" ] || run_as "php /var/www/html/occ app:install $install_options" 
                         do
                             echo "Retrying install..."
                             try=$((try+1))


joshtrichards (changes requested):
`occ app:install` and `occ maintenance:install` are *very* different commands. Something else is going on in your environment if this fixes something for you. `occ maintenance:install` is only available at install time.

matthiasemde:
Ok. Guess I was a bit too eager there... I'll give an update, if I find out what I messed up there.

PR #2504, opened 2025-12-04
Keep the main codebase in container ephemeral storage

### LINKED ISSUE #2044: Publish an image that is ready for scalable cloud deployments
Hi Folks,

thank you for providing a next cloud container image!

Unfortunately I spent a lot of time to get nextcloud running in my Kubernetes cluster. Unfortunately it's still not running and I would like to consider/discuss the following aspects.

## rsync on startup?

As far as I can see, the entrypoint script `rsync`s files from the container image to a different location:
https://github.com/nextcloud/docker/blob/7bd3b7b6dbd694bab3612a28b17a9767f2339c9e/docker-entrypoint.sh#L177

In my eyes, a containerized application should be self-containing and source code should be ran from the image (ideally stored there in read-only mode, immutable for security reasons). 

Can someone explain to me, why this image copies source files (even third party code, like the aws library) to another location before starting? For me, this seems to be an anti-pattern in the world of container images.
I read something about upgrading, but I think upgrading in the container world means - not to compare any version.php files residing in a persistent file system - but to change the tag of the image. **The application itself (again running from a readonly part of the filesystem) should detect its current version and compare this to the database. Then it has to handle the upgrade process (if required) with lockfiles, leases, migrations, etc.** Also from a performance perspective I disagree with the approach to rsync application files before application startup. There could be slower persistent media like NFS and startup takes a long time, then you get into trouble with the readinessProbe (i know, I can set a startupProbe - but it feels dirty)...

## instanceid?

The running application writes an `instanceid` to the configuration which has to be persisted, otherwise the application will try to re-install after pod restarts.

Can someone explain to me, why this `instanceid` is needed and why it has to be written into the configuration? I would expect it in the database, but not in the file system. **The containerized application should not get any state** (we have a database for that) and if it has to, it should be stored in the `data` directory. Otherwise scaling the application is a pain as well.

I would be really happy if someone could shed some light at this `rsync` idea. Until then, I'll have to build my own image.

Thank you very much :)

BR Matthias

Related issues:
https://github.com/nextcloud/docker/issues/1006
https://github.com/nextcloud/docker/issues/1582
https://github.com/nextcloud/docker/issues/1050
https://github.com/nextcloud/docker/issues/869

mcronce:
This has several advantages
* Faster startup for initial container and during upgrades
  * On my system (shared storage for `/var/www/html` was cephfs on 2x10Gb network and consumer-grade NVMe drives), updates between 29, 30, 31, and 32 went from ~135 minutes to ~5 minutes.  The main culprits for the difference were the `rsync` step and the code integrity check.
* Improved performance when volume storage is slower than container storage.  This is common in cases where multiple replicas are running - especially on multiple machines, e.g. Kubernetes, because shared mutable network storage is necessarily slower than local.  Even with the Linux filesystem cache, it needs to cache less optimistically than with a local filesystem to ensure consistency.
  * On my system, fully loading the login screen went from ~18 seconds to ~800ms
* More idiomatic container setup - the application is contained in the container.  This means that tampering with application code isn't permanent, without an explicit code integrity check step.

It also _may_ fix #2044; I don't want to speak for the person who originally filed it, but it sounded like they were asking for the same thing.

This is, obviously, a breaking change - anybody currently mounting in all of `/var/www/html` needs to now mount in several subdirectories instead.  I'm not sure whether the breaking change is preferable over the maintenance burden of having a separate set of images; if a separate set of images is desired, I'll redo it that way.  Notably, the helm chart already mounts them all in separately; the only change needed there is to no longer mount in `/var/www` and `/var/www/html`.  I haven't looked for docker-compose files or anything like that.

Other than that, I don't believe I broke anything.  The example theme and any changed "built-in" config/custom apps will still be copied over when the version changes.

### DIFF
--- a/32/apache/Dockerfile
+++ b/32/apache/Dockerfile
@@ -132,7 +132,11 @@ RUN { \
     chown -R www-data:root /var/www; \
     chmod -R g=u /var/www
 
-VOLUME /var/www/html
+VOLUME /var/www/html/data
+VOLUME /var/www/html/config
+VOLUME /var/www/html/custom_apps
+VOLUME /var/www/html/themes
+VOLUME /var/www/tmp
 
 RUN a2enmod headers rewrite remoteip ; \
     { \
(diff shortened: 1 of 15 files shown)


especially-relative:
This has been open for a while but it would be great if it got more attention

It solves a really significant design flaw and anti-pattern that has existed in the `nextcloud/docker` since it was created. I haven't encountered single other container across any of containerized applications that I have used in the last 10+ years which use rsync or similar to throw it's own source to a volume, and this is a reasonable fix for it.

Yes, it may impact some setups, but that needs to be addressed eventually. Either with multiple different tags, a pile of new env vars for optional behavior, or the cleanest solution (this PR) which is to just have the image that people see as the "default" nextcloud container behave in a standard, predictable way.

especially-relative:
```yaml
      containers:
        - name: nextcloud
          image: docker.io/library/nextcloud:33.0.0-fpm
          imagePullPolicy: Always
          command: ["php-fpm"]
          volumeMounts:
            - name: nextcloud-image
              mountPath: /var/www/html
              subPath: usr/src/nextcloud
              readOnly: true
      volumes:
        - name: nextcloud-image
          image:
            reference: docker.io/library/nextcloud:33.0.0-fpm
            pullPolicy: IfNotPresent
```
There is a workaround for a cleaner deployment using image volume types but it would still be nice to not even need to do this. At least I don't have to maintain my own images anymore.

PR #2508, opened 2025-12-11
Extract /usr/src in its own stage

trajano:


trajano:
oops didn't mean to create this

PR #2520, opened 2026-01-21
feat: Add support for Redis host TLS

sguyennet:
### Summary

This PR adds support for using `tls://` at the begining of the REDIS_HOST environment variable in the entrypoint script as described here:
https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/caching_configuration.html#connecting-to-single-redis-server-over-tls

This allows users to use TLS connection to an external Redis server.

### Testing

Verified with a managed Redis server that require a TLS connection.


J0WI:
This duplicates https://github.com/nextcloud/docker/pull/2334

PR #2541, opened 2026-03-04
yamllint

J0WI:


.examples/docker-compose/insecure/mariadb/apache/compose.yaml
tzerber: Out of curiosity, since when it's needed to have document separator in compose files, or the linter didn't know they are compose files and added them? While I know what `---` in yaml means I haven't seen it in compose files. I have nothing against it(and it doesn't break examples), just curious (and I try to learn from others ).

PR #2550, opened 2026-03-23
feat(redis): Eliminate runtime-generated `redis-session.ini`

### LINKED ISSUE #763: Introduced redis session handler breaks entrypoint of non-root images

joshtrichards:
Fixes #763 

- avoids stale `redis-session.ini`
- reduces mutable filesystem state
- makes PHP config more declearative
- aligns Redis PHP session config with the existing env-substitution pattern already used for other PHP settings like `PHP_MEMORY_LIMIT`
- eliminates incompatibility of Redis session handler configuration with "rootless" scenarios (#763)

Requirements:
- Existing container-level `REDIS_*` image configuration variables remain as they exist today
- Redis is enabled only when configured

For standard environments (i.e. those that already work today with the existing implementation), the result should be:

* No change

For environments running containers w/o root:

* They should start working (i.e. using the specified REDIS_HOST as the PHP session handler... instead of generating an error about being unable to write to the ini file and falling back on file-based session storage)

**This is a prototype in its current state.**

Key things to confirm:

* What happens when the env vars are unset?

P.S. If the empty values being always passed through is problematic.... This tests the full “drop redis-session.ini entirely” idea. If this runs into challenges, we could generate a separate ini file (only when Redis is configured) in the entrypoint, but still source values from env. That would test the env-based composition separately from the fallback problem.


joshtrichards:
@copilot resolve the merge conflicts in this pull request

PR #2557, opened 2026-03-26
Refactor Redis connection for TLS and unix sockets

### LINKED ISSUE #1785: Obscure way to use unix socket for Redis

J0WI:
fix #1785
fix #1968
fix #2334
fix #2520

### DIFF
--- a/.config/redis.config.php
+++ b/.config/redis.config.php
@@ -4,14 +4,16 @@ if (getenv('REDIS_HOST')) {
     'memcache.distributed' => '\OC\Memcache\Redis',
     'memcache.locking' => '\OC\Memcache\Redis',
     'redis' => array(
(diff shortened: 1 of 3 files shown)


sguyennet:
Any news regarding this PR?

Some cloud providers force the use of TLS for the Redis connection. It would be nice to have that supported out of the box in the image without the need to create a custom one.

PR #2598, opened 2026-08-28
Allow configuring the cron.php schedule

sebastiangaiser:
The interval for cron.php is baked into the image at build time as */5 * * * * in /var/spool/cron/crontabs/www-data, and cron.sh offers no configuration at all. Changing it currently requires replacing the entrypoint and reimplementing the crond invocation downstream.

Add NEXTCLOUD_CRON_SCHEDULE, honoured by /cron.sh. When set, the crontab is written to a writable directory and crond is pointed at it with -c, so the baked crontab stays in place and an image update cannot silently restore the old interval. When unset, -c receives busybox crond's own default directory, so the behaviour is unchanged. Rewriting the baked crontab in place is not an option because it is owned by root with mode 644 and therefore not writable when the container runs as an arbitrary user.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

### DIFF
--- a/32/apache/cron.sh
+++ b/32/apache/cron.sh
@@ -1,4 +1,12 @@
 #!/bin/sh
 set -eu
 
-exec busybox crond -f -L /dev/stdout
+crontabs=/var/spool/cron/crontabs
+
+if [ -n "${NEXTCLOUD_CRON_SCHEDULE:-}" ]; then
+	crontabs="${TMPDIR:-/tmp}/nextcloud-crontabs"
+	mkdir -p "$crontabs"
+	echo "$NEXTCLOUD_CRON_SCHEDULE php -f /var/www/html/cron.php" > "$crontabs/www-data"
+fi
+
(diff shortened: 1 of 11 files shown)


J0WI:
The 5 minute schedule is recommend by Nextcloud: https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/background_jobs_configuration.html#cron

If you need a special value, mounting a custom script is the way to go. I don't think that this is a common enough case to introduce a separate ENV. Alternatively, you can trigger `cron-php` with `docker exec` from outside.

sebastiangaiser:
Hey, yes that's recommended but I would like to change it anyway. 
Background: I'm using the Nextcloud Helm chart, where you can configure a [CronJob schedule](https://github.com/nextcloud/helm/blob/7b2c957eaa4b69345b11ffcf96a983d34aad1f71/charts/nextcloud/values.yaml#L711). My idea was to have the same possibility directly inside the container.

